PR #36, opened 2015-01-26
read timestamp as now if one isn't present

xdl:
If the message doesn't come a timestamp, still process it using the current time.

(Motivation for this was that the messages published using the rabbit web console don't come with a timestamp by default)


iancooper:
Good spot, we want to be as tolerant as possible

PR #38, opened 2015-02-03
Fixed RabbitMQ TCP channel not closing when stopping Dispatcher

MrHypnos:


iancooper:
Thanks for working through that, great fix.

PR #74, opened 2015-04-23
Add Machine.Specifications.Runner.Console to fix test script

dstockhammer:
TestMe.bat tries to execute the runner from Machine.Specifications.0.8.3, but the package isn't installed.

In 0.9 versions of mspec, the console runner has been removed from the main package and is now distributed separately in the Machine.Specifications.Runner.Console package.
I have added that package to the test project and updated the batch file to call the correct version of the runner.


iancooper:
Ta.

PR #75, opened 2015-04-23
Remove setter from IRequest.Id 

dstockhammer:
Remove setter from IRequest.Id and make setters in all implementations private.


iancooper:
Ta.

PR #84, opened 2015-05-29
Improve logging of message creation exceptions in the RmqMessageCreator

george-ayris:
To aid debugging when de-serialisation of RabbitMQ messages goes wrong.


holytshirt:
thanks @george-ayris :+1:

PR #87, opened 2015-06-03
This fixes #30 RabbitMQ pipeline errors.

### LINKED ISSUE #30: RabbitMQ gives Pipeline Errors when using more than one peformer
The Taskmailer example will give pipeline errors when you use more than one performer

ianbattersby:
The only issue I could find with this was that RMQ doesn't support pipelining as per the AMQP specification so needs a little time to clear the previously executing operation. Used the `ConnectionPolicyFactory` settings as isn't dissimilar.

For review/comment :)

### DIFF
--- a/Brighter/Examples/TaskMailer/App.config
+++ b/Brighter/Examples/TaskMailer/App.config
@@ -31,7 +31,7 @@
   </rmqMessagingGateway>
   <serviceActivatorConnections>
     <connections>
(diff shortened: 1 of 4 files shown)


ianbattersby:
Erm, I should have tried more than 2 before PR'ing this. BRB ;)

ianbattersby:
Needs some more thought, the fact that these Channel operations are banging into each other obviously means event separate instances of IModel aren't thread safe, which means locking operations between threads.

PR #104, opened 2015-07-06
When requeing a message we should use the default (empty) exchange so…

yiannistri:
… that it gets delivered to the right queue


ianbattersby:
👌

benjamin-hodgson:
Why not use `basic.reject` or `basic.nack` to requeue messages? https://www.rabbitmq.com/nack.html

yiannistri:
I guess it depends on whether we want the message to be delivered to the back of the queue or not. From what I understand since v2.7.0 of RabbitMQ, the broker will requeue the message in the order it was published. Thoughts?

PR #109, opened 2015-07-08
Removed null check for consumer on creation

denizkocak:
If I restart RMQ server, logs and RMQ console suggest that the consumers are connected successfully but they throw EndOfStreamException when reading messages from queue.
The problem was, there was a null check on consumer creation logic which was preventing the it's queue connection to be refreshed after the restart of RMQ server. Removing that check fixes the issue.


holytshirt:
:+1:

PR #111, opened 2015-07-09
Removed the logger from RequestHandler<T> base class

benjamin-hodgson:
`RequestHandler` took a logger into its constructor, so that when it continues the pipeline by calling `_successor.Handle` it's able to log the names of the current handler and the next handler.

We found this painful, because when we converted to Brighter every single one of our handlers had to take an `ILog` dependency - and it was _Brighter's_ `ILog`, not the logging interface of our choice. (Even if we were using LibLog, we still wouldn't be able to use the copy of `LibLog` that we'd imported.) So the options were:
  * Have _two_ loggers (of different interfaces) injected into every
    handler - one so that Brighter can log pipeline continuations and
    another so that we can do our application-level logging.
  * Abandon our old logging interface and use the _Brighter_ `ILog`
    everywhere.

Neither option is particularly palatable. In my opinion, the logger should be injected into Brighter _once_ at the entry point to the whole system (when we build the command processor), not at every handler.

So with that in mind, I've excised the `ILog` from `RequestHandler` altogether. I replaced it with a more generalised solution - an `event` (a C# `event`, not a Brighter `Event`) on `IHandleRequests` to indicate that the pipeline is continuing. `RequestHandler` invokes that event when it calls `_successor.Handle`, and propagates calls from its successor up the pipeline. In `CommandProcessor` itself, the event is bound to the actual logger.

I kept the old constructor (and field) in, because there'll be code which depends on it, but I deprecated them.

I'm going to throw up another pull request to remove usages of the deprecated constructor in a little while

### DIFF
--- a/Brighter/paramore.brighter.commandprocessor.tests/CommandProcessors/CommandProcessorTests.cs
+++ b/Brighter/paramore.brighter.commandprocessor.tests/CommandProcessors/CommandProcessorTests.cs
@@ -23,6 +23,7 @@ THE SOFTWARE. */
 #endregion
 
 using System;
+using System.Linq;
 using System.Threading.Tasks;
 using FakeItEasy;
 using Machine.Specifications;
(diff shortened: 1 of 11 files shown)


iancooper:
I don't think we want to do this. It's a breaking change for most clients, and I'm not sure that it's not based on a misunderstanding of how to use LibLog

PR #112, opened 2015-07-10
Revert Newtonsoft to 6.0.8

tapmantwo:
Newtonsoft was recently upgraded to latest, however it has behavioural changes that cause other components (in our case, Nest) to fail.  As such, reverting this as the uplift wasn't strictly required.


ianbattersby:
+1

Issue #121, opened 2015-07-27
When a message is discarded, the error log should dump out the content of the message


benjamin-hodgson:
At the moment, the error says:

```
MessagePump: Have tried 30 times to handle this message c92f7655-ed00-4eae-8a50-5562fbda2a25 (original message id 2e62348b-1c72-4de3-aeda-0d39aa4bb6b3) from Files.MyHuddle.UserRemovedFromGroup.Event on thread # 48, dropping message
```

When we then go to manually replay a message, there is no information in the log about which aggregate the message was from. It would be much easier to debug issues (by replaying messages, etc) if the error contained a JSON dump of the message.

holytshirt:
@benjamin-hodgson check the merge and close if you are happy

benjamin-hodgson:
:+1:

PR #122, opened 2015-07-28
Add message body to log output when a message is dropped.

tapmantwo:
Hi,

When a message is dropped, we get a useful error message alerting us that it has happened.  However, it only logs the message id, and that makes taking remedial action a bit more difficult, because we have to locate the original message to understand what has changed.

It would be great if the message could also include the message body.

### DIFF
--- a/Brighter/paramore.brighter.serviceactivator/MessagePump.cs
+++ b/Brighter/paramore.brighter.serviceactivator/MessagePump.cs
@@ -303,12 +303,13 @@ namespace paramore.brighter.serviceactivator
 
                     if (Logger != null) 
                         Logger.ErrorFormat(
-                            "MessagePump: Have tried {2} times to handle this message {0}{4} from {3} on thread # {1}, dropping message", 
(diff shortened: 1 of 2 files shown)


Brighter/paramore.brighter.serviceactivator/MessagePump.cs
```diff
@@ -303,12 +303,13 @@ private void RequeueMessage(Message message)
 
                     if (Logger != null) 
                         Logger.ErrorFormat(
-                            "MessagePump: Have tried {2} times to handle this message {0}{4} from {3} on thread # {1}, dropping message", 
+                            "MessagePump: Have tried {2} times to handle this message {0}{4} from {3} on thread # {1}, dropping message. Message Body: {5}",
```
holytshirt: Probably better to put in a Environment.NewLine so the body is on a new line. Will be easier to read then

holytshirt:
This fixes issue #121

PR #123, opened 2015-07-28
Ignoring packages folder

spadger:
Hi, You may be doing this deliberately, but in case you aren't I have added the packages folder to the gitignore file.  Over time, its making the repo take longer and longer to clone.


holytshirt:
@spadger Hi, yes we are doing deliberately. We been stung too many time with nuget been down. We know it makes cloning a pain. Hopefully as nuget becomes more stable we'll remove it and clear it's history.

spadger:
Good stuff :)

holytshirt:
Join the Gitter chat https://gitter.im/iancooper/Paramore and can bring it up there too. Good place to see what others think too.

iancooper:
Going to close. Yes, we originally did this deliberately due to issues with both NuGet stability and concerns around loss of packages. But, it's painful in that the size of the repo now makes it difficult to clone. So we probably do want to move away from this. The problem is we would need to clean the packages file from the repo. That's hard, although there are some tools that promise to do this.

PR #127, opened 2015-07-29
Revert "Brighter msg viewer repost"

penderi:
Reverts iancooper/Paramore#126


penderi:
We shoudl delete in favour of the original if possible.

iancooper:
No need, not this test that is failing

PR #129, opened 2015-08-03
fix logging bug

benjamin-hodgson:


holytshirt:
oops think we'll take that one :+1:

Issue #76, opened 2015-04-29
Get an example running with Azure Service Fabric


iancooper:
It should be possible to use Azure Service Fabric for hosting services instead of TopShelf Windows Services and IIS hosted APIs. I'm going to do a proof of concept, for part of an upcoming Microservices talk.

iancooper:
@dstockhammer now has his

loneshark99:
Is this done yet?

dstockhammer:
No, not really. I've been playing around with Azure Functions and I think that environment is superior to Service Fabric for our needs. However there's a lot of churn with VS Function tools at the moment so that I've decided to wait until a VS 2017 version is available. See https://github.com/Azure/Azure-Functions/issues/201

loneshark99:
ok

PR #148, opened 2015-10-21
Fixes Issue #76 - Should Send be constrained to take a Command and

penderi:
Publish an Event?
- Extracted new IEvent interface
- Constrained CommandProcessor etc
- MessagePump now IGNORES messageType and uses type for dispatch


iancooper:
Switching on type always makes me nervous because OO says, use polymorphism. In addition appveyor is failing. Let's have a chat

iancooper:
Change the way we check for invalid type for MessageType. Be less tolerant, you should match message type and irequest type

PR #149, opened 2015-10-22
Issue 110 log

penderi:
Default constructor on RequestHandler (calls GetLogger)
Added ILogProvider to CommandProcessorBuilder


iancooper:
Reworking no logger name

PR #152, opened 2015-10-22
Fix logger breaks

penderi:


iancooper:
We might not to run through this, as I changed some stuff

PR #159, opened 2015-12-26
Azure service bus

yiannistri:
Initial attempt at supporting Azure Service Bus. Added main project and example project which will be hosted using Nancy. Currently creating a new MessageSender every time we publish which is expensive, eventually we will introduce a MessageSender pool to increase throughput.


iancooper:
@yiannistri Merged in, thanks my friend

PR #160, opened 2016-01-01
Async start

Red-F:
Let's see if we can use this code as a base for improvements and discussion for issue #3.


iancooper:
@Red-F Cheers, I've pulled the branch so that we can all take a look at this one and comment

PR #161, opened 2016-01-05
Add async/await support for Send via the implementation of SendAsync

Red-F:
Enable the use of async/await with the in-line (synchronous) processing of a command.

Extend the IAmACommandProcessor interface with "Task SendAsync(T command) where T : class, IRequest;". And add the interface IHandleRequestsAsync to be implemented by the async request handler.

Extend the framework to support the registration of async handler factories, subscription by async handlers and various other required plumbing;


iancooper:
Thanks again for this PR

PR #163, opened 2016-01-19
This PR adds the remaining async versions of the builtin handlers;

Red-F:
Added async versions of the builtin handlers;
Added supoort for .ConfigureAwait when awaiting


iancooper:
Thanks

PR #167, opened 2016-01-25
Implement PostAsync

Red-F:
Introducing IAmAnAsyncMessageStore and IAmAnAsyncMessageProducer;
Added constructors to CP for async message store and async message producer
Re-factored MessagingConfiguration and CommandProcessorBuilder to allow usage of async message store and async message produce via the fluent API

In addition made some test methods truly async to prevent future hard to find errors.


iancooper:
Thanks. We should probably use the Async Polly methods. I'll review everything else to check on it

PR #176, opened 2016-03-21
Update BuildingAnAsyncPipeline.html

edblackburn:
When discussing async should references to RequestHandler<T> now refer to RequestHandlerAsync<T> ?


edblackburn:
Actually should https://github.com/edblackburn/Paramore/blob/patch-1/BuildingAnAsyncPipeline.html#L144 IAmAHandlerFactory refer to IAmAHandlerFactoryAsync too? Is there any reason why you'd use the none Async type or member considering the clarification here: https://github.com/iancooper/Paramore/blob/gh-pages/AsyncDispatchARequest.html#L87

iancooper:
Ta.

PR #180, opened 2016-04-08
support all types supported by rabbit in message headers instead of just strings

dstockhammer:
Currently, when a message header contains anything other than a byte[], the message consumer will blow up because the message creator tries to convert all header values into an utf-8 string.
The rabbit mq client supports more types than strings, (like int, long, double, bool and null), so this change will only convert to string if the value is byte[], otherwise just use the value.


dstockhammer:
I just pushed another change we discussed to my master, because I didn't realize that it will (obviously!) be added to this PR. I should probably have created another branch for this change in isolation, so if you want I can do that.
On the other hand, both changes are small and kind of related, so they might as well be in the same PR.

iancooper:
That's cool, I'll just accept them as rollup fix

PR #182, opened 2016-04-18
Fix SqlMessageStore duplicate key under Polly (resolves #178).

### LINKED ISSUE #178: SqlMessageStore duplicate-key exceptions fail under Polly policy
We noticed some time ago that we were getting Polly failing on inserting messages into the message-store when RMQ wasn't available; what we observed was;

1. Message is queued with `Post()`
2. Message is added to the SQL message store
3. Attempt is made to connect to RMQ and fails - Polly kicks in - hazzah!
4. On retry message is added to the SQL message store
5. SqlException is triggered as already exists
6. Polly kicks in and starts a new retry without following `try/catch` statements

Do the catching of the `SqlException`'s which then checked to make sure if it was merely a duplicate-key and ignore it, simply didn't work under Polly. One option could be to tell Polly to ignore SqlException but this would exclude a raft of other possible scenarios. More so using exceptions for logic-flow - especially with SQL Server proper - seems wrong. Another option _could_ be to try and write a Policy that filters the `SqlException`'s but this policy would be unique to SqlMessageStore and only there for SqlServerCe.

My thoughts are that anyone in their right mind wouldn't be using SqlServerCe for production and at some point we need to get rid, but for the purposes of a quick fix having a SQL Server version that does it properly and a SqlServerCe version that does it's best but isn't atomic, is perhaps workable.

PR incoming for consideration.

ianbattersby:
I'd quite like to get this in if no objections remain?

### DIFF
--- a/Brighter/paramore.brighter.commandprocessor.messagestore.mssql/MsSqlMessageStore.cs
+++ b/Brighter/paramore.brighter.commandprocessor.messagestore.mssql/MsSqlMessageStore.cs
@@ -56,8 +56,7 @@ namespace paramore.brighter.commandprocessor.messagestore.mssql
     public class MsSqlMessageStore : IAmAMessageStore<Message>, IAmAMessageStoreAsync<Message>,
         IAmAMessageStoreViewer<Message>, IAmAMessageStoreViewerAsync<Message>
     {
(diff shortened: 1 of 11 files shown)


iancooper:
Ta.

PR #192, opened 2016-05-11
Update SqlMessage/CommandStore Schema and unique key checks.

ianbattersby:
As noted in PR #191 we observed some slow INSERTION times because the physical page splits in the message store tables became quite large. To resolve this we added an IDENTITY integer based column with `MessageId`/`CommandId` instead having a "Unique Index". Ironically this changes the error code back to `2601` but have now updated code to check for both `2601` and `2627` to cover both schema approaches by default (in a hope to be backward compatible).


holytshirt:
looks good :+1:  @ianbattersby you still happy with this?

PR #193, opened 2016-05-17
Adding binary message support to allow sending and receiving of Google Protocol Buffer encoded messages

smclewin:
Changed the MessageBody object to carry a byte array rather than a string (Value).  Whether the byte array or the compatible string property Value is used is driven based on the ContentType in the MessageHeader object.  To be compatible with existing Brighter solutions the ContentType defaults to text/plain.

Only the RabbitMQ messaging gateway uses the binary message body feature.

Also included are four example projects under the folder Examples/ProtoBufMessages.

ProtoBufReceiver receives and writes messages to the console.

ProtoBufSender generates messages

ProtoBufMessages is a shared assembly that contains Protocol Buffer defined messages, as well as custom build steps to call the protoc.exe compiler to convert the .proto files into c# objects.  The protoc compiler calls to a custom plug-in that generates Brighter extensions for ProtoBuf code, so that ProtoBuf objects implement interface IRequest (or ICommand, or IEvent) and to create the boilerplate message mapper code.

protoc-gen-brighter is a C++ project that only builds in release mode.  It creates the protoc-gen-brighter.exe plug-in that is called as part of the .proto file compilation.  The executable is embedded with the rest of the Google Protocol Buffers libraries. The project is included here to allow modification.


holytshirt:
Pulling this locally to have a look at. 
To many changes to review on github, the UI is locking :smile:

smclewin:
By percentages very little of the change is within Brighter.  More is with the example projects and the single largest contributor to file count is the ProtoBuf SDK, which is only needed for the C++ headers and libraries used to build the protoc-gen-brighter.exe plug-in for the protoc compiler.  That plug-in generates the bits of Brighter glue code to make protobuf objects native to Brighter by implementing IRequest or a derivative.  

But yes, by sheer volume of files it is a lot of change.  Thanks for taking the time to review.

PR #200, opened 2016-08-16
Ensure any passed logger is passed to all dependencies

penderi:
- Minor but we think it will greatly help our ability to Send a relatively large volume of messages in a short time.


mehdime:
What is the reason for this out of interest? A common pattern used for loggers (which we use ourselves) is to use loggers named after the class in which they're used. The name of the logger is then included in all log entries making it a lot easier to:

1. Find out where a log entry comes from, and
2. Filter (in or out) the logs written by a specific class

By passing logger instances around, we're largely defeating the point of named loggers. 

This is arguably not the end of the world but I'd love to know why we're doing it.

iancooper:
@mehdime Testing. If you use static loggers it makes testing harder.

mehdime:
@iancooper I'd definitely not use static loggers either. We inject our logger instances (and let StructureMap in our case create logger instances named after the concrete type that's being built).

Sorry but still not sure what the reason for this is :)

PR #205, opened 2016-11-14
Brighter assemblies produce NET451

Red-F:
If you're interested, this PR will add .NET 4.5.1 artefacts for:

- the paramore.brighter.commandprocessor.* assemblies with the exception of the the -.messageviewer.* assemblies
- all the -\lib assemblies
- some of the examples


iancooper:
@holytshirt  is looking into this

PR #208, opened 2016-12-21
internal Clock

dstockhammer:
Clock is used for unit testing and should *not* be part of the public API


iancooper:
I'm not sure I agree on this one. We use Clock internally to set times, and let you override them. By making it internal we rely on our own tests using InternalsVislbleTo which is a smell. If we are able to use it in testing, it ought to be pe part of our API. I'm less worried about name clashes with other projects, that is what using statements are for. I'm not inclined to take this one at the moment. 

Feel free to argue your case more and we can reopen, especially if I am a 'lone voice' on this

PR #212, opened 2017-02-14
Interpreter component.update

thynquest:
I have removed the unused type parameter (GetHandlers and GetAsyncHandlers methods) since the type is already passed during the creation of the interpreter object (see Generic Send<T> method on CommandProcessor class)


iancooper:
Thanks for this one!

PR #217, opened 2017-03-29
Upgrade to VS2017 / msbuild

dstockhammer:
Also moving to xunit. I know you said you're not a big fan of this, but I figured it since it was easy enough to do, I'll give it a spin.
Looks good so far, a couple of tests are still failing. It could be that those are known issues though, maybe it's worth for somebody else to have a look.


dstockhammer:
TaskListUI doesn't build from CLI after uninstalling VS2015. We should probably either remove the examples, or upgrade all of them to net core/standard.

ianbattersby:
+1 for xunit. Getting the test output right is key but otherwise it's a big win imho.

dstockhammer:
Getting closer I think. There are now two main issues with the tests:
* They break with parallel execution. This seems like a valid issue, because there's a lot of static stuff going on that just doesn't work when executed in parallel.
* They hang after completion. Pretty sure that's because some parts of the code aren't "async all the way down" and deadlocks happen.

I already removed Nitro.AsyncEx in an attempt to fix the second issue, but it's still happening. Will need to dig deeper and find every potentially blocking call :S

dstockhammer:
Alright, tests now run smoothly. I've ignored a couple of tests because they either can't run in parallel or fail on AppVeyor; but I think we can sort those out once we're happy with the rest of the upgrade.

Here's what I did:
* Upgraded to VS2017 / msbuild
* Converted all tests to xunit
* Converted all assertions to FluentAssertions
* Updated AppVeyor build script to work with csproj
* Removed MessageViewer and TaskList samples. Will move them to contrib or separate repo.
* Upgraded packages where appropriate
* Samples target `netcoreapp1.1` where appropriate
* Cleaned up unused imports, files, etc

@iancooper What do you think?

iancooper:
Thanks Dan. Agreed with the async tests, nitro was a workaround for MSpec. We should figure out where the hang is. I get a few 'test has finished but is still running' warnings in the IDE.

PR #223, opened 2017-04-12
Convert files to UTF-8 and fix encoding issues

dstockhammer:
I noticed that some files are encoded in various Windows char sets or UTF-8 with BOM, which is not recommended. Especially the Windows char sets cause problems on other OSes and, perhaps most importantly, cause merge conflicts and can't be viewed with GitHub's diff viewer.

While I was there, I also fixed broken characters in the MIT license header.

The third change is perhaps a bit controversial, but I noticed that there is another file header that is horribly out of date and has wrong dates, file names and assembly names. I removed that.


dstockhammer:
Works fine in Rider 👍

PR #233, opened 2017-06-27
Fixing broken docs links

shanselman:


iancooper:
Thanks for this Scott

PR #241, opened 2017-06-29
Fix broken link

JosephWoodward:


iancooper:
Thanks. I wonder if I should update that image as well. Hmmm.

JosephWoodward:
@iancooper With the one used in the [stickers you recently had printed](https://pbs.twimg.com/media/DDWDPTGXsAAm-5E.jpg)? I think so, it looks loads better. And Don't forget to update the Twitter profile picture too.

PR #244, opened 2017-07-08
fully async dispatcher / message pump

dstockhammer:
As discussed in gitter. This is a prerequisite for proper Azure Service Bus support.

This PR is WIP, please don't merge. I mainly want to gather some feedback and see how the tests run on AppVeyor.


iancooper:
Trying to review today

iancooper:
Lots of questions:

(1) Why kill ConnectionBuilder with a progressive interface? There does not seem to be benefit to that
(2) ConnectionBase and the two derived types for Connection. I get that we should use polymorphism to represent a type difference between async and non-async over testing a property, but hopefully this didn't necessitate killing above?
(3) Open did not previously depend on being Shut, as it allowed you to inject a new consumer into a running service activator. I might be persuaded on this, partially because it is really hard to do this in Python where we rely on multi-processing over threads - but at that point you could just drop open as a public interface and rely on receive. The downside is that at runtime you can no longer send a message to the activator via the control bus to get it to adjust what it is running, which as a nice feature.
(4) Why the switch in ConsumerFactory?
(5) I'm a bit unsure why we need to introduce cancellationtoken in the dispatcher. More on reasoning here would help me. We use an STA message pump where the convention is to add a message to the queue. This works well across the board, including in Python which is multi-process. I don't really want to change this for a cancellation token. Sure that is used in TPL but the message pump paradigm is more important here than a cancellation token.
(6) Overall I don't really want the global cancellation token change. See how if flows out everywhere. This is a fundamental change away from the message pump philosphpy that signals via a queue. Signalling via a queue is old school and very effective. Cancellation Token is not the way to go here

dstockhammer:
1+2) I agree that this is weird, more work is required. I dropped the builder for now because I expect bigger changes to channel and keeping it sync was a pain. Will bring the builder back once channel is finalized.
3) Fair point, I didn't quite realize that that's how it's supposed to work. Will review.
4) Ah yes, that's a relic of an earlier version that had a separate async factory. Forgot about this, will remove.
5+6) This is an interesting point. My assumption was that you opted for quit message because there was no better way of doing it at the time. My thinking was that `CancellationToken` is the idiomatic way of doing this type of thing in c#. The benefit of a cancellation token is that it reliably stops every thread without having to wait for a timeout. I don't quite understand why a quit message would be preferable, maybe we can chat about this in gitter or some other more direct channel.

iancooper:
Sure. On 5+6 we are attempting graceful shutdown by processing work in flight before shut down. This means what you want is to insert the 'next work item' as close. The message pump approach helps you do this.

Yes, this means you have to wait for shutdown, but this is a standard expectation I think, that you will close out existing work. Nginx for example uses the same approach, ru to end on existing jobs on a shutdown request.

Cancellation tokens are great for 'kill this now' because it signals all the way down the pipe, but I find that can be a little bit viral, in that everywhere in your pipe has to keep checking for this. If you want 'kill this now' you can just exit the process. Now that is disorderly shut down without clean up, but most use cases for cancellation token I have seen have long running work which requires polling the token anyway,  usually 'slow' jobs that are I/O bound. There is not much actual relief from a cancellation token here.

However, if we wanted pre-emptive kill and nack to queue without processing, I would use a slightly different signalling approach to send a message to all running tasks. I think that is a seperate conversation.

iancooper:
I'm going to close, and see if I can get you to resubmit after a chat

PR #252, opened 2017-08-12
Removed the allocation required to create an async handler

Drawaes:
Any optmisations (typeof(generictype)) will be optmised by the JITTER. This removes the need to do

var handler = new AsyncHandlerFactor<t>().Createxxxx

Which is a wasted allocation, seeing as the AsyncHandlerFactory class could never be changed without a code change anyway


iancooper:
Thanks for that!

PR #253, opened 2017-08-12
Moved common project properties to a common.props file for easier solution management

Drawaes:
Made a build folder
Made a common.props
Removed all common project properties and referenced common.props


iancooper:
Ta

PR #254, opened 2017-08-12
Remove unnecessary parameter.

vgaltes:
The input channel factories were receiving a message producer factory parameter that was not used. This PR eliminates the parameter and fixes all the references.


iancooper:
Thanks, good PR!

Issue #257, opened 2017-08-17
Remove "Mixed Platforms" solution platforms from Brighter.sln


whunsley:
The command:

```
dotnet sln add Brighter.sln <path_to_my_new_solution>
```

Will add the project to the solution with Debug and Release builds for the following platforms Any CPU, x64, x86.

When you run the command you get the warning.
```
dotnet restore
warning MSB4121: The project configuration for project "MyNewProject" was not specified in the solution file for the solution configuration "Debug|Mixed Platforms".
```

Removing the "Mixed Platforms" from the Brighter.sln for all projects will stop the warning and make the solution more usable with the dotnet core command line tools.  It should have no impact on the current builds as this build configuration is not being used.

PR #258, opened 2017-08-17
Remove "Mixed Platforms" solution platforms from Brighter.sln

whunsley:
Fix for issue #257 to make the Brighter.sln
work better with the dotnet core command
line tools.


iancooper:
Thanks

PR #264, opened 2017-09-29
Removed reference to TODO in HelloWorld example

brainwipe:
The Handler Factory that's given in Step 3 already has some code for creating the Command Handler. We don't need to put in the non-production creation in Step 4.


iancooper:
Thanks

PR #276, opened 2018-03-13
Updated GreetingsExample to fix small error in code example

DevJonny:
Just two lines in the GreetingsExample documentation that don't match the the code equivalent and prevent the example compiling.


holytshirt:
👍  awesome thanks man

PR #285, opened 2018-04-24
fix missing async support interface

Red-F:
The IAmAMessageProducerAsync was missing from MsSqlMessageProducer
cast added to samples to remove ambiguity from MessageConfiguration constructor


holytshirt:
All good

PR #295, opened 2018-07-17
Add Exists() method to CommandStore, use that in the CommandSourcingH…

tapmantwo:
…andler, meaning no need to load the entity and require the default ctor. (That is still required for reading from the store - I aim to submit another PR that allows the reading/writing to the store to use the registered message mappers).


tapmantwo:
Please review the Async code - I took what was there as a template...

PR #310, opened 2018-10-04
Fix kafka consumer configuration

a2glukhov:
Due to http://kafka.apache.org/documentation/ auto.offset.reset consumer configuration section it has other valid values


iancooper:
Thanks for helping out, I'll add you to our contributors list.

PR #311, opened 2018-10-09
Once per handler option for command store

tapmantwo:
At present, the command store writes to the db when a command is first handled, and checks for duplications by checking for the existence of a command of a particular type with a particular id.

This doesn't work for scenarios where there are many handlers for the same command type, particularly if they are in different procesess.

I've introduced the concept of a 'context key' for command sourcing, allow you to specify a way to identify the particular context in which a command is being handled.  I anticipate that you would normally set this to the name (or indeed, type) of the handler, but it's flexible enough for you to decide on some other form of grouping if you want.  It will then whether a particular command has been handled already in a specific context, allowing multiple handlers to work with the command store.


tapmantwo:
Not sure about the dynamo tests - they seem to pass locally.

iancooper:
Thanks

PR #313, opened 2018-10-11
Remove new() constraint from commandstore GetAsync and GetAsync method

tapmantwo:
Change the commandstore Get and GetAsync methods to throw an exception if the item doesn't exist (you can check first with the Exist method).  The means the new() constraint can …be removed.


iancooper:
Thanks Richard

PR #314, opened 2018-10-23
Change order of the command store handler so that it only stores if t…

tapmantwo:
…he command is handled succesfully.


iancooper:
Thanks @tapmantwo

Issue #312, opened 2018-10-10
RMQ Timeout is aggressive at 5s, and should be configurable, we should back-off more on timeout


iancooper:
We have an aggressive timeout of 5s set for the required heartbeat timeout for RMQ. The RMQ team recommend a number between 5 and 20s. We may wish to lower this, to reduce noise. 

We currently set this in RMQMessageGatewayConnectionPool. We should continue to support defaults, if the user does not set them, but we ought to allow passing values in from outside via RMQMessagingGatewayConnection

We reconnect on a timeout, but we do not delay before reconnecting. The issue here is that we might be putting an already under-strain node under further strain, by connecting to it again, without a breather. This can be exacerbated by the 'Thundering Herd' problem, as if we disconnect a group of clients, they may all try to reconnect immediately.  We should probably introduce a short, randomized delay to stagger re connection attempts (in RmqMessageConsumer Receive).

PR #345, opened 2019-01-29
#312 | Adds configurable Connection Heartbeat for RMQ Timeout

DevJonny:
* Adds configurable heartbeat for RMQ heartbeat
* Defaults this to 5 seconds


iancooper:
Thanks @DevJonny

PR #354, opened 2019-02-12
Changes OnceOnly from DEBUG and throw exception to WARN with no exception

DevJonny:
Following changes have been made:

* Updates OnceOnlyAttribute tests to check for absence of exception
* Removes throwing an exception when OnceOnly: true is set, and a message is seen more than once
* Changes the DEBUG statement for this case to be a WARN statement
* Removes OnceOnlyException class as it's no longer used

@iancooper / @holytshirt: this is a little more drastic than I had originally thought it would be. The error logs (and in our case Error emails that TO have asked us about) are triggered by the exception being thrown rather than a log statement. What are your thoughts?


DevJonny:
@holytshirt and I have just discussed this. As the original idea was throw an exception and let the client deal with how they handle that exception.

Do we keep that approach or do we handle the exception inside of Brighter?

PR #403, opened 2019-03-13
#312 | Sets random delay on reconnect to RMQ

DevJonny:
Added a random (between 5 and 100 ms) delay when reconnecting to RMQ.

@iancooper I wasn't sure if this was the right place to add the delay or not? Is the 100 ms too long, is 5 ms too short?


iancooper:
A good place to do it, but you need a .Wait() on that Task Delay or it won't do anything ;-)

iancooper:
PS it's possible that failing test is falling afoul of the new attribute cache in the pipelinebuilder and you will need to call the clear cache method in your test setup. not sure.

DevJonny:
Doh! Not the first time I've missed .Wait(), thanks @iancooper! I'll push an update. I'll catch up with Toby RE: failing test.

iancooper:
I know, I miss Wait() all the time. It's the reason I look for it

Issue #485, opened 2019-05-08
RabbitMQ unavailability is causing Dispatcher to stop


acraven:
If RabbitMQ is unavailable for a period longer than the cumulative retry duration the Dispatcher will stop (and will need to be restarted).

An exception is being thrown in `RMQMessageGatewayConnectionPool.ResetConnection` when the connection does not exist in the connection pool. This causes the message pump task to end, which in turn causes the Dispatcher to stop when all tasks have been completed.

Given `TryRemoveConnection` can be called with or without the connection existing, it suggests the `var connection = s_connectionPool[connectionId];` in `ResetConnection` is redundant. And since `CreateConnection` calls `TryRemoveConnection`, the `TryRemoveConnection` and `GetConnectionId` are redundant too.

acraven:
Having tested this change with our service, it didn't solve the problem; the message pump simply stopped due to another unexpected exception.

My proposed fix (https://github.com/BrighterCommand/Brighter/pull/493) is to remove the CreateConnection from the ResetConnection method and simply remove the connection if it exists. The next iteration in the message pump re-connects to Rabbit.

The ResetConnection method is indirectly used in about 4 exception handlers; it throwing a connection related exception results in the real exception not being wrapped in ChannelFailureException causing the message pump loop to end.

This fix allows RabbitMQ to be stopped for a lengthy period without the dispatcher stopping.

iancooper:
Right, but ResetConnection was added to resolve a specific set of errors that happen during a partition in RMQ which we uncovered by doing some chaos engineering with blockade. So removing that call would need us to to test for why we added CreateConnection there (was it just to refill the pool?) It might be better to leave the call, but ignore failure to connect on a ResetConnection, perhaps after a window of normal retries.

So I think we need a more comprehensive approach to addressing the issue including running blockade tests again on anything that impacts the core recovery loop.

Not saying we should not do it, but its 'handle with care' around that code

iancooper:
@acraven Can you try this? Modify the CreateConnection method so that it catches and logs any exceptions on attempting to recreate the connection over allowing the exception to bubble out of the method? That way, if RMQ is unavailable we won't exit with an unhandled exception but just run the loop again, but we will still recreate the connection if RMQ is available. Can you try that, test it, and let me know what happens? (PS Prefer to catch the specific exception you are seeing, not all exceptions from CreateConnection please)

acraven:
[...]

PR #493, opened 2019-05-09
Remove re-connection from RMQ ResetConnection

acraven:
This relates to issue https://github.com/BrighterCommand/Brighter/issues/485


iancooper:
Let's discuss under the issue first please

PR #682, opened 2019-08-13
Updating EventStore to use new Outbox APIs

DevJonny:
- Updates existing EventStore Outbox tests to use EventStore container instead of in-memory node thingy


DevJonny:
Rolled changes into PR #683

PR #743, opened 2019-09-12
Kafka improvements

iancooper:
Somes fixes for Kafka issues found so far in testing


src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageConsumer.cs
```diff
@@ -76,8 +81,10 @@ internal class KafkaMessageConsumer : IAmAMessageConsumer
                         _logger.Value.Error($"ConsumeError: Member Id: {consumer.MemberId}, error: {error}");
                 })
                 .Build();
+            
+            _logger.Value.InfoFormat($"Kakfa consumer subscribing to {_topic}");
```
holytshirt: Using string interpolation in the logger means the string is always created and _topic.ToString() will be called. If you do 

 _logger.Value.InfoFormat($"Kakfa consumer subscribing to {Topic}", _topic);

the .ToString() call is deferred on only called when needed. If your logging is set to a higher level then it will not build the string

PR #834, opened 2019-10-23
Event Store Outbox implementation

DevJonny:
It would be good to get some feedback on this PR from someone who knows Event Store better than I do. I've had to do some relatively janky stuff due to it being append only. Mainly around "MarkDispatched" and "OutstandingMessages" as I've had to add a second event to stream to mark it as dispatched, but with the previous Event Id in the metadata.

Also the GetById methods are not implemented due to the need to read the entire stream to find a message? Is this something we should support with the Event Store Outbox? I would be relatively quick to implement.


holytshirt:
@DevJonny I can help you rebase this tomorrow and move it to master

PR #840, opened 2019-10-25
Split tests

iancooper:
Break the tests up by the infrastructure that the are dependent on, to make the tests easier to use i.e. you don't have to have all the infrastructure running


holytshirt:
Thanks @iancooper, this will make the build much easier

PR #895, opened 2019-11-21
SQS Gateway tests changes for running tests on Azure pipelines

holytshirt:


iancooper:
Makes sense to me, go ahead and merge. And thanks for doing that

PR #931, opened 2019-12-10
Remove old async message pump that tries to use STA approach; to be r…

iancooper:
This merges our message pump code back into one, dropping the async STA message pump. The async message pump is probably best not simply derived from the STA one for now, as the callback deadlocks with the pump.

The goal is to follow this up with some additional implementations of IAmAMessagePump that have different behaviours to STA and can be 'opted' for


iancooper:
Merged with one approver in Toby's absence

PR #998, opened 2020-01-15
issue-4:

homcl:
Release notes updated with fact that Brighter now subscribes to blocked and unblocked RabbitMQ connection events and that it generates a log entry when either event occurs.


iancooper:
Merged as trivial.

Thanks @homcl for doing that, much appreciated

PR #1026, opened 2020-02-03
Merge extensions into main brighter project

holytshirt:
This is for #1027

Added some colour to the one of the samples
image


holytshirt:
Done, all in master

PR #1032, opened 2020-02-05
Implement EnableAutoCommit to lower load on the Kafka cluster

timvw74:
This PR enables the ability to commit offsets after a set amount of time instead of after every message.  This will lower the traffic on the Kafka cluster as every acknowledgment is a new message on the __consumer_offsets topic.  The downside is that if a consumer goes down before committing it's offset then when the partition is assigned to a new consumer that consumer will re-process all the messages that haven't been acknowledged.  This will not be an issue if messages are idempotent.


iancooper:
Thanks @timvw74 I'll prompt @DevJonny or @holytshirt to take a peek too

iancooper:
Thanks @timvw74 I'll add you into our contributor list.

PR #1055, opened 2020-02-26
issue-303: Removes code that can potentially cause key not found error

homcl:
The problem doesn't seem to be that a message is not resent on connection reset. There is code for that. The problem seems to be that a 'key not found' error (when trying to get a connection using a non existent connectionId key) is thrown. This ' key not found'exception is handled as a general exception in the MessagePump. The general exception doesn't involve re-sending the message. The RMQMessageGatewayConnectionPool now handles the potential of 'key not found'.


iancooper:
Thanks @homcl Can you add something to the release notes to indicate that this is issue is fixed

PR #1094, opened 2020-04-03
Fix typo

hiddenboox:
Hi, after some searching during my POC I just found small typo.


iancooper:
Thanks for this

hiddenboox:
I am pleasure to commit even small things like this

PR #1151, opened 2020-05-20
Update for RMQ 6.0; fix a Kafka deprecation warning

iancooper:
EDITED For clarity

This PR deals with package upgrades that required code changes
 - It fixes a Kakfa deprecation warning
- It fixes RMQ client switching to ReadOnlyMemory<byte>. (detail below)
- It also update the DependencyInjection package) to ensure it was not built against out-of-date packages

DETAIL ON RMQ PACKAGE CHANGE
-------------------------------------

- Because we have to copy the body before returning from PullConsumer, and we buffer messages, we have to copy the supplied body
- Because we are netstandard20, and not nestandard21 we have to copy into our MessageBody as we don't have Span<T> overloads for classes like Encoding
- So we now have two allocations, although RMQ has a pool, which really makes the situation worse :-(
- Advice is wait until .NET50, don't go to nestandard21 as breaking changes, so we may be able to remove one allocation then.
- To be fair, we could probably do a lot to reduce our allocations and that is a project in itself, so I am not sure this is really a crisis for us as we probably have a lot of allocations to deal with

### DIFF
--- a/samples/AWSTaskQueueSamples/Greetings/Greetings.csproj
+++ b/samples/AWSTaskQueueSamples/Greetings/Greetings.csproj
@@ -8,6 +8,7 @@
     <ProjectReference Include="..\..\..\src\Paramore.Brighter\Paramore.Brighter.csproj" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="3.1.4" />
     <PackageReference Include="System.Reflection.TypeExtensions" Version="4.7.0" />
   </ItemGroup>
 </Project>
--- a/samples/AWSTaskQueueSamples/GreetingsPumper/GreetingsPumper.csproj
+++ b/samples/AWSTaskQueueSamples/GreetingsPumper/GreetingsPumper.csproj
@@ -9,4 +9,8 @@
       <ProjectReference Include="..\Greetings\Greetings.csproj" />
     </ItemGroup>
 
+    <ItemGroup>
+      <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="3.1.4" />
+    </ItemGroup>
+
 </Project>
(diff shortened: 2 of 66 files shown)


samples/AWSTaskQueueSamples/Greetings/Greetings.csproj
```diff
@@ -8,6 +8,7 @@
     <ProjectReference Include="..\..\..\src\Paramore.Brighter\Paramore.Brighter.csproj" />
   </ItemGroup>
   <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="3.1.4" />
```
holytshirt, line 11: Did you mean to commit all these DependencyInjection packages?

iancooper: I updated the package dependencies as part of pulling RMQ 6 in, and these got picked up

iancooper: I have also updated the merge to reflect changes to packages that have since occurred via dependabot

iancooper: We can roll that back if it really bothers you, but it would save effort if we just accept we need to upgrade it

iancooper: Another way to look at this.

This PR fixed package upgrades that we normally do via Dependabot. One couldn't be done by merging as it contained breaking changes (RMQ), one was a deprecation issue, which again Dependabot won't resolve; and the other was an effort to make sure that all the packages were up to date, before updating any packages manually

iancooper: Updated scope of PR

holytshirt: dependabot annoys me it's good for notifications, it's the merges are a mess. 

I'll may have to remove when I merge in my branch as they not needed.

otherwise let's rebase this in

PR #1164, opened 2020-06-06
Working idea for Generic Mapper

holytshirt:


iancooper:
@preardon Is this now done and redundant with your work?

preardon:
@iancooper yeah I would say the initial version of this is in

iancooper:
@holytshirt So as @preardon has a version of this in (sure Paul can explain how it works), I think we should close this PR. What do you think?

PR #1248, opened 2020-09-16
Add NuGet package

iancooper:


holytshirt:
You don't need to do `dotnet pack ...` the nuget package is built as part of the build which includes source link.

iancooper:
@holytshirt OK, I had not seen that we had enabled source link in our packages, or that we were already creating nuget packages. I'll peek. It still doesn't resolve the issues we seem to have with LibLog and Pack, but it might help me move forward. I'll look at the artefacts

iancooper:
@holytshirt Cool, I can find them in the build directories now. Thanks for the steer. I can park the log stuff for now. Joy.

PR #1394, opened 2021-02-14
Fix No Opt Outbox

iancooper:
This drops the No Opt Outbox altogether, and replaces it with a decent InMemoryOutbox. This also includes a better InMemoryInbox.

Some characteristics:

 - Cache outbox or inbox entries
 - Set a TTL for entries, background thread will sweep to remove
 - Set a max size, background thread will compact
 
These are usable in production, with the usual in-memory limitations: outbox will lose its contents on shutdown risking that messages are not sent; inbox is not shared across processes so will only protect if used with multiple peformers not multiple processes.

But most importantly, you can use this without worrying that a in-memory box will eat up all of your memory and then crash

Also includes a sweeper implementation, works for any outbox, but tested against the in memory outbox. It shows how to implement. May add to the Kafka branch samples to show how to use it with hosted services.


src/Paramore.Brighter/InMemoryInbox.cs
```diff
@@ -104,22 +180,26 @@ public class InMemoryInbox : IAmAnInbox, IAmAnInboxAsync
...
+            
             if (!Exists<T>(id, contextKey))
             {
                throw new RequestNotFoundException<T>(id);
             }
 
-            var inboxItem = _commands[CreateKey(id, contextKey)];
-            if (inboxItem.CommandType != typeof (T))
-                throw new TypeLoadException(string.Format($"The type of item {id} is {inboxItem.CommandType.Name} not {typeof(T).Name}"));
+            var inboxItem = _requests[InboxItem.CreateKey(id, contextKey)];
```
holytshirt, line 190: Probably use _requests.TryGet() so you can handle found/not found in one operation?

holytshirt (approved):
Looking good from a quick review,

src/Paramore.Brighter/InMemoryInbox.cs
iancooper: I'll peek at it. Thanks

iancooper: OK, implemented that, thanks for the suggestion

iancooper:
@holytshirt Made that fix, can you check the PR again :)

PR #1401, opened 2021-02-15
fix: Store/Retrieve missing fields to/from the Outbox

iancooper:
The optional headers were omitted from the Outbox work. This is a problem as the message data is being truncated.

This fix is about storing and retrieving those missing fields


iancooper:
@holytshirt Marking this as ready for review. Failure on the in-memory tests is just a timing issue with threads being scheduled for the outbox clean up I believe. Can ignore for review, as has not changed in this code.

holytshirt (approved):

PR #1407, opened 2021-02-17
Improve creation of messaging infrastructure

iancooper:
Our existing behavior is for consumers and producers to declare what they need; the option here is to pass in a flag, in case you are working with infra as code (such as generated from AsyncAPI) to do this. In that case we will just validate that the required infrastructure exists


iancooper:
Kafka does auto-creation of a topic when you send to it, and doesn't really have a 'queue' so there is less infrastructure. There is an admin client, which would allow test for and create as an explicit step. I'm going to add that into the Kafka update, not this PR.

iancooper:
From the release notes:

- Changes to how we configure transports - renaming classes and extending their functionality
 --- Connection is renamed to Subscription
 --- Added a matching Publication for producers
 --- Base class includes the attributes that Brighter Core (Brighter & ServiceActivator) need
 --- Derived classes contain transport specific details
 --- On SQSConnection, renamed VisibilityTimeout to LockTimeout to more generically describe its purpose 
 --- Seperated from GatewayConfiguration, that now has a marker interface, used to connect to the Gateway and not about how we publish or subscribe
 --- We now have the option to declare infastructure separately and Validate or Assume it exists, still have an option to Create which is the default
 --- We think it will be most useful for environments like AWS where there is a price to checking (HTTP call, and often looping through results)  
 --- Added support for a range of parameters that we did not have before such as dead letter queues, security etc via these platform specific configuration files  
 - Provided a short form of the BrighterMessaging constructor, that queries object provided for async versions of interfaces

holytshirt (approved):
Lets try bring it in

holytshirt:
@iancooper  was going to bring it in, I updated the branch from master. Which had updates for the CI to build using dotnet 5 and also install the 3.1 sdk to run the tests.

2 unit tests are failing. Can try have a look tomorrow

iancooper:
Merged the PR, because the tests that were failing because of the merge issues around renaming a field now pass, and other tests appear to be erratic problems with EventStore and AWS tests.

PR #1421, opened 2021-03-05
Put Async Pipelines Back

iancooper:
This PR is intended to support us having async dispatcher pipelines again. Obviously, these need documentation because they have side efffects developers may not understand:

- Requests lose guarantee of processing "in order"

There is a lot to consider here:

- Do we want threadpool based callbacks? 
-- This may be the easiest to get 'working'
-- We now have an issue with backpressure, because the async dispatcher will keep eating work from the queue whilst it is waiting for non-blocking I/O to complete, which could end up with a lot of work queued if we support using the threadpool for callbacks.
-- Leads to "max in flight" solutions that increment and decrement a counter for number of messages before we pause reading the queue
-Alternatively, do we want a single threaded event loop
-- Easier to apply backpressure - only allow so many pending events in our queue
-- Even without that, biggest issue becomes queue length, not context switching.
-- Thread affinity fixes some issues, such as HttpContext
-- Won't get you the same throughput, because the continuation will await the next message hitting non-blocking I/O or finishing with that message

Our preference is for the latter model, because it should be simpler to reason about, but it requires a custom synchronization context. This was causing a deadlock -literally freeze machine - issue that was hard to debug


iancooper:
So I think this may get async pipelines working, running continuations on the same thread, but there is a fair amount of testing to do still. However, @preardon if you wanted to play by flipping the runAsync flag on your subscription to true and see if it explodes, its a starting point.

@holytshirt just FYI for now

iancooper:
Closed as PR out of date

PR #1427, opened 2021-03-10
Fix some issues with AWS (failing tests and TopicARN validation)

iancooper:


iancooper:
WiP as I need to add a test around the TopicArn stuff

iancooper:
It's possible that this one, I accidentally branched from, Kakfa. So I have left this as a draft behind that commit as it is fixes only

holytshirt:
> It's possible that this one, I accidentally branched from, Kakfa. So I have left this as a draft behind that commit as it is fixes only

Yes you did

iancooper:
It also seems to possess whatever screw up adding those env variables to connect to Confluent creates for the Kafka tests

Issue #1444, opened 2021-03-18
ASB - allow creating of topics/subscriptions to be optional


penderi:
**Is your feature request related to a problem? Please describe.**

Cloud hosting typically leaves infra concerns (topuic/subscription creation) to be outside of the scope of a component or service. Moreover reduced-rights deployment means attempts to create will fail.

Add the ability to ASB transport to only create topics and subscriptions when specified. Default = allow (as current)

mentioned in #1438

preardon:
@penderi @iancooper I'll have a PR in for this in ~30 minutes

iancooper:
@preardon Guessing you have seen the optional creation enum on the V9 publication and subscription?

preardon:
@iancooper I have not, I was just adding creation Booleans to AzureServiceBusConfiguration as I figured I would keep this confined to ASB Config. I can have a look to wire it up to the over arching config if you like

iancooper:
@preardon So on Subscription https://github.com/BrighterCommand/Brighter/blob/master/src/Paramore.Brighter/Subscription.cs and Publication https://github.com/BrighterCommand/Brighter/blob/master/src/Paramore.Brighter/Publication.cs we have an enum OnMissingChannels that indicates what you want us to do:

- Assume - don't validate or create, it's there trust me (low latency)
- Validate - check its there, if not raise a channel failure
- Create - create it

We assume for create producer creates channel (topic, routing key) but client creates queues etc.

That gets passed into your producer or consumer, and the code there has to do the work to determine what to do. When may vary. So with RMQ because it is a no-op if it exists and low-latency the Channel checks when it is called, but in AWS the ChannelFactory does it, not the Channel, because creation is high latency and we need to check if something exists before we create it.

So you have to think about where as well. I'm guessing that ASB is a REST API for creation, so it might make sense to look at the ChannelFactory in the same way AWS does

iancooper:
+ @penderi and @preardon Please make sure you are talking so we don't have two folks trying to fix the same issue :-D

penderi:
@iancooper don't worry @preardon is a mate at Flagstone, all good.

holytshirt:
Let me know if you happy to close this now @iancooper

iancooper:
@holytshirt Going to look shortly

iancooper:
@holytshirt I just wanted to check on the following. Forwarding my comments from the PR:

[...]

PR #1446, opened 2021-03-18
Plumbed OnMissingChannel through to ASB #1444

preardon:
As per #1444 The Azure Service Bus Library now honours the MakeChannel property on both Subscriptions and Publications. Also set Default Behaviour to Create (as per previous Logic)


holytshirt (approved):
Looks good to me

holytshirt:
Thanks so much @preardon!

iancooper:
@preardon So the only thing that might be missing is that the consumer doesn't check for the topic and create it if asked to create. I don't know enough about the management wrapper to know if that will automatically check for the topic and create it, if it is missing, but it would probably need that if not. The consumer tends to check for everything - the producer only the topic.

Issue #1447, opened 2021-03-19
AWS SQS - Unable to parse messages from non-Brighter services


honkuan86:
### Describe the bug
When we are using Brighter to exchange messages with services from different technology stacks (Java, etc) that are not using Brighter as message producer in AWS SQS, Brighter failed to parsed the message and throws error, because the Message Attributes that Brighter expecting is not provided.

### To Reproduce
Publish message to queue without using Brighter, then process the message with Subscriber built with Brighter.

### Exceptions (if any)
MessagePump: Failed to parse a message from the incoming message with id 00000000-0000-0000-0000-000000000000 from event-queue on thread # 10

### Further technical details
- Brighter version: v8.1.1399

PR #1448, opened 2021-03-19
Fix message parse failed issue with AWS SQS

honkuan86:
This PR is related to #1447 .

The changes is at _SqsMessageCreator_ to default the _HeaderResult.Success_ to true for all headers, so message can be parsed successfully when SNS message attributes are not present.


iancooper:
Thanks for this @honkuan86 . I'll skip worrying about the failing CI tests because it looks environmental. This change makes sense as it follows our tolerant reader mantra of trying to make it work even if you don't send us everything we need.

Can you sign the CLA? Then we are good to merge this

honkuan86:
> Thanks for this @honkuan86 . I'll skip worrying about the failing CI tests because it looks environmental. This change makes sense as it follows our tolerant reader mantra of trying to make it work even if you don't send us everything we need.
> 
> Can you sign the CLA? Then we are good to merge this

Hi @iancooper , I have signed the CLA. Thanks.

iancooper (approved):

holytshirt:
@honkuan86 thank you for the PR!

PR #1449, opened 2021-03-19
Check if Topic Exists before creation Subscription #1444

preardon:
Ensure that Topic is created before attempting to create Subscription


iancooper (approved):

iancooper:
Thanks for that @preardon

PR #1467, opened 2021-03-29
Redis Streams

### LINKED ISSUE #1185: eSwitch to Redis Streams and StackExchange Redis
ServiceStack.Redis is only free up to a limited number of API calls per hour, after that it is a paid product. We use ServiceStack.Redis because our implementation of queues over Redis requires use of a Blocking Pop operation which cannot work on a multi-plexed channel over a single threaded database like Redis. This creates an issue for high-volume users, they are forced to pay for ServiceStack

Since we wrote our Redis support, Redis has released support for Streams. Streams are a more natural building point for queues than the list based solution we have used up to now. If we switch to streams we can also switch the client to StackExchange.Redis, which is free.

We could drop back to using an earlier version of ServiceStack, before it became a paid for product, but this exposes us to a range of new issues.

**Additional context**
None

iancooper:
The existing notification-based queues work using the ServiceStack client, which is not free at scale. Redis Streams provide a good alternative to this that we can use. It is a streams model, similar to Kafka but we can support that in a producer-consumer model


iancooper:
I am going to close this PR, as are going to use this as an test case for auto-generation

Issue #1457, opened 2021-03-22
Replace Newtonsoft.Json with System.Text.Json


preardon:
Update of Dependencies

Replace the use of Newtonsoft.Json to use the newer System.Text.Json in Brighter

Performance Gains ( https://devblogs.microsoft.com/dotnet/whats-next-for-system-text-json/ ) I'm aware this link is old, however I believe its only gotten better and know that there have been a few extra features added as part of 5.0

System.Text.Json
https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to?pivots=dotnet-5-0

penderi:
I hadn't even thogut of this, but I like it!

preardon:
@penderi you know me, I just see Newtonsoft.Json and want to rip it out and replace it with System.Text.Json

Looks like its only really used for the In Memory Inbox, but if this table is to be believed
![image](https://user-images.githubusercontent.com/7712876/112006040-400e7800-8b1b-11eb-9545-77544ab3a87a.png)

(and its even an iteration old) then 2x desterilisation and 1.5x serialisation could be worth it for the effort required

iancooper:
@preardon it should definitely go on our list, but perhaps post v9 at this point. We do have some older dependencies, such as LibLog and this one that need to go, but I suspect we will only get to LibLog if we want to "call" this release as done. Further reducing the set of our dependencies outside the BCL for the core libraries would also be good. If we end up with just Polly for that, it would be a huge win.

iancooper:
That said @preardon feel free to do the work, and we can see when it lands.

preardon:
@iancooper I'll take a look today from memory you were only using it in one or two places, if it isn't much I'll get a PR in for it

iancooper:
@preardon it's in a few places on Brighter, where we want to serialize, such as monitoring, as well as the Outbox/Inbox. You will also find it's used in transports and a lot in the tests. Worth doing, but a bit of a chore. Thanks for picking it up.

preardon:
@iancooper yeah this went a bit more than I original thought ~111 file PR will be inbound after some extra testing)

On another note, do you have strong feelings about Property Name Casing in Json, I tend to prefer Camel for json and Pascal for Dotnet. I've just run into some tests where I have to make a decision?

holytshirt:
I prefer Camel for Json and Pascal for Dotnet too. 
[...]

PR #1470, opened 2021-03-29
Replaced NewtonSoft.Json with System.Text.Json #1457

preardon:
Replaced NewtonSoft with System.Text

Appear to have intermittent issues with the In Memory inbox/outbox and SQLite on my machine. Lets see how the Unit Tests run in your Pipeline


preardon:
Linking to #1470

iancooper:
@preardon The tests can be intermittent because there is threading involved for the checks of size and expiry. Timing issues may mean they don't run, even with a delay

iancooper:
@preardon Thanks for this, I'm going to accept it. It will be useful to kill that dependency for folks.

PR #1471, opened 2021-03-29
Added a Json Number -> String Converter

preardon:
Added a Number As String Json converter to fix an issue in the Redis Messaging Gateway (deserilisation to Dictionary<string, string> where some of the values were not strings)


iancooper:
@preardon Thanks for picking this up. To save me some effort, and seeing as you are banging them out, update your achievements in our contributors file as you tackle stuff. Thanks for all the help

PR #1473, opened 2021-03-31
Only run the Kafka tests locally to see if issues are Confluent

iancooper:
We are timing out on the tests, use the XUnit traits filter to exclude the Confluent tests


iancooper:
Going to merge ahead of @holytshirt build changes to at least get to green

PR #1487, opened 2021-04-17
fix: clean up of dispose and delays in Dispatcher and Dispatcher tests

holytshirt:
- On Start, wait until dispatcher is running before continuing
- Remove delays from tests
- Use more async-y tests


iancooper (approved):
Thanks for cleaning these up Toby

Issue #1254, opened 2020-09-22
Add support for structured logging


vmartos:
Hello,

We are using Brighter for the message bus in our application, we have centralized logging using elasticsearch + kibana. We saw that in certain fields there are type conflicts (long and string with the same field), all of them are related with Paramore.Brighter.* logs.

![image](https://user-images.githubusercontent.com/10826413/93856255-074f2c80-fcb9-11ea-8893-5a99d6a1f373.png)

My feature request is to use structured logging with the defacto standard Microsoft.Extensions.Logging.Abstractions, instead of LibLog and change the logs in a structured way (don't use {0}, {1}, etc logs), so there should not be problems with the same field having different types.

I already see there are intentions of removing LibLog: https://github.com/BrighterCommand/Brighter/issues/242

One example of structured logging could be:

instead of:
`_logger.Value.WarnFormat("RMQMessagingGateway: Connection to {0} blocked. Reason: {1}", 
                Connection.AmpqUri.GetSanitizedUri(), args.Reason);`
use:
`logger.LogInformation ("RMQMessagingGateway: Connection to {uri} blocked. Reason: {reason}", Connection.AmpqUri.GetSanitizedUri(), args.Reason);`

so basically is removing the {0}, {1}, etc in logs and substitue with fields, and try to use the same fields for the same data type.

If structured logging is not feasible, because some users do not use structured logging, maybe string interpolation could be a better option:

`_logger.Value.Warn($"RMQMessagingGateway: Connection to {Connection.AmpqUri.GetSanitizedUri()} blocked. Reason: {args.Reason}");`

I hope everything is clear, if you tackle https://github.com/BrighterCommand/Brighter/issues/242 please take in mind to also tackle this to have structured logging.

Thanks for this great library!

iancooper:
We already have a discussion of the need to remove LibLog, but this is valuable input. We should continue the discussion under #

iancooper:
Hey @holytshirt this is worth thinking about when we tackle replacing LibLog in #242

preardon:
@iancooper @holytshirt I will have a PR in for this soon, one thing I would like to discuss are the Fields

At the minute I have used the following

- Id
- CommandType
- EventType
- HandlerType
- RequestType
- Type
- HandlerCount
- HandlerName
- NextHandler
- HandlerTiming
- Topic
- ChannelName
- Request
- OutstandingMessageCount
- SecondsSinceLastCheck
- MessageCheckTime
[...]

Issue #242, opened 2017-06-29
Remove LibLog


iancooper:
Since we switched to XUnit for testing we have hit issues with tests that check on log output. I suspect that this is because LibLog uses statics, which essentially mean that we have a 'shared fixture' and that is going to break with the use of threads running multiple tests that may all try to initialize and clear settings to the logger.

An answer may be to look at initialization for the test suite with XUnit (and tear down) to remove timing issues.

Usually any CI issue can be fixed by re-running but it can be aggravating on a PR not to know if a merge will build, because we had a failing test caused by a parallelism issue.

![image](https://user-images.githubusercontent.com/45537/27681295-aa074dfa-5cb6-11e7-9db2-8347a18dc70e.png)

iancooper:
LibLog has now been deprecated: https://github.com/damianh/LibLog

"My recommend course of action for library developers is to use Microsoft.Extensions.Logging.Abstractions as it is now the defacto standard logging interface in the .NET ecosystem. For library developers that wish to continue to use LibLog or need to make adjustments for their needs should copy the code into their project (after all, that's just what LibLog did).

Thanks to all contributors and users for the support over the years.

Damian"

Given that we now use HostBuilder, we may as well move towards accepting the use of Microsoft ILogger

iancooper:
We should not the comments in #1254

iancooper:
We should also note #299

iancooper:
When looking at replacement of LibLog, I believe that there are a number of related issues:

(1) We should consider where a framework 'logs'. It may be better for a framework to use TRACE and DEBUG to allow consumers of the framework to diagnose faults, but not log, as logging from the framework may be verbose and unhelpful for highlighting problems in production code. Under this model, any existing logging we have for Information, Debug, or Trace logging levels would be a DEBUG statement and Warn, Error and Fatal would be TRACE. We would recommend against deploying a DEBUG build and only deploy a TRACE build for production issues.

(2) Supporting your logging then becomes only an issue of the request logging handler, and we should look to advice on high performance logging: https://www.stevejgordon.co.uk/high-performance-logging-in-net-core

[...]

PR #1493, opened 2021-04-22
Replace liblog with Microsoft.Extensions.Logging #242

holytshirt:
This is for issue #242 and potentially #1254
LibLog has been deprecated and the new kid in town is to use Microsoft.Extensions.Logging.

As we don't support passing a logger through our app so we have to use a static instance.

```csharp
    public static class ApplicationLogging
    {
        public static ILoggerFactory LoggerFactory { get; set; } = new LoggerFactory();
        public static ILogger CreateLogger<T>() => LoggerFactory.CreateLogger<T>();
    }
```

In DI extension stuff I grab the LoggerFactory from the container 
```csharp
            var loggerFactory = provider.GetService<ILoggerFactory>();
            ApplicationLogging.LoggerFactory = loggerFactory;
```

Have not done all the work to convert the logs to do structured logging for issue #1254


holytshirt:
Hey @preardon here is where I'm up too with moving to `Microsoft.Extensions.Logging` from LibLog.
Please review and let me know what you think.
Cheers!

iancooper (approved):
Looks good. I can't review them all, but this is the agreed approach. Probably best to merge this one in, and see if @preardon can pick up the structured logging for us as separate PR

PR #1494, opened 2021-04-22
Support Publisher Confirms from RMQ 

iancooper:
We did the hard work for this under two other changes:

* Outbox support, particularly the in-memory support so Outbox is always on, which simplifies use of it by producers.
* Kafka producers, which always work by asynchronous confirmation of the message being posted, and therefore tying them to an Outbox to ensure we actually send becomes fairly critical.

At that point it became fairly easy to plug RQM confirms into this model. it just needs to support the right interface and raise the event when it receives confirmation that the message has been published

Closes#275


tests/Paramore.Brighter.RMQ.Tests/MessagingGateway/When_confirming_posting_a_message_via_the_messaging_gateway.cs
```diff
@@ -74,7 +73,7 @@ public RmqMessageProducerConfrmationsSendMessageTests ()
         [Fact]
         public async Task When_confirming_posting_a_message_via_the_messaging_gateway()
         {
-            _messageProducer.Send(_message);
+            await _messageProducer.SendAsync(_message);
```
iancooper: We don't support SendAsync on an RMQ producer as there is no Async method on RMQ. I'll revert then patch a fix to use a taskcompletion source so at least that code path won't throw for now

iancooper:
@holytshirt I fixed that issue around the TaskCompletionSource for SendAsync on the RMQ message producer. Any more actions you would like to see, or move this one into master?

holytshirt (approved):

Issue #1491, opened 2021-04-19
Update MsSql Inbox and Outbox to use Azure.Identity (MSAL)


preardon:
The Current MsSQL inbox and outboxes only current support SQL Authentication and no the newer MSAL Libraries

I would like a config switch to allow the use of Token Based Authentication to the SQL Server (MSAL)

I do no believe that there are any alternate solutions.

preardon:
@iancooper If you're happy with this I'll get it done in the next few days mate

iancooper:
@preardon Very happy mate. Thanks for adding that.

preardon:
Looking at this again, I'm going to do a quick refactor to plumb through Cancellation tokens to all Async Methods

preardon:
@iancooper Is there any Documentation you want be to update with these Configuration Items?

iancooper:
We use ReadTheDocs at the moment, and you can find the source in a the docs directory. It might be worth switching to GitBook at some point. I don't think we have per-Db documentation, we tend to rely on tests and samples, so it depends whether you think it needs it, as someone who uses that platform

PR #1498, opened 2021-04-24
Added Token Based Authentication to MsSQL inbox and outbox #1491

preardon:
@iancooper This should do the trick, I haven't added any MSAL based tests in but I've run all of the SQL Auth Tests, I was able to get a token locally, let me know if you want any MSAL Based tests Added?

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -125,6 +125,7 @@ Brighter contributors (sorted alphabeticaly)
   * Improvements to ASB Transport 
   * Replaced Newtonsoft.Json with System.Text.Json
   * Created Social Media Card
+  * Added Token Based Authentication to MsSQL Inbox and OutBox
 
 **[tmschlot](https://github.com/tmschlot)**
   * Various fixes
--- a/src/Paramore.Brighter.Inbox.MsSql/ConnectionFactories/IMsSqlInboxConnectionFactory.cs
+++ b/src/Paramore.Brighter.Inbox.MsSql/ConnectionFactories/IMsSqlInboxConnectionFactory.cs
@@ -1,0 +1,12 @@
+﻿using System.Data.Common;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Paramore.Brighter.Inbox.MsSql.ConnectionFactories
+{
+    public interface IMsSqlInboxConnectionFactory
+    {
+        DbConnection GetConnection();
+        Task<DbConnection> GetConnectionAsync(CancellationToken cancellationToken = default(CancellationToken));
+    }
+}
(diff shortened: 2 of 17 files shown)


preardon:
@holytshirt Any chance of getting this into v9?

holytshirt:
I'll have a look when I'm back home, if all is good I'll push out another
release.

I need this too for my place

preardon:
@holytshirt I've just use DefaultAzureCredential which isn't what I do at work ( usually I specify ManagedIdentityCredential or sharedTokenCache with the ability to specify account form environmental variables dependant on whether you are developing locally or not) I would be happy to change to this approach if you and @iancooper saw value we could just add a Boolean to specify whether to use the SharedTokenCache or not so that we could have a bit more control locally.

preardon:
I've Updated the with the Above and plumbed through the new Cancellation Token for new Async Methods

holytshirt:
Nice thank you!
Just wondering if we should pass this
```
 (await credential.GetTokenAsync(
                        new TokenRequestContext(new string[1] {_configuration.AuthenticationTokenScope}),
                        cancellationToken)).Token;
```
as a delegate/func/action making it easy to change?
 what do you think?
@iancooper @preardon ?

iancooper:
I wonder if the duplication around connection means where we are at the point where it would be worth factoring out a ConnectionFactory, passing it the various parameters that can lead to different auth approaches, and exposing a Create and CreateAsymc on it, and trying to factor out the common code between those methods

preardon:
@holytshirt I was thinking about something Similar as at the minute I'm pretty much giving users the ability to Specify SharedTokenCredentialCache, or to Ensure that their environment points them to the correct Path with DefaultAzureCredential I've personally had some small issues with always using DefaultAzureCredential on my dev machine and as such I usually Specify to be safe, also this is Of course a very Azure implementation I think Passing it as an Action could be good.

@iancooper A connection Factory also sounds good.

If there any Appetite to Move this to EF Core so that we can use  things like DbMigrations to help create the Schema, I feel it might be a bit hefty for what we are trying to do here however.

preardon:
@iancooper Did you want me to make a Connection Factory for these?

iancooper:
> @iancooper Did you want me to make a Connection Factory for these?

Yes please

preardon:
@iancooper How is this mate?

I have added Connection Factories for

- SQL Authentication
- Default Azure Credential
- Managed Identity (Azure Web App or Azure VM)
- SharedTokenCache (useful for Local Development)

iancooper (approved):
Looks good to me. At some point we may want to review if the Outbox/Inbox split which exists to avoid you taking dependencies on types you don't need) is worth the duplicated code. But don't fix that now, I think we should raise a discussion on that.

 But under those parameters, this is great, and good to go IMO

preardon:
@iancooper Excellent Thanks Mate

@holytshirt Can we please get this into the v9 release?

iancooper:
> @iancooper @holytshirt Looks like the build timed out, can we please re-run and potentially increase timeout?

I've kicked if off again, though GA is temperamental and tends to stay broken on a build once if fails.

iancooper:
I'll merge, those tests are failing for CI issues over issues related to this PR

PR #1506, opened 2021-04-27
[Feature] Support different approaches to SNS Topic Validation

iancooper:
ListTopics is rate limited at 30 calls per second.  In a large deployment that can mean the strategy of clients validating an Arn or TopicName will be rate limited. it is better to use GetTopicAttributes via Arn where possible. This also allows validation of an Arn outside of the caller's account i.e. the publisher and subscriber are in different accounts. ListTopics only list topics for the account, but GetTopicAttributes works anywhere. 

GetTopicAttributes requires an SNS Arn. You either set the routing key to an arn, or we can infer the sns arn from the name by convention. Names inferred by convention will be assumed to be in the executing code's account - so the only way to get a cross-account Arn is to pass the Arn in directly, we can't infer from the executing code.

Find by name still works for the current account, it will make a List Topic call, which is rate limited to 30 per second

### DIFF
--- a/src/Paramore.Brighter.MessagingGateway.AWSSQS/AWSMessagingGateway.cs
+++ b/src/Paramore.Brighter.MessagingGateway.AWSSQS/AWSMessagingGateway.cs
@@ -19,11 +19,11 @@ namespace Paramore.Brighter.MessagingGateway.AWSSQS
             _awsConnection = awsConnection;
         }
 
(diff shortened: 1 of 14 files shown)


iancooper:
@holytshirt Going to commit this, as I'd rather you reviewed the RMQ pub confirm if you have time over SNS/SQS details

Issue #1509, opened 2021-05-01
When I Deserialise A Message in V9, it does not Deserialise


preardon:
v9 Bug (since System.Text.Json)

it appears when Desearalising Message and Message Body System.Text uses the incorrect Constructors

preardon:
Closing now this PR is in master

PR #1510, opened 2021-05-01
Fix DeSeralisation Issues around Message and MessageBody #1509

preardon:
@iancooper Fixed this issue and Wrote a Test, Let me know if you want me to do any more


preardon:
@iancooper Core Tests are all Green Locally if you're happy with this change and the tests time out

iancooper:
Can't really see what changed, GitHub not being helpful on deltas, but it seems fine. 

Can you do me a favor whilst you are there? We don't need the Postback property of a message any more, as we no longer use the message pump for async pipelines. We also don't need the MT_CALLBACK (I think?) message type which tells you to use the Postback property on a message. 

Could you delete those?

Let me know if that pops up an issue with something that seems to stop working. I don't think it should..

preardon:
@iancooper effectively I just Specified the exact constructor for the serialiser to use, it was tripping up over itself. I will delete as requested. I'll run the tests and get the PR updated in the morning.

preardon:
@iancooper I removed MT_CALLBACK and PostBackItem, had to Delete the Execute Method and the one places that called it (Which only called it for Async) All Core Tests are Green Locally

iancooper:
Cheers @preardon

Issue #1516, opened 2021-05-04
AWSSQS - SqsSubscription Constructor parameters incorrectly assigned


honkuan86:
### Describe the bug
Found that the SqsSubscription constructor params (bufferSize, noOfPerformers) is assigned incorrectly to the base Subscription class, causing the Dispatcher always create the performers according to the value of bufferSize.

### To Reproduce
Run the AWS sample project should be able to reproduce it.

### Exceptions (if any)
No exception, as it is a behavior bug.

honkuan86:
@iancooper , I'll pick this up

honkuan86:
@iancooper , I can pick this up.

iancooper:
@honkuan86 Please do. Thanks for spotting this one.

PR #1521, opened 2021-05-05
Fix for SqsSubscription Constructor parameters incorrectly assigned issue

honkuan86:
Related Issue: #1516


iancooper (commented):
Good spot, the reverse in parameter order from Subscription is a pain though. Maybe we should re-order these parameters on the SqsSubscription constructor to match the order on Subscription as well. That should be more comprehensible, as I suspect this problem occurred because  of the ordering not being the same here

honkuan86:
> Good spot, the reverse in parameter order from Subscription is a pain though. Maybe we should re-order these parameters on the SqsSubscription constructor to match the order on Subscription as well. That should be more comprehensible, as I suspect this problem occurred because of the ordering not being the same here

Yup agreed on re-ordering the parameters. Should we consider those who already using this class ? If we re-order the parameters, the changes will be hard to notice, since both of these are int type, so they won't be notified from the compiler until they run the program.

iancooper:
> Yup agreed on re-ordering the parameters. Should we consider those who already using this class ? If we re-order the parameters, the changes will be hard to notice, since both of these are int type, so they won't be notified from the compiler until they run the program.

The switch to the SqsSubscription class comes with v9, so only early adopters will be using, who should have taken it as "pre-release'. So I would rather change now, before we go to final, than leave it to later when we will have the problem you identify

iancooper (approved):

iancooper:
Thanks for that. I'll merge

PR #1522, opened 2021-05-07
Breaking out the tests again

iancooper:
We have not had clean build since build and core tests were merged, going to break core tests back out in order to help us get green as currently the build is an obstacle to release


iancooper:
Fixed the issue with long test times by removing the async Task change to dispatcher tests

Issue #1409, opened 2021-02-21
Bug: Missing Tests for MSSQL Gateway


iancooper:
### Describe the bug
The MSSQL has no regression tests

### To Reproduce
Look at the project structure, we might expect them to be in the MSSQL tests project under a MessagingGateway folder, but they are not there

### Further technical details
Whilst a transport may use Test After as opposed to Test First (because you are writing a plugin not designing contract first) we do need to ensure that we don't regress a transport when we make changes to the core, which does require automated tests we can run as part of our CI process

iancooper:
@preardon Have we moved the dial on this one at all?

preardon:
@iancooper I wasn't looking at this one, I can pick it up if you would like and can give it the Azure.Identity Treatment while I'm there if you like

iancooper:
@preardon Be cool if you could, I'd like to kill any bugs we have related to areas we are working on. (I'd like to kill anything marked Bug, but this is a start).

preardon:
@iancooper is Anyone going to get upset, If I modify the Test Helped to point to a docker instance instead of The local one currently being used for Inbox and OutBox

preardon:
@iancooper Found a few bugs will fix while I'm in the Ditch

iancooper:
@preardon No, that's a good idea.

iancooper:
> @iancooper Found a few bugs will fix while I'm in the Ditch

Yeah, I thought that might happen. Hoping to have the logging review done today

preardon:
@iancooper Have we had issues with v9 and Deserialization of Messages in General?

preardon:
Logged a Bug and PR for the Above #1509 and will continue with this after the fix is in

preardon:
@iancooper I'll have a look, can we spin up Docker in your Pipeline to connect these to?

preardon:
@iancooper I've found it and Pushed a change

iancooper:
Closed this, thanks @preardon

PR #1526, opened 2021-05-08
MSSQL messaging gateway Tests #1409

preardon:
This gives us Decent Coverage
image

I haven't Covered some of the parts (Async, Reject, ect.) as They are not implemented. I was thinking of throwing in some Sad path Infrastructure tests as well @iancooper let me know if there is anything else you want me to cover in particular

Also, tests have been converted to use a MsSQL on Docker


iancooper (approved):
This is a great start, we can add more as situations crop up that require them

Issue #1198, opened 2020-07-17
Remove RavenDB from documentation


iancooper:
**Describe the bug**
We no longer support RavenDb as a store

**To Reproduce**
Some of our documentation still refers to RavenDb: https://www.goparamore.io/new-page-2

**Expected behavior**
There should be no references to RavenDb in our documentation

**Additional context**
We should review the list of storage providers for accuracy in our documentation

preardon:
@iancooper I believe I've done this as part of the PR let me know if there is anywhere else

omelianlevkovych:
Hi @iancooper , is this issue still relevant? 
If so I will be happy to work on this as my first contribution.

iancooper:
@omelianlevkovych It still is I believe. The Docs repo under the org has recently been moved over to markdown. I am working on making it available via GitBook soon. But if you look in the markdown there, I believe there are some hanging references to Raven.

iancooper:
@omelianlevkovych Ah, I have noted above that it might have been done by Paul.

If so, let us know what you can help with, we are always happy to welcome new folks.

iancooper:
So the discussions forum is good. We also have some issues that are flagged with help_wanted. I'll set up a discussion, invite you and, and we can figure out the easiest way to help you onboard

iancooper:
@omelianlevkovych 

I posted a discussion here: https://github.com/BrighterCommand/Brighter/discussions/2045

Figuring out if we can close this issue would help too :-D

PR #1527, opened 2021-05-09
Removed All references to Raven DB from Help #1198

preardon:
Not sure how to test the changes below, but here are all references removed


iancooper:
I suspect this is the wrong place to look i.e. these are not the source files. Let me give you more info next week.

PR #1531, opened 2021-05-11
Fix the final 'red' tests

iancooper:
Experimenting to fix the remaining red tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,7 +163,7 @@ jobs:
 
   kafka-ci:
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 20
     needs: [build]
     env:
       CONFLUENT_BOOSTRAP_SERVER: ${{ secrets.CONFLUENT_BOOSTRAP_SERVER }}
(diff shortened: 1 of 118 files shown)


iancooper:
Now tracking #1538 

@preardon I have this, as fixed for EventStore and needs rolling out but though you might want to know about this behavior. This blog post is useful: https://josef.codes/custom-dictionary-string-object-jsonconverter-for-system-text-json/

It also tells us that only EventStore is testing that we read the message header back successfully from the Outbox, so some new tests needed around this one

preardon:
@iancooper Sorry about this Mate, let me know what I can do to assist, I'm keen to do whatever I can to get Brighter v9 out ASAP

iancooper:
> @iancooper Sorry about this Mate, let me know what I can do to assist, I'm keen to do whatever I can to get Brighter v9 out ASAP

Oh gosh, no worries. Should have it done tomorrow, and you can give it the once over for me

iancooper:
@preardon In the end I just opted to use Josef Ottoson's approach to writing a custom converter that we could decorate the bag with, so that System.Text.Json works as expected. it means that we don't need to write this code everywhere.

Not done, I need to add tests in the Outboxes to ensure that we read and write some headers, so that we will pick this bug up again in future, if something changes. That was the root of us missing this, broken EventStore tests meant we didn't see this issue there, but perhaps worse it was the only Outbox testing for correct serialization of a bag.

iancooper:
It looks like another test is being fragile on CI. Runs fine locally, but something we changed is now spinning on the containers. I may have to run locally with Act to diagnose

Issue #1540, opened 2021-05-17
The Azure Service Bus Transport does not support Rejecting


preardon:
At the Current point in time the Azure Service Bus Transport does not support Rejecting the Message

Line 211 of AzureServiceBusConsumer
`s_logger.LogWarning("Reject method NOT IMPLEMENTED.");`

and as such messages will not find their way to the Dead Letter Queue

preardon:
I've just submitted PR #1541 For this

preardon:
@iancooper I am under the Assumption (Based on SQS Gateway) that we want the Reject Command to Put the Letter on the Dead Letter Queue instead of a Just Abandoning the Lock (Which would either Retry or Dead Letter dependant on retires)?

iancooper:
@preardon Yes, we have an explicit requeue op for 'make this available to others' (although its common to delay to allow transient issues to correct). Reject means don't re-queue this and put on the DLQ if we have one.

preardon:
@iancooper Excellent, the PR #1541 Should Give us this for ASB

iancooper:
@preardon Cool, will look at lunch or after close of play today

PR #1541, opened 2021-05-17
Add Dead Letter to ASB Transport #1540

preardon:
Plumbing through ASB Dead Letter (On Rejection)

@iancooper This shouldn't have a effect on The Other users as it would will only run when Non Destructive Read is enabled (ackOnRead = true;)

### DIFF
--- a/src/Paramore.Brighter.MessagingGateway.AzureServiceBus/AzureServiceBusConsumer.cs
+++ b/src/Paramore.Brighter.MessagingGateway.AzureServiceBus/AzureServiceBusConsumer.cs
@@ -208,7 +208,26 @@ namespace Paramore.Brighter.MessagingGateway.AzureServiceBus
 
         public void Reject(Message message)
         {
-            s_logger.LogWarning("Reject method NOT IMPLEMENTED.");
+            //Only Reject if ReceiveMode is Peek
+            if (_receiveMode.Equals(ReceiveMode.PeekLock))
+            {
+                try
+                {
+                    EnsureSubscription();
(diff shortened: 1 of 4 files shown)


iancooper (commented):
Hey @preardon question on ASB. Do you get the DLQ for free when you create a queue. Not sure where you are setting it up here. Probably, or I missed something (other transports require explicit config)

preardon:
@iancooper Yup You get Free DLQ and Retires on Messages in ASB
image
Here is a Screenshot showing, you get an Active and DL For Each Subscription of a Topic

preardon:
Also, @iancooper as you may have gleamed from this I have a plagiarised version of your Greeting Sample app for ASB if you would like me to throw it in the Samples Folder

iancooper:
> Also, @iancooper as you may have gleamed from this I have a plagiarised version of your Greeting Sample app for ASB if you would like me to throw it in the Samples Folder
Please do

iancooper:
Happy to merge when you are

preardon:
@iancooper I've made 1 other change in another Branch but I'm still mulling over if we should do it or not. It allows us to swap from using Brighters Requeue Retires or Using ASB's Retries and DLQ from this.

Lets just get this one in and talk about the other Change later if we want it :)

iancooper:
@preardon There is stuff worth pursuing here, but for 9.X. We can have some proper design discussion so you know what we have, how we think it could be improved and we can figure out what success looks like

iancooper:
I suspect just adding the DLQ works fine. I'll see if I can set up some time to pair with you

Issue #1542, opened 2021-05-18
ASB Transport - Async Producer


preardon:
**Is your feature request related to a problem? Please describe.**
I would like The ASB Transport to be Async

PR #1543, opened 2021-05-18
Added Async to ASB Message Producer #1542

preardon:
Issue #1542


iancooper:
@preardon I'm wondering a little, if now that we have a callback mechanism to commandprocessor for success/fail on a publish to middleware, if we can leverage that with async send to allow you to use a callback.

If we get that call setup, we can talk about it

Issue #1519, opened 2021-05-05
Messages that are retried too many times should move to DLQ if there is one


iancooper:
### Describe the bug
To avoid a poison message scenario we only allow you to requeue a message x times. In the event that you are tried beyond that count we ack you, so you won't go onto any DLQ.

In addition, we treat all user code exceptions that bubble out of the handler as application errors i.e. we just ack and move on. We might chose to let you opt to retry a failure that you think is timing related using our requeue option (attribute?)

### To Reproduce
Create a handler that throws a user code exception, it will not reach the DLQ

preardon:
Hey mate, I am happy to pick this one up if you want.

From what I can see, in the Message Pump, on When Catching Generic Exceptions we Log an Error, Exit the Catch Block and then the Code Continues on to Acknowledge.

I'm Guessing in the Catch Block we want to call Reject Message and then Continue on to the Next Iteration of the Loop? Perhaps Increment The Unacceptable Message Limit as well

iancooper:
@preardon Feel free to give it a go (separate PR please).

So, if the channel supports a DLQ, we need to count the number of times you requeue a message. To do that, we need to: 

* Maintain a count in the header of the number of times it has been requeued. (We might do that already, nagging suspicion). * Increment the counter on requeue
* When the count exceeds a user-defined value (set in the base Subscription class) - send to the DLQ via Reject
* If we send to the DLQ, increment the unacceptable message limit.

At some point we need to look at the unacceptable message limit, because it really needs to be a window. But DON'T tackle that here, just get the DLQ on enough requeues working

iancooper:
PS Channel is your friend here I suspect. It wraps the per-platform message consumers. So if you want to put some generic code i.e. incrementing the count in the header, this is where you would do it.

iancooper:
@preardon Yes, we have a HandledCount on the message and the pump currently discards when hit, but should reject

The wrinkle here is that HandledCount is used for the explicit DeferMessage exception that can be raised by a consumer. That is how you should reject to retry the message (don't forget you just have a handler, not the consumer, so you can't reject, instead you throw a specific exception

iancooper:
@preardon I can do a Zoom call later if you want more help understanding this. But have a stab, and let's see what it looks like.

As ever, its Core, so TDD please :-D

iancooper:
[...]

PR #1544, opened 2021-05-18
Send Commands to the DLQ if their limit is reached #1519

preardon:
The Code so Far works (Based on my Unit Tests, will run sample app against them soon) for Commands (not events at this stage, I've got a few Questions there 😊

Anyway, I have set the Pump to Requeue on Generic Error as this will mean that if you set the Retry count to 0 it will Automatically Reject the Message.

The next Change I have made is to have RequeueMessage return a Bool to denote Weather the Message was Requeue or Rejected, This is because if we have Rejected we shouldn't still need to Acknowledge The Message, also we had it set to ACK the Message if Disposal was Turned on and the Count was exceeded, I feel this should be where we do our Reject of the Message

In Terms of Action of Transports in our System they Have the following Actions

ASB
	-> Reject : Currently Nothing, Proposed DLQ if Non Destructive read enabled (Peek Lock / AckOnRead)
	-> Requeue : Producer.Send
	
Kafka
	-> Reject: ACK
	-> Requeue: Nothing
	
MSSQL
	-> Reject: Not Implemented
	-> Requeue : Send

Redis
	-> Reject: Remove
	-> Requeue: Add new, Remove Existing

RestMS
	-> Reject: Nothing
	-> Requeue: Nothing

RMQ
	-> Reject: Reject
	-> Requeue: Basic Publish

SQS
	-> Reject: DLQ\Delete
	-> Requeue: Change Visibility

### DIFF
--- a/src/Paramore.Brighter.MessagingGateway.AzureServiceBus/AzureServiceBusConsumer.cs
+++ b/src/Paramore.Brighter.MessagingGateway.AzureServiceBus/AzureServiceBusConsumer.cs
@@ -109,7 +109,8 @@ namespace Paramore.Brighter.MessagingGateway.AzureServiceBus
             s_logger.LogDebug("Received message from topic {Topic} via subscription {ChannelName} with body {Request}.",
                 _topicName, _subscriptionName, messageBody);
             MessageType messageType = GetMessageType(azureServiceBusMessage);
-            var headers = new MessageHeader(Guid.NewGuid(), _topicName, messageType);
+            var handledCount = GetHandledCount(azureServiceBusMessage);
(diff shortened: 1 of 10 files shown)


preardon:
@iancooper I've got a few Questions here around how you want the pump to Function, also raised in the related #1519 Something you mentioned about Requeuing only if a DelayActionException is raised

preardon:
@iancooper I have just tested this with my Sample app on ASB and it works as we discussed. Was also a little Cheeky and added a small ASB teak to do with this into this Branch, I put it in another Commit, Please let me know if you want me to remove it and submit it as a Separate PR

iancooper (approved):
It looks right, let's get it out and get feedback.

Issue #1545, opened 2021-05-20
When Adding Service Activator, if you Specify BrighterMessaging it does not register an Async Handler Factory


preardon:
When Configuring as such
services.AddServiceActivator(options => { options.Subscriptions = subscriptions; options.ChannelFactory = new AzureServiceBusChannelFactory(asbConsumerFactory); var outBox = new InMemoryOutbox(); options.BrighterMessaging = new BrighterMessaging(outBox, AzureServiceBusMessageProducerFactory.Get(asbConfig)); }).AutoFromAssemblies();

Specifically when adding the BrighterMessaging (So that my Handler can Post commands as well)

We end up hitting the following (ServiceCollectionExtensions.cs Line 55)
`
if (options.ChannelFactory is null)
{
//TODO: Need to add async outbox

            taskQueuesBuilder = options.BrighterMessaging == null
                ? messagingBuilder.NoTaskQueues()
                : messagingBuilder.TaskQueues(new MessagingConfiguration(options.BrighterMessaging.OutBox,
                    options.BrighterMessaging.AsyncOutBox, options.BrighterMessaging.Producer,
                    options.BrighterMessaging.AsyncProducer, messageMapperRegistry));
        }
        else
        {
            taskQueuesBuilder = options.BrighterMessaging == null
                ? messagingBuilder.NoTaskQueues()
                : messagingBuilder.RequestReplyQueues(new MessagingConfiguration(options.BrighterMessaging.OutBox,
                    options.BrighterMessaging.Producer, messageMapperRegistry, responseChannelFactory: options.ChannelFactory));
        }
`

The issues with this is if we Provide a ChannelFactory (Which we must for consuming) And We Specify BrighterMessage (Which we need for Producing), We end up with RequestReplyQueues

and then when we build (CommandProcessorBuilder.cs line 224)
`
if (!(_useTaskQueues || _useRequestReplyQueues))
{
return new CommandProcessor(

                subscriberRegistry: _registry,
                handlerFactory: _handlerFactory,
                asyncHandlerFactory: _asyncHandlerFactory,
                requestContextFactory: _requestContextFactory,
                policyRegistry: _policyRegistry,
                featureSwitchRegistry: _featureSwitchRegistry);
        }
        else if (_useTaskQueues)
        {
            return new CommandProcessor(
                subscriberRegistry: _registry,
                handlerFactory: _handlerFactory,
                asyncHandlerFactory: _asyncHandlerFactory,
[...]

PR #1546, opened 2021-05-20
Issue/1545 no async handler factory #1545

preardon:
@holytshirt A relatively small fix Let me know what you think?

Also Added my Sample Project as a 2nd commit so if you just look at commit 1 it'll show you the scope of the Specific Changes


iancooper (approved):
Looks good, let's merge it ahead of my merging the header fixes tomorrow morning

PR #1572, opened 2021-06-02
Update to TaskCompletionSource

holytshirt:
- update  TaskCompletionSource to run RunContinuationsAsynchronously


iancooper:
I'm  still a little nervous here. Looking at this: https://devblogs.microsoft.com/premier-developer/the-danger-of-taskcompletionsourcet-class/ it seems to force asynchronous behavior, which may be what we want if we say we are 'async' but we have a custom synchronization context which means that we use the same thread, and not a thread pool thread, to service continue with statements. What I am nervous of is how this interacts: do we start running our continuations on thread pool threads?

I guess we have two options: (a) merge and see if anything breaks (b) prove that this doesn't introduce a new thread in the continuation i.e. set up an async pipeline for the dispatcher?

Maybe, let's merge and then we can play with this? It's not changing the API, so we can look at other options if we need to

iancooper:
@holytshirt Assuming you were going to merge?

holytshirt:
@iancooper done

iancooper:
@holytshirt One to play with if either we or @preardon get the time I think, check that behaves as expected still, with some thread ids to check

Issue #1574, opened 2021-06-03
MySqlOutbox retrieve outstanding messages incorrectly


honkuan86:
### Describe the bug
The current MysqlOutbox implementation of getting outstanding messages is currently unable to get the correct messages due to the filtering below,
`TIMESTAMP < DATE_ADD(CURRENT_DATE(), INTERVAL @OutStandingSince SECOND)`
As CURRENT_DATE() is only returning the date and not time, hence the filter result will not be accurate when we set the interval in seconds.

So CURRENT_TIMESTAMP() should be used instead, as it return the date and time.

### To Reproduce
Run the outbox sweeper with MySqlOutbox in the sample project should be able to reproduce the issue.

iancooper:
We'll need to check on the tests for this, as to why we didn't pick up, and look into the other SQL providers for repetition of the error

preardon:
I believe We've found an Issues with This Query Still Does not appear to work for me there where statement is problematic as it will work if nothing is dispatched, but otherwise the row numbers are out. PR incomming

iancooper:
@preardon Is this still outstanding?

preardon:
@iancooper from what I remember this is fixed and merged in, but I don't think its been published on nuGet as yet

iancooper:
@preardon I'll close the issue for sanity

PR #1575, opened 2021-06-03
MySqlOutbox outstanding messages query fix

honkuan86:
Fix the OutstandingMessages query in MySqlOutbox.

Related to issue: #1574


iancooper (approved):

iancooper:
Thanks for this @honkuan86 I may need to check if the other SQL providers have the same error

Issue #1583, opened 2021-06-13
MsSQL Outbox Error when calling OutstandingMessages - System.InvalidCastException: The SqlParameterCollection only accepts non-null SqlParameter type objects, not SqlParameter objects


preardon:
I am vaguely sure this was working, and I can not yet see why it is doing this, however at the minute I am currently getting

System.InvalidCastException: The SqlParameterCollection only accepts non-null SqlParameter type objects, not SqlParameter objects.
   at Microsoft.Data.SqlClient.SqlParameterCollection.ValidateType(Object value)
   at Microsoft.Data.SqlClient.SqlParameterCollection.AddRange(Array values)
   at Paramore.Brighter.Outbox.MsSql.MsSqlOutbox.CreatePagedOutstandingCommand(DbCommand command, Double milliSecondsSinceAdded, Int32 pageSize, Int32 pageNumber)
   at Paramore.Brighter.Outbox.MsSql.MsSqlOutbox.OutstandingMessages(Double millSecondsSinceSent, Int32 pageSize, Int32 pageNumber, Dictionary`2 args)
   at Paramore.Brighter.OutboxSweeper.Sweep()
   
I find it confusing that this would be an issue in the Brighter as I would imagine that there are tests and someone would have brought it up. 

If anyone has any ideas as to how my configuration could have caused this then that would be much appreciated, otherwise I am going to keep looking and post an answer here when I have one

preardon:
I've found something that fixes this, but Its on the Brighter side,

If we change the Reference in MsSqlOutbox to Microsoft.Data.SqlClient from System.Data.SqlClient, Then it Magically starts working, I'm not Overly sure Why this works, or how it would work if we were to pass a Null Value into the Parameters.

@iancooper Given that Microsoft have been looking at replacing System.Data.SqlClient with Microsoft.Data.SqlClient, what are your thoughts on me updating these References?

holytshirt:
I would use ` Microsoft.Data.SqlClient`. ` System.Data.SqlClient` is only for legacy support.

preardon:
@holytshirt that was my understanding as well. 

I have played with this a bit this afternoon and it looks to work fine so far, didn't quite get to the ci tests in the pipeline due to the artifacts post failure.

I've been spending some time in one around outbox today, will have some questions/suggestions but can obviously wait until v9.x

iancooper:
[...]

PR #1593, opened 2021-06-14
Updated References from System.Data to Microsoft.Data

preardon:
#1583 

This should sort the remaining References to all three projects for MsSQL


preardon:
@iancooper @holytshirt 
This the problem started again, so I have gone through and removed all References to System.Data

iancooper:
Thanks @preardon let' see how that plays out. The Github package feeds are annoying, @holytshirt If they are not in a separate step we can't see if the core build and tests run anyway. What do you think to moving them back out?

preardon:
@iancooper  I ran all SQL Tests locally before I Pushed

PR #1594, opened 2021-06-14
Updated Outstanding Command SQl Query

preardon:
#1574 

I have moved the Dispatched Is Null into the First part of the query so that the Numbers are only for Rows that have not been dispatched

I have Changed Sort to ASC so that we get dispatch from the oldest Frist

I have set the Date Add to Negative so the logic reads where TimeStamp is less than @outstanding ms ago not less than @outstanding ms in the future


iancooper:
@preardon Any chance that you could check the other SQL providers and see if we need to change their logic too?

preardon:
@iancooper Yup, not a worry mate

iancooper:
@preardon Good man. This is new this release, so bound to be a few kinks

preardon:
@iancooper  I'm not the most Familiar with MySQL however I don't think this will work as its not creating a new ID for the Paged Query

SELECT * FROM {0} AS TBL WHERE `CreatedID` BETWEEN ((@PageNumber-1)*@PageSize+1) AND (@PageNumber*@PageSize)

preardon:
@iancooper Haven't run tests, however here is SQLite, MySQL, and PostGres

iancooper:
> @iancooper I'm not the most Familiar with MySQL however I don't think this will work as its not creating a new ID for the Paged Query

I'm not either, so if you think it needs fixing try it, you are as likely to be right as the original code. We just need some tests to confirm probably

iancooper:
Not sure I have much chance to help run/write tests etc before Friday. But let me know how you are getting on. Hopefully I will make the Wednesday meetup, but will be in the office for meetings so possibly I might have to skip it

preardon:
I'll have a go at throwing some tests in when I get a minute

iancooper:
> I'll have a go at throwing some tests in when I get a minute

Thanks, the SQL is a bit of a slog. I assumed we had some tests that ought to be giving us confidence in the results from the SQL, but presumably not. Feel free to focus on one and then do others, so that if if gets overwhelming we can divide it up. I seem to recall it is a bit of a slog.

Issue #1560, opened 2021-05-25
Declaration of a Service Lifetime Scope in ServiceActivator


preardon:
At the Minute the Service Activator is an IHostedService CommandHandlers processing these messages cannot consume a Scoped Service without Injecting a ScopeFactory and creating a new scope inside my Handler

I would like Brighter to Declare a scope, the Real question is where
- If we declare a Scope in the pump around the dispatcher this would solve the issue however the scope would be shared Further down stream (i.e. for Events we handle in chain and this will share the scope between them)
- If we declare a Scope inside of the Command processor this would also solve the issue however it may be  a bit low down in the code

Any Thoughts guys?
@iancooper @holytshirt

iancooper:
We don't want to bake the IoC library into Brighter's Command Processor,. Dispatcher etc. That is a design decision and comes from this article by Mark Seeman: https://blog.ploeh.dk/2014/05/19/di-friendly-framework/ There was a *strong* design discussion with the ASP.NET team around some of their design decisions here, but those conversations mostly fell on deaf ears.

So we have effectively Create and Release, and we expect implementations of HandlerFactory to work with that i.e. you pass us an instance when we say Create and you can do as you will with it when we say Release.

Whilst we support plugging into HostBuilder, we don't want to bake in the ASP.NET scopes to Brighter's core. (It looks as though the new lightweight HTTP offering won't require their own IoC container).

I can take a look and see what we might be able to do, but one thing we don't want to do is put the ASP.NET DI mechanics for scoped services inside Brighter's engine. What would be better is to figure out a way to manually control lifetime via Create and Release that worked with ASP.NET's viral approach to this

penderi:
Mate this is awesome, thanks @iancooper.
I agree about Di being on the age - and rightly too  - in the framework. Adding direct DI scopes wouldn't follow the current design at all.
How ameniable are you to an optional/confgurable IScopeFactory being added to the MessagePump? (I *think* that's the right level... ). We can then provide NoOp version and a version in the DI package that plugs into MS DI - when used it will wrap the call to DispatchRequest in a new DI Scope. 

They key behaviour we need is that each consumed message is within a scope so that dwonscale scoped dependencies (hello EF Core) are shared amongst repositories. Without it we're having to referenec and use the scopefactory in our handlers.

Here's hoping I'm not missing

iancooper:
[...]

PR #1598, opened 2021-06-18
#1560 - serviceactivator scope declaration - initial scratch for thou…

penderi:
### DRAFT

Issue - #1560 

Initial scratch around the discussion that allows serviceactivator to pass a scope handler factory through to the pipeline. This seems to work in a naïve situation (a scope per handler call) which will work in a simple worker scenario for the dependencies.

However, attributes and other handler chaining, will receive a new scope. This is not what we want. We'll need to track the pipeline id, perhaps, and release on final dispose there. 

MORE WORK REQUIRED


iancooper:
Do you want me to convert to a draft, so you can keep updating, or take  a look (probably Sunday) with a view to merging for now and iterating?

penderi:
Oh no - not worth a review yet. The remaining issue on my local is the scope awareness. I'm trying to minimise the changes, but I can't see a non-breaking or even minor change to ensure each pipeline creates a new scope, and that new scope is used in a handlerfactory. I can successfully track scope but I need to initiate at the start of the pipeline and use it in the build of the pipeline to create the handlers.

I'm sure we'll have something early next week for review!


________________________________

iancooper:
> Oh no - not worth a review yet. 
OK, I'll mark as draft

penderi:
@iancooper still draft BUT this is the simplest change I can do. Everything is there... bar the useful bit lol.

ISSUE - I can't see how IAmAHandlerFactory.Create(Type handlerType) has enough information to know about the scope we're using.

This is the crux of the singleton factory / transient-pipeline issue. I've the LifetimeScope controlling and identifying the scope. That's cool. Ideally we'd pass the lifetimeScope down to the factory on the create. That would sort it, but it's a bigger change than I was hoping for.

Other than that... I've use the HandlerFactory interface to optionally have scope methods. This may be leaky, but I can't see a great way to do it without much greater change.

iancooper:
@penderi I'll try to have a think about that this week

iancooper:
@preardon @penderi Is this still live, or has it been overtaken by events on our work on EF Core support?

preardon:
@iancooper We are still working on this but I believe @penderi has a slightly different solution at the minute

iancooper:
@preardon Cool, I'll keep it open

iancooper:
@penderi @preardon working my way through, to figure it all out

Issue #1599, opened 2021-06-19
Update Outbox Implementation to Allow do be Injected via DI


preardon:
In some Circumstances I would like my MsSql Outbox to share the DB Connection to my code so that I can wrap db updates and Outbox Transactions in a Single DbDatabase transaction

It appears that I could make this changes Specifically in the Command Processor Builder, Also this appears to only affect Task Queues

I also want to have a little bit of a look at how the MsSql Handles its connections as at the Minute running a Post Opens and closes connections to the DB at least 3 times

PR #1601, opened 2021-06-19
Update Brighter Builder to allow DI of outbox as well as UseExternalMessaging

preardon:
@iancooper @holytshirt 

The below is very rough and for the minute please ignore the majority of the changes in MsSqlOutbox.cs I'm looking at what has to happen to get Transactionality into there shared with Applications.

The main change I would like to discuss here is the DI Change

`
services
                .AddBrighter(opt =>
                {
                    opt.BrighterMessaging = new BrighterMessaging(producer);
                    opt.PolicyRegistry = new DefaultPolicy();
                    opt.CommandProcessorLifetime = ServiceLifetime.Scoped;
                })
                .UseMsSqlOutbox(outboxConfig, typeof(MsSqlOutboxSqlAuthConnectionFactory), ServiceLifetime.Scoped)
                //.UseMsSqlOutbox(outboxConfig, typeof(MsSqlEfConnectionFactory), ServiceLifetime.Scoped)
                .MapperRegistry(r =>
                {
                    r.Add(typeof(GreetingEvent), typeof(GreetingEventMessageMapper));
                    r.Add(typeof(GreetingAsyncEvent), typeof(GreetingEventAsyncMessageMapper));
                });
`

This requires moving the Outboxes outside of MsSqlOutboxConfiguration and into DI Along with the Connection Factory

Anyway I'm going to Keep looking into how to get the transaction from the Application down onto this level, I don't see how this could currently work with the existing implementation as all of the Opening and closing of connections would end them


preardon:
This is now a Working Prototype for Both Scoped, Singleton and Shared DB Transactions (MsSql) Scenarios

Changes and things to note

Previously the Outboxes were Specified in the Brighter Messaging Configuration however now if you want an outbox you can Add one using an Extensions to the IBrighterHandlerBuilder i.e. Services.AddBrighter().UseMsSqlOutbox();

I have had to Make some changes to how MsSql outbox works, Previously we would Dispose of the connection after each command, this does not work when we have a Connection and / or Transactions Shared between calls.

The Logic now goes

1. Get Connection
2. if Connection not open, Open Connection (If there is an ongoing connection the connection will be open, otherwise it will be closed)
3. Execute SQL (If this isn't a Non Query Close the Data Reader)
4. If this isn't a Shared Connection, Dispose the Connection, else if this connection isn't part of an open Transaction then Close Connection

preardon:
There still appears to be one issue when using a Shared Connection, when we run the Outstanding Message check as a background task, sometimes there can be concurrently open Data Readers. Currently tossing up getting it its own Connection  or putting further checks in.

@iancooper on a related not, do you think we need this check on Clear as we should really be running the Sweeper which would negate the need for this

iancooper:
@preardon Let me investigate that

iancooper:
@penderi @preardon I know that your use case is focused on MSSQL, but I wonder if we ought to consider all potential EF targets here (Postgres, MySQL, and SQlite I presume). I suspect in all of those cases we would want to share the connection with EF so as to ensure the transactional scope included them. It feels like this is more a UseEntityFramework than a UseMSSQL addition to the fluent configuration because that is the key constraint. That might change how we think about this, because it could introduce some abstraction.

preardon:
@iancooper I was thinking that this could also be used for MsSql by having a DI Sql Connection to share between the App and the outbox. If the Pattern is include the Outbox in the Transaction so that you only send the event if the entities are saved to the DB and vice versa you need a DB Transactions over the Outbox as well.

I am happy to also Create a MsSql outbox if you thing this is better as well however making a generic one of these may be fun 

I do agree that if we do this for MsSql we should look to do it for all other Outboxes and inboxes as well

iancooper:
> If the Pattern is include the Outbox in the Transaction so that you only send the event if the entities are saved to the DB and vice versa you need a DB Transactions over the Outbox as well.

So the idea is that the Deposit call gets included in the transaction scope - write to the Db, write to the Outbox - so that we get consistency. You can then call Clear outside of the transaction to push the work now (you can also let the Sweeper pick it up, I prefer explicit as well, YMMV).

So for any Db with transaction support, you want to be able to use the same transaction across the save and deposit.

I guess for a generic version I was assuming that we would need to be able to grab the connection from a "provider" that would abstract out differences between the Db type. So as we are explicitly worried about EF, it could be an IAmAnEFConnectionProvider and you would inject that and grab it from there. But I'm not close enough to this, so could be nonsense.

I've not really thought about how Dapper works with this, but I guess if we could support EF, Dapper and ADO.NET we have the major bases covered

iancooper:
PS I am not sure that the Clear or Sweep needs to use the same connection though, it's not in the same transaction. We don't guarantee you won't get duplicates, so I am fine with the risk that you update the Outbox of an older send at the same time we sweep, as it should just result in another send

preardon:
@iancooper  I completely agree and meant to mention it. if you want to rely on Transactionality then you should only Deposit in the Trans not Post (as this will Post & send)

The other thing I was thinking which I believe you eluded to was that the name "ConnectionFactory" no long appears be be a correct name "ConnectionProvider" sounds much more correct (as if we want to cater for Transactions then we don't want to create a new on) and I believe this would fit as you could UseMsSqlOutbox and inject a provider type of EF, or Dapper or ADO.  However I suppose your point of using the connection not abstraction out type, The way I see it we have a few options here

- Use Providers and leave it up to the Developers to Implement their own Provider to get the connection from their provider for the specified Outbox
- Create and EF (or Dapper) Outbox (Generic) and then provide the entities for Devs to use in their Db Context (as to get the Transaction they need to be inside the same DB Context)

If we go with Approach 1 we just need the Provider to be able to grab a Connection and existing Transaction from the provider, but wouldn't be using said provider to do the Workings. 

I do agree that clear/Sweep shouldn't be in the same transaction, my questions was should we be looking to sweep on each clear or leave it to the sweeper.

I think this might be easier discussed in a short Zoom call 😊

iancooper:
> I think this might be easier discussed in a short Zoom call 😊

Agreed, let's agree a time via WhatsApp. I may need to review first, so probably tomorrow earliest

preardon:
#1599

preardon:
@iancooper @holytshirt I've got this into a bit of a Better state,

The Instantiation looks as follows now

`
services.AddBrighter(options =>
                    {
                        options.PolicyRegistry = policyRegistry;
                    })
                        .UseInMemoryOutbox()
                        .UseExternalBus(producer)
                        .MapperRegistryFromAssemblies(typeof(GreetingEvent).Assembly);
`

If you don't specify a UseOutbox then it defaults to In Memory

iancooper:
> @iancooper @holytshirt I've got this into a bit of a Better state,
> 
> The Instantiation looks as follows now
> 
> `services.AddBrighter(options => { options.PolicyRegistry = policyRegistry; }) .UseInMemoryOutbox() .UseExternalBus(producer) .MapperRegistryFromAssemblies(typeof(GreetingEvent).Assembly);`
> 
> If you don't specify a UseOutbox then it defaults to In Memory

That looks much more coherent to me at least.

iancooper (approved):

iancooper:
I've merged these, we can iterate if need be

Issue #1602, opened 2021-06-21
When ASB Transport is Unable to Acknowledge the Lock (Due to Expiry) the Message Pump has an Unhandled error


preardon:
When using an ASB Transport if it takes longer than the Lock time on the Subscription to Complete the Handle the Command then an uncaught Microsoft.Azure.ServiceBus.MessageLockLostException is Thrown

@iancooper Is this the Behaviour you would like, or would it be better to just Log the Error and continue

iancooper:
Where does the exception get thrown? I think I'd need to understand more. Some options I can think of:

* Transport: Can we extend the lease if we are not finished? Raise a channel failure exception if no.
* Handler: Should be treated as an application exception anyway, but we will try to Ack the message in that case, so Ack would need to be aware that if lease has expired do nothing, so would need to catch any expired exception specifically. We would probably have already logged the application exception from the handler, so we might not even need to log in Ack that we did not Ack as expired, or at least do so at Info level only

iancooper:
@preardon OK, seen the PR. We might want to understand the trade-offs on catching all exceptions as opposed to this one which we know is fine

preardon:
The very last thing the Message pump does in its loop is Acknowledge the Message (this is not in a try catch block).

we have already lost the lock by this time, I was considering a Channel Failure exception

it is the act of ACKing the message that gives us this error (ACK after least expiry) we are logging the error in the Catch

preardon:
@iancooper I've updated this to catch and not throw on this exact exception.  However one thing that is on my mind around here is that if there is another error this will kill this performer, do we want to wrap the Acknowledge in the Pump so that we can apply logic here (Channel Failure) and not kill the Performer?

iancooper:
it's a tricky one as the trade-off here is 'fail fast and surface it' over log and keep working. For now, I'm inclined to only "log and keep working" for exceptions we predict that we know we can ignore

PR #1618, opened 2021-06-29
Removed Throw from the Acknowledge on ASB Comsumer

preardon:
#1602 This was killing the Pump if Processing exceeded the Lock time


iancooper:
Is it worth catching a specific exception that we believe is the issue when a lock has expired here, or do you want a catch all here?

iancooper (approved):

PR #1635, opened 2021-07-18
Add an Override Box Connection Provider

preardon:
@iancooper Here is my first whack at this just wanted to get it in front of you before going much further,

Not sure how Happy I am at calling the Type IAmABoxConnectionProvider as we are only using it to override (If available) on Deposit so perhaps IAmAnOverringBoxConnectionProjecter, or perhaps something with the word Scoped in it given that is our intention.

Also I need to give this a bit of a run through in Service Activator as I can see issues if you live in a world where you want scoped and non Scoped but there will be ways of getting around that just will be a current issue.

From my tests this does work better than before which is good.

Also what are you thoughts about throwing a Bool on Deposit post to disallow override, at the minute I'm thinking YAGNI however worth a discussion.

Let me know what you think


iancooper:
@preardon Thanks mate. I'll take a look

preardon:
@iancooper Thanks Mate let me know, I'm looking at cleaning it up a little and making sure I've rebased correctly, this is the one I did off of your old code and it looks to have 2 of your commits that haven't made it to master as yet. but if you could let me know if you are happy with the approach I'll get it clean and in

iancooper:
@preardon That is strange, as I squashed and merged to master so I am not sure where those missing commits come from.

iancooper:
@preardon This seems to be on the right lines. I've scanned it, and the idea is there: the 'scoped' command processor has the scopes provider; it passes the scope across to the bus when it calls it to add to the outbox; we use that scope to participate in the transaction..

In terms of naming, it would seem that the key issue is you want to participate in the transaction for that scope, and you need  to get the transaction being used, so I would probably name it:

IAmATransactionConnectionProvider

or

IAmAScopeConnectionProvider

I lean towards the former, but the latter might be more obvious to folks using ServiceCollection Scopes with EF, who would seem to be the primary users of this facility right now.

iancooper:
I don't think we need to disallow override on the Deposit btw. Either your command processor is scope aware through this or its not. I think that is the choice

PR #1667, opened 2021-08-10
Outbox scoped provider

preardon:
@iancooper this is the new draft PR, spent some time re acquainting myself with what was going on here.

I propose IAmABoxTransactionConnectionProvider as the name (as it deals with Inbox and outbox)

There are a couple of things going on here that I've had to remind myself

1. The Ability to specify a Different Connection on Adding (so that you can share in a Transaction)
2. The Idea of a "Shared" connection so that you don't dispose the object between calls (At the minute a single Post will dispose and Re Open the connection multiple times)

I have also added the update to the Interface to the Other providers even though they do not yet do anything with them (Will need to look into this)

What are your thoughts on the two use cases above, this was previously done and in main before we split out the external bus, I see some value in here as we will probably want to use this optimisation later on but it does leave me thinking about  splitting the connection provider interfaces into 2 (Standard and Transaction) I suppose I should probably try to get something similar working for MySql and see how it shakes down.

### DIFF
--- a/samples/ASBTaskQueueExamples/GreetingsReceiverConsole/Program.cs
+++ b/samples/ASBTaskQueueExamples/GreetingsReceiverConsole/Program.cs
@@ -46,7 +46,7 @@ namespace GreetingsReceiverConsole
                     };
 
                     //create the gateway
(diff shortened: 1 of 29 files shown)


preardon:
Rebased, should be good for a first rumble

iancooper (approved):
Thanks for this. A lot.

Issue #1634, opened 2021-07-16
CommandProcessor doesn't support Request/Reply and async handlers in the same configuration


DevJonny:
### Describe the bug
The [CommandProcessor.cs](https://github.com/BrighterCommand/Brighter/blob/a3ea14247e00b1b7c52d37e298f5be580a006504/src/Paramore.Brighter/CommandProcessorBuilder.cs#L254) constructor for Request/Reply doesn't allow for adding an `IAmAHandlerFactoryAsync.cs` instance.

Using either the `CommandProcessor` or the `CommandProcessorBuilder` you are unable to set the `IAmAHandlerFactoryAsync` and a `responseChannelFactory` in the same constructor.

I'm not sure if this is by design, or not. If it is by design, what is the recommended approach to supporting both Request/Reply and Async Handlers in the same project other than having two separate CommandProcessors.

I'm happy to pick this up if it is not by design.

iancooper:
@DevJonny Yeah, it's not by design. Things are a little better on this in 9, but it still needs work to be right. We have clarified some names, and intent there around what constructor you should use, but adding request-response on top is still not right.

Have a look if you can.

DevJonny:
Great, thanks for clarifying @iancooper I’ll definitely take a look and hopefully get a PR sorted.

PR #1685, opened 2021-08-25
Issue #1634 | Adds support for async handlers when using RPC

DevJonny:
Fix for Issue #1634 

- Adds an additional constructor for supporting sync/async handlers and RPC calls
- Fixes compile errors in tests.

### DIFF
--- a/src/Paramore.Brighter/CommandProcessor.cs
+++ b/src/Paramore.Brighter/CommandProcessor.cs
@@ -93,7 +93,7 @@ namespace Paramore.Brighter
 
         /// <summary>
         /// Initializes a new instance of the <see cref="CommandProcessor"/> class.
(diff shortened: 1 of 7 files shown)


iancooper:
@jonnyollifflee Have we looked at how this works with the ServiceCollection extensions? In other words, if you want to set up an RPC queue without building CommandProcessor directly. I may have just missed this in the PR of course.

DevJonny:
Let me double check @iancooper.

DevJonny:
@holytshirt if you're free today can we have a chat about this? I'm going around in circles a bit.

iancooper:
@DevJonny We have a 2-3PM call. @holytshirt can send you the details. Come along.

Issue #1582, opened 2021-06-13
Async Outbox Sweeper


preardon:
At the moment Brighter only has synchronous Outbox Sweeper

I would also like to be able to use it with Async Outboxes

Low Priority, I can look to get this out after v9

preardon:
Also, just a thought, is there any reason we wouldn't just make this Static (if we are just passing in All the Deps anyway)

iancooper:
So the Outbox Sweeper is intended to run on a dedicated background thread. Making it async would probably make it slower, because there would be nothing else to schedule against it. Still, I don't mind you doing an async version, I'm just not sure of the value. We could make it static rather than an instance.

preardon:
@iancooper Had some time this morning so I thought I'de get this one knocked off

preardon:
It looks like we have some Unit Tests failing on master (for MsSql Outbox)

preardon:
@iancooper Is it correct that if there is no message found in the outbox we want it to return a MT_None Messages instead of a null. I understand that in the Command Pump we want this, however its fed from a queue not an outbox?

preardon:
I've found some Bugs, I will get a fix in for these

iancooper:
I'll need to take a look. Let me stick my current work on a PR, as there is a little bit of change to OutboxSweeper

preardon:
Not a problem I'll rebase onto yours

iancooper:
See this: https://github.com/BrighterCommand/Brighter/pull/1708

I need to test the MySql stuff before it gets out of draft.

preardon:
I'll can have a fix for the MySql Sweeper in 30 minutes if you haven't done it already

preardon:
@iancooper I haven't rebased on your branch yet, would you like me to?

iancooper:
May be easier for you to go first and me to rebase

iancooper:
If you are done, we could just merge you in?

preardon:
Looks like I have a small change to make to InMemory Outbox (its never defined as Aysnc however some of the test call sweepAsync), so I'll have a look at making InMemory an Async Outbox viewer (Even though there is really no point)

Or would you prefer that In Memory to fail if you try to Async?

iancooper:
Nah, just use the TaskCompletionSource pattern to return a task on the async api implementation for the in memory message box.

iancooper:
[...]

PR #1707, opened 2021-09-15
Added Static and Async Methods to Sweeper

preardon:
Added Static and Async Methods to Sweeper as well as implementing OutstandMessagesAsync in AsyncOutboxViewer 
#1582


preardon:
@iancooper If you are Happy with this I am too now

I've fixed some issues with Outstanding message issues in outbox as well as making the Sweeper both Static and Async

Issue #1668, opened 2021-08-12
AWSSQS - Message is Acknowledged and deleted after Requeue


honkuan86:
### Describe the bug
There is an issue with the MessagePump that will Acknowledge messages after requeue. While it may work for other MessageGateway, but for AwsSqs it behave unexpectedly, because in AwsSqs, Acknowledge a message means deleting off from the queue. So the requeued message will be deleted before reach to DLQ.

Noticed the behavior changed after this commit #1544 

A indication or a flag is needed for a requeued message, so that it will not be deleted when the message is being Acknowledged.

honkuan86:
Hi @iancooper , @preardon , is this behavior expected? What I'm trying to achieve is that, when a handler throws a DerferMessageAction, I'm expecting the message to be pushed back to the queue until the max receive is reached then Sqs will move that message to DLQ. But it is not working as expected currently. Am I understand it wrongly ?

preardon:
@honkuan86 the expected behaviour is that it ACK's the message and puts in back on the back of the queue until the max retry count has been reached, then it will put it on the DLQ

iancooper:
Yeah, that could be an issue, as @preardon  suggests we would expect if you ask for a requeue we would push back onto the queue, not delete. So the place to look will be at how we have implemented requeue, where we might have just called ack, because we do that in some other transports

PR #1710, opened 2021-09-20
AWSSQS - Prevent message to be deleted after requeued

honkuan86:
This PR is related to #1668


iancooper:
My only concern here is what happens if I want to acknowledge i.e. I re-queued it, and now it is done with and I can delete it. Without a mechanism to reset the clock, won't this continue to be presented until the redrive policy forces it onto the dlq, even if I process it on a later step? Acknowledge ought to me: done with delete it. Requeue calling acknowledge is the issue I suspect, not how Acknoweldge works

iancooper (changes requested):
See the comments for concerns on this

iancooper:
Let me peer at how this works. For some brokers this has to be explicit ack when you requeue (Kafka has to increment the offset and you need to copy to a new queue, RMQ needs to delete and requeue or it just puts back on the head of the queue etc.) RMQ's redrive policy is useful, but we need to know the properties of the broker somewhere to say that the requeue on AWS does not need a delete, but that ack should still delete if it is completion

honkuan86:
@iancooper , from what I have tested is that, when a message is requeued in MessagePump, in sqs we are changing the message visibility on the queue, at this point the HandledCount will be incremented as well. The HandledCount is reset when MessagePump received the same message again from the queue.

What I'm trying to do is that to prevent the message to be deleted right after the Requeue method is invoked, when the Acknowledge method is invoked by the MessagePump. 

There might be a better way to handle this though.

honkuan86:
@iancooper , another approach is to let the MessagePump to skip the iteration in a loop when a message is requeue, then only acknowledge the message after it has been processed.

iancooper:
I'll poke around and come back with a suggestion today

iancooper:
Looking at master, and not these changes.

The only place we delete in SQSMessageConsumer.cs is Reject and Acknowledge (Ack).

We would expect to delete on an Ack, because under a notification based system, processing the message (even if it ends in an exception that we don't catch) deletes the message from the queue. Someone has handled it. Application errors belong to the application at this point - we can't know if it is poison pill and will just keep failing - so we also delete for an error. If the error is transient, retry via a Polly policy.

Now, sometimes, even after using a Polly retry, you may choose to requeue - with a delay.  Presumably, you have more information than we can have about why the code should eventually succeed. In that case you need to throw DeferMessageAction from your handler.  DeferMessageAction tells us to Requeue.

If we Requeue a message via throwing a DeferMessageAction then I believe the existing code should work as expected - the message VisibilityTimeout will be changed to requeue the message. Now we only keep requeueing up to a threshold, to avoid poison pill issues. But as an SNS redrive policy does pretty much the same thing, if we have a DLQ, even if we reach the ReqeueCount, we still just change the VisibilityTimeout over deleting.

So I believe the code works as intended - unless there is a bug in that flow.

So perhaps we need to talk about what you are experiencing which means that this is failing?

honkuan86:
@iancooper, The current flow in MessagePump (in Master) is
**Handler throws DeferMessageAction** --> **RequeueMessage** invoked --> **Acknowledge** invoked (Message deleted after Acknowledge), 

so even we have changed the VisibilityTimeout, we won't be able to see the message anymore, because it has been deleted.

This is the issue that I'm facing, or maybe I have the wrong expectation ?

iancooper:
On a RequeueMessage we should not be Acking the message, we just skip to the next iteration of the loop (from the messagepump) code:

             catch (DeferMessageAction)
                {
                    if (!RequeueMessage(message)) continue;
                }

We only return false here, and thus ack and delete, if RequeueMessage returns false

   protected bool RequeueMessage(Message message)
        {
            message.UpdateHandledCount();

            if (DiscardRequeuedMessagesEnabled())
            {
                if (message.HandledCountReached(RequeueCount))
                {
                    var originalMessageId = message.Header.Bag.ContainsKey(Message.OriginalMessageIdHeaderName) ? message.Header.Bag[Message.OriginalMessageIdHeaderName].ToString() : null;

                    s_logger.LogError(
                        "MessagePump: Have tried {RequeueCount} times to handle this message {Id}{OriginalMessageId} from {ChannelName} on thread # {ManagementThreadId}, dropping message.{5}Message Body:{Request}",
                        RequeueCount,
                        message.Id,
                        string.IsNullOrEmpty(originalMessageId)
                            ? string.Empty
                            : $" (original message id {originalMessageId})",
                        Channel.Name,
                        Thread.CurrentThread.ManagedThreadId,
                        Environment.NewLine,
                        message.Body.Value);

                    RejectMessage(message);
                    return false;
                }
            }

            s_logger.LogDebug(
                "MessagePump: Re-queueing message {Id} from {ManagementThreadId} on thread # {ChannelName}", message.Id,
                Channel.Name, Thread.CurrentThread.ManagedThreadId);

            Channel.Requeue(message, RequeueDelayInMilliseconds);
            return true;
        }

I expect the issue here is that in the above code, where there is a DLQ, we don't want to discardrequeued messages if we are relying on the AWS Redrive Policy. I guess we could test for a DLQ here to determine if we should delete, but you can probably prove this by setting your requeue count to -1 which will not discard, but will allow the AWS redrive policy to hoover up the request if the number of retries exceeds the threshold set in the redrive policy

If that is the case, we may want to look at whether it's easiest to be DLQ aware in message pump so that the check for discard messages also looks for a DLQ setting, or just force the threshold to -1 if you have a DLQ to indicate that the redrive policy owns the decision not Brighter

honkuan86:
`             catch (DeferMessageAction)
                {
                    if (!RequeueMessage(message)) continue;
                }
`
I think if we change to always skip the iteration regardless of Reject / Requeue, this will prevent message being Ack.

`catch (DeferMessageAction)
                {
                    RequeueMessage(message) continue;
                }`

This is also how the codes flow before this commit #1544

iancooper:
But if we fail to Requeue, we should ack, otherwise it will become stuck in the queue. So the better question is whether something is failing to requeue that you expect to be requeued?

What is your Requeue count setting? if you set it to -1 i.e. have Brighter give up control of deciding to delete your message because it has too many retries, does it behave as you expect?

My gut is that this is the problem as two mechanisms are trying to discard messages after a certain number of retries. AWS which will put it on a DLQ and Brighter, which will delete it. If you want the AWS DLQ you need to turn off Brighter's dropping off messages being handled too many times.

If that is the case (when you set -1 for RequeueCount it works as expected) then we can have a conversation about whether we set that for you when you tell us that you have a DLQ

honkuan86:
I'm testing with this configuration,

`new SqsSubscription<GreetingEvent>(
            new SubscriptionName("event-topic"),
            new ChannelName("event-queue"),
            new RoutingKey("event-topic"),
            bufferSize: 10,
            noOfPerformers: 1,
            timeoutInMilliseconds: 3000, lockTimeout: 30,
            makeChannels: OnMissingChannel.Assume,
            findTopicBy: TopicFindBy.Name))`

I have not set the requeueCount, so by default it is -1.

However, it is still behave as what I've mentioned where the message will still get Ack after Requeue.

iancooper:
@honkuan86 OK, let me try testing it to see what it is wrong. There must be a bug somewhere so that it is not behaving as expected. Walking the code doesn't suggest you should be losing messages.

iancooper:
@honkuan86 On this now. Going to see if I can write a test to confirm the behavior I expect it should have

iancooper:
@honkuan86 I have fixed the issue at the root of this - we should not ack on a successful requeue, but 'continue' the message pump loop instead. That change should have gone to master. Once you can verify expected behaviour we can close this PR and the issue.

Thanks for raising this, although we didn't take this fix, the underlying issue did need to be fixed now we have DLQ support across many transports.

PR #1725, opened 2021-09-22
Renames sample folders to remove sample/example

DevJonny:


iancooper:
You are the best!

DevJonny:
There was a build error in one of the same projects around `Microsoft.EntityFrameworkCore.SqlServer` version 5.0.9 -> 5.0.10 so I just pushed that to `master` so hopefully should clear the build errors.

Issue #1699, opened 2021-09-07
CommandProcess.Call() breaking due to Subscription not being RmqSubscription


DevJonny:
### Describe the bug
When CommandProcessor.Call is invoked a generic `Subscription` is created and passed to the `ChannelFactory.CreateChannel()`. The [ChannelFactory](https://github.com/BrighterCommand/Brighter/blob/e605ef111cca1c15e6a053cc51bec3087fa2a63c/src/Paramore.Brighter.MessagingGateway.RMQ/ChannelFactory.cs#L53) enforces the type of the Subscription to match the type of messaging gateway.  

### To Reproduce
This can be reproduced using the `RMQRequestReplyExample` project in Samples.

### Exceptions (if any)
```
Unhandled exception. Paramore.Brighter.ConfigurationException: We expect an RmqSubscription or RmqSubscription<T> as a parameter
   at Paramore.Brighter.MessagingGateway.RMQ.ChannelFactory.CreateChannel(Subscription subscription) in D:\dev\huddle\Brighter\src\Paramore.Brighter.MessagingGateway.RMQ\ChannelFactory.cs:line 57
   at Paramore.Brighter.CommandProcessor.Call[T,TResponse](T request, Int32 timeOutInMilliseconds) in D:\dev\huddle\Brighter\src\Paramore.Brighter\CommandProcessor.cs:line 724
   at GreetingsSender.Program.Main(String[] args) in D:\dev\huddle\Brighter\samples\RMQRequestReplyExamples\GreetingsClient\Program.cs:line 75
```

DevJonny:
@iancooper was `.Call()` designed only for RMQ or all Messaging Gateways. Just working through ways of fixing this and the check for the right type of subscription is trashing most of my ideas.

One I did have was having a separate method on the `IAmAChannelFactory` that would allow for a simpler set of values to be passed to create the temporary queues, but I'm not sure that will work for all Messaging Gateways, but I'm most familiar with RMQ and a much lesser degree SNS/SQS. I'll keep having a play around and see what else I can come up with.

iancooper:
@DevJonny It ought to work for things like SNS+SQS too, but I can't recall now how far I got beyond making RMQ work TBH We should not close out that option. The Channel Factory idea does have some merits

PR #1726, opened 2021-09-22
Fixes Issue #1699 | CommandProcessor.Call(..) works again

DevJonny:
CommandProcessor.Call(..) wasn't working because the individual gateways didn't support the generic `Subscription` only specific for that gateway. The following changes have been made:

- `ServiceCollectionExtensions.UseExternalBus(..)` now takes in an optional `IEnumerable<Subscription>` for each reply queue type
- On invocation of `.Call(..)` the appropriate Subscription is grabbed, and the `RoutingKey` and `ChannelName` are set to a new GUID
- Fixes `ServiceCollectionsExtensions` to supply an Outbox when using RPC
- Updates the `RMQRequestReply` sample to use the new changes


iancooper:
Thanks. This is a great fix.

iancooper (approved):

PR #1729, opened 2021-09-25
Reduce command processor ctor's

preardon:
@iancooper Here is a first go

One test failing, interesting one, it appears that when you "Mark Dispatched" if your producer supports Publish Confirmations it goes down that route weather its configured or not

### DIFF
--- a/samples/KafkaPartitionTesting/GreetingsSender/Adapters/Program.cs
+++ b/samples/KafkaPartitionTesting/GreetingsSender/Adapters/Program.cs
@@ -115,7 +115,7 @@ namespace GreetingsSender.Adapters
                             })
                         .Create();
 
-                    services.AddSingleton<IAmAMessageProducer>(producer);
+                    services.AddSingleton<IAmAMessageProducerSync>(producer);
 
                     services.AddBrighter(options =>
                     {
--- a/samples/KafkaTaskQueue/GreetingsSender/Adapters/Program.cs
+++ b/samples/KafkaTaskQueue/GreetingsSender/Adapters/Program.cs
@@ -114,7 +114,7 @@ namespace GreetingsSender.Adapters
                             })
                         .Create();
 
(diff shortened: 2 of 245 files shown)


iancooper:
Looking at this I wonder if the answer is to go further.

We could treat IAmAn**** as a base 'marker' interface with no methods, and have explicit IAmAn***Sync and IAmAn***Async that derive from it. So IAmAnOutbox can have IAmAnOutboxSync and  IAmAnOutboxAsync.

Then the CommandProcessor constructor only ever takes the base IAmAn****, for example  IAmAnOutbox and we **always** query the interfaces that you support i.e. sync and async, thus removing any need to repeat the constructors for both types but supporting you supplying us an implementation that supports one or both.

That seems to give flexibility to everyone, the only thing it blocks is entirely separate codebases for sync and async but I think that edge case will be sufficiently unusual to not support it.

What do you think?

preardon:
If we wanted to go down this path I see only two options, either do as you say to have a common Interface and then split sync and Async, or update all interfaces to contain both sync and Async.

I see using the common interface with both sync and Async ones as easiest

preardon:
@iancooper There is 1 Failing test that I would like you to look at,

When_Posting_Via_A_Control_Bus_Sender_Async (ControlBusSenderPostMessageAsyncTests.cs)

No sure why it was working before, however the current logic states that if the Sync Message Producer is ISupportPublishConfirmation then don't mark as dispatched. I understand that for RPC we don't particularly want to mark as dispatched however this logic means that if our producer supports it (and we don't want to use RPC) that it all always want to go down this path

iancooper:
@preardon Let me take a look, and see if I can figure out why that one is miserable

iancooper:
@preardon Going to merge and review the test after, as that will be easier I think at this point. I think this is a significant step towards reducing the api complexity, so let's make progress with it.

PR #1746, opened 2021-10-05
Serviceactivator scope declaration - using scopes in PipelineBuilder

penderi:
DRAFT

Issue - #1560 

@iancooper apologies for the delay. But at least I have something. Although I'm reticent. 

The best thing about this PR is it isolates the scope to what I think is the ideal esp for a worker - scope to an Event or Command received on by the Dispatcher. This is by creating the scope for each pipeline (in PipelineBuilder) - thereby covering all pre/post attributes and handlers.

I have *tried* to reduce concerns such as:
- reduce change to core interfaces
- keep scoped changes in service activator as much as possible
- keep it simple
- allow it to be extended

The are 2 real, well, turds to be frank:
- Use of IServiceScope in the core lib. I know I can adapt our way out of that so no biggie right now.
- The addition of lifetimeScope to *all* handler factories, but only used by Scoped

That last one could be mitigated by making a scope factory / DI aware but that starts snowballing change.

Comments awaited (dons helmet)


penderi:
@iancooper ah shoot mate. I'm sorry I should have sign-posted better! This is the one and please do delete

iancooper:
@penderi If you take a look for me, we can figure out what we want to do to merge this

iancooper:
@preardon OK, let's go this way, I don't have a strong opinion, so happy to go with what everyone else intuitively thinks, as I  may be the outlier.

iancooper:
It would be handy to have at least one example that shows it, but only if it supplements not replaces @preardon's one :-D

iancooper:
I'm traveling to NDC TechTown and back this week, so I may get some airport coding time.

Issue #1760, opened 2021-10-16
Have the TimedOutboxSweeper Configurable


preardon:
Currently when registering the TimedOutboxSweeper there is no way to configure the time between ticks (Currently defaulted to 5 seconds) and the age of a message before sweeping (Currently defaulted to 5000ms)

I was considering adding IOptions so that we can have these configurable

PR #1761, opened 2021-10-17
Feature/1760 make sweeper service configurable

preardon:
This should do what it says on the tin, I have set the defaults to what was previously in the code

@iancooper how do you feel about the Property names "TimerInterval" and "MinimumMessageAge" I am not wedded to them

#1760


iancooper:
Good job, thanks

PR #1779, opened 2021-10-21
Sample of scoped handler factory with ASB

penderi:
This PR adds a new sample for the ASB Task Queues - a scoped receiver - to demonstrate using scoped handler factories to receive ASB messages. The example demonstrates a shared scoped component (a counter) used in both an attribute and handler. Key results:
- The scoped instance is refreshed for each dispatch (event/command)
- The scoped instance is shared in the pipeline

We can also compare behaviours against the non-scoped receiver - in the ASBTaskQueue folder the receiver is configured to the same topic. We can see the scoped counter is shared *across dispatched events*, which highlights the issue that scoped instances were becoming singletons (!).

Note: on adding a new project in VS the project GUIDs changed in the solution file, bloating the commit.


preardon:
Looks alright to me

Issue #1797, opened 2021-10-26
When using a Scoped Box Transaction provider with a Scoped Service Activator the Transaction Provider can not be resolved


preardon:
When Adding a Scoped Transaction Provider with a Scoped Service Activator the Scope is not used when trying to resolve the Scoped Box Transaction Connection Provider

preardon:
Ok, I have done a bit of testing and it looks to work.

When Use Scoped is true it works and I can observer separate scopes.

When Use Scoped is false it doesn't fail how I expect (i.e. it appears to resolve a Dependency instead of throwing) however there are a few other issues that occur later, also I Observe Objects living longer than the the scope (As Expected)

I can confirm that the Outbox Participates in the Transactions as well.

PR #1801, opened 2021-10-29
Added CommandProcessorProvider

preardon:
to look at fixing #1797 

- Created IAmACommandProcessorProvider in ServiceActivator to allow flexibility around how the CommandProcessor is provider
- Removed the Scoped Handler Factory for now (Still need to do more testing however my working theory is that if we are using a scoped  Command Provider this should hopefully trickle down into the dependencies in it)

My First Tests Look good however I want to do some more in depth ones

@iancooper let me know your thoughts so far, I will also discuss this with @penderi around the scoped handler provider stuff


iancooper:
@preardon Just got internet back after moving; will look tomorrow morning.

preardon:
@iancooper not a problem, I've done some bigger examples and it works.... Even when it's not supposed to 🤣 so I will work out why and push my examples

iancooper:
Hi @preardon,

So my assumption is that you have picked up on the change we made to create an External Service Bus, to build a scoped command processor per request, because this mimics what we are doing for scoped dependencies in ASP.NET anyway. And once the CommandProcessor has a scope, we can move away from the Scoped Handler Factory solution because we already handle that by virtue of it pulling in a CommandProcessor for that scope?

I certainly like the synergy between the two hosts (ASP.NET and ServiceActivator) as it would potentially remove the complexity about anything that needs to treat them differently. So I like this idea a lot.

I think we need to:

- Move the code inside DispatchRequest (which may require derived classes to override a method we call into inside) as Dispatch is where we execute the pipeline.
- Consider if we should remove the scoped handler factory, with all apologies to @penderi because at that point they become confusing complexity that we don't need. That's on me, for suggesting that we solve it that way.

But I think you are right, that having made the ExternalServiceBus change, this is a bettter option.

preardon:
@iancooper that is correct and my current working theory, that if we declare the scope around a single instance of the Command processor (Transient) then any IServiceProvider injected under that scope will be inside that scope. Also if you don't care about scopes and use my unscoped CommandProcessorProvider then it should create you one per instance of the pump and provide the same one each time you ask (mimicking what we have already). I also quite like that all changes were able to be contained inside ServiceActivator or its extension libraries.

I will move to code inside dispatch as requested as I agree it is more correct there. And given this is all done on provider it will be relatively easy for other DI types to implement.

 This is exactly why I wanted to talk to @penderi about what I've done here as it does make some of his work redundant. I'm also having an issue at the minute where if I UseScoped to false it still works so need to track that down

I have created a new "worker" example that effectively receives a Command, stores a value in a Db and emits an event so is a mix of scopes, box scoped transaction provider, and receiver. I am hoping to build this up a little so that it's a good test bed for these slightly more complex scenarios as well as a decent example / quick start for users who want a worker process with EF ect.

iancooper:
@preardon I think this is an improvement over my plan that @penderi worked with, as it better exploits that split. Let's head down this path instead as I would always favor simple where we can.

iancooper (approved):
Looks good to me, has the commandprocessor selection within the Dispatch now as dicussed etc.

PR #1854, opened 2021-11-16
Fixed error Handling around ASB Acknowledge

preardon:
This will fix the ASB Acknowledge to catch correct Exception


holytshirt:
Fixed missing bracket

PR #1926, opened 2021-12-29
Clear Outbox Async Checking Sync Producer

preardon:
This should be checking the correct Producer


iancooper:
I think I might have picked that up with the producer changes anyway, but perhaps get this in for when I merge, so that I can review.

iancooper (approved):

Issue #1764, opened 2021-10-18
Outbox Sweeper - Bulk Operations


preardon:
At them minute when calling Sweep on the outbox the code path is

- Get a list of the first 100 dispatched messages older than a certain age
- Past the IDs of those messages into the Clear Function which then does a ForEach loop of
- Get the Individual Message
- Send the Individual Message
- Mark the Individual Message (if post back is not supported)

This means Per Clear operation we are doing up to 201 SQL calls and 100 Bus calls

Considering the Outbox supports Bulk get if we were to implement bulk message send on our Producers and Bulk Mark Dispatched we could get this number down from a potential 2 SQL calls and 1 bus call

I know this isn't a high propriety, and to be honest at the time of starting to write this I thought we already had a Bulk Send so its a little more work that I was thinking.

Also I think it could be good to make the max amount of messages to sweep configurable

@iancooper Thoughts

preardon:
Agreed, lets put it on the 9.1 backlog

PR #1927, opened 2021-12-29
CommandProcessor Bulk Clear #1764

preardon:
This should allow the Command Processor to Bulk Clear from the Outbox and send Messages to Transports in Batches

@iancooper I would appreciate you're thoughts on this when you have a second before I get too much deeper, however I understand that this is for v9.next


iancooper:
This looks fairly solid and I don't have any issues with how this is working. We may want to make sure that a publisher confirm works with this, but that would mean we would need to update the transport for RMQ etc, and I am fine with doing that after you get ASB in

preardon:
@iancooper thanks for this, I will continue down this path and get a PR in

preardon:
This Adds the following Functionality

- Outbox - Get Multiple by ID
- Outbox Mark Multiple as Dispatched by ID
- Command Processor - Bulk Clear
- New Interface for Producers that support Bulk
- Outbox Sweeper - Ability to set to Use Bulk

@iancooper this should be a MVP to allow ASB to dispatch as Bulk, let me know what you think

iancooper (approved):
I think this works. What we might be able to look at is the same model we use elsewhere - launch a thread to run the bulk update and use a mutex to prevent re-entry if a bulk clear is already running. But I can point you in the direction of our prior art there.

PR #1936, opened 2022-01-10
Postgres outbox transaction provider

### LINKED ISSUE #1838: PostgreSql outbox transaction support
### **Problem**
It is not possible to include events that are written to the PostgreSql outbox table within the same transaction as the entity writes.

### **Solution**
Make use of the connection/transaction provider patterns, so that it is possible to ensure that events added to the PostgreSql outbox table occur within the same transaction as the entity writes.

samrumley88:
PostgreSql Inbox updated to support connection provider pattern.

PostgreSql Outbox updated to support transaction connection provider pattern (inc EntityFramework connection provider). 

I've tried to follow the existing pattern which has been applied to MSSql, MySql and SqLite. @preardon not sure if this is of any interest to you?

Closes #1838


preardon (approved):
looks good to me, thanks for doing this

PR #1943, opened 2022-01-14
Add LookupByOrDefault to the Producer Registry to allow setting Default

preardon:


iancooper (approved):
This works, although we may want to consider how to provide a 'common' producer better as the GetDefaultProducer is itself a 'hack' designed to get around the issue that Outbox configuration properties are on the publication, not an Outbox config, so we have coupled this to something we intend to remove <g>. Not too worried as when we figure out both those issues we will hopefully spot the note on the code and kill it

PR #1965, opened 2022-02-01
Extend Web EF Core example to show scopes in Host Builder with Brighter

iancooper:
This is about showing how a worker process, listening for messages, can use our support for scopes.

It will support:

* Inbox - to show setting this up
* Outbox - to show using Outbox from within a worker process
* EF Core - as this will be the majority ORM

It won't _require_ a release, although it does add some missing options for checking that Inbox or Outbox exist, that others might appreciate, so it's been flagged a 9.x


iancooper:
Currently a WiP, will move take draft status of once it is good

PR #1977, opened 2022-02-08
Move the sweeping logic into the Command Processor

preardon:
#1764


preardon:
After our discussion this does the following
  - Removes the IAmAnOutboxViewer interfaces and moves all of their functionality into the Outboxes
  - Added Logic to Clear Outstanding Messages to the ExternalBus Service and plumbed through to the Command Processor
  - Added a semaphore to External Bus so that we can run one clear operation at a time (To Try to prevent messages being sent twice)
  - Updated Sweeper service to use the new
  - Allowed bulk send on Async Implicit Clear

preardon:
@iancooper @holytshirt Should be good for review now, let me know what you think

iancooper (approved):
This works but the other option is to use a thread for clearing. It would always return a Task and is thus could be used async or fire and forget but would not depend on async I/O as it would be non-blocking. I'm not sure that you would ever want to wait for the outbox to clear, unless you were clearing specific Guids i.e. I know which messages I want to force (as today)

iancooper:
Not sure about Sync requiring Async Outbox, it would be nice to avoid that. Getting a little hard to read in this form though, so I don't mind if you want to go to master, then raise a new PR to address that - it might be easier for us all to review and provide feedback from that point, as the rest of the changes are valuable

PR #1980, opened 2022-02-11
Pipeline failures are ConfigurationExceptions

iancooper:
If we allow pipeline builder errors to surface without wrapping them up in a ConfigurationException, we can't separate them in the message pump from application errors. As such we ack the message, take the next item on the queue, and then swallow that application exception, ack the message, take the... and so on, until we eat the queue without actioning it. Worse, this failure can be silent if our logging is not at a detailed enough level.

This fix puts in a backstop exception handler that wraps all pipeline build issues with a ConfigurationException to ensure that we have the correct behavior - terminate the pump - when we fail to build the pipeline (usually due to misconfiguration of the handler pipeline of missing IoC dependencies).


preardon (approved):
Looks good, from what I can see this will Reject the Message (Only problematic if you've configured your rules wrong, I suppose it will DLQ) and then break

iancooper:
> Looks good, from what I can see this will Reject the Message (Only problematic if you've configured your rules wrong, I suppose it will DLQ) and then break

Yeah, so the situation before was that it silently dropped all the messages on the floor, so this is much better :-D

PR #1981, opened 2022-02-11
Moved the Implicit clear to not cross sync and Async

preardon:
I have noted that they both run as Background threads and change ClearOutboxAsync to ClearAsyncOutbox

### DIFF
--- a/src/Paramore.Brighter/CommandProcessor.cs
+++ b/src/Paramore.Brighter/CommandProcessor.cs
@@ -520,13 +520,14 @@ namespace Paramore.Brighter
         
         /// <summary>
         /// Flushes any outstanding message box message to the broker.
+        /// This will be run on a background task.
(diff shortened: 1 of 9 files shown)


iancooper (approved):
Looks much better

preardon:
@iancooper can you please make sure you are happy with the in memory unit test changes. Usually I'd be happy but I changed the results in a unit test

iancooper:
So the changed test implies that "when they are too new don't sweep". So you might have changed that behavior, which is why the test fails, but it could be fine. Let's figure it out. Under the approach I tended to use, I would clear explicitly and sweep in the background. To avoid the double posting issue, it would look for items that seemed to be hanging around after a few retries. So this test says, if the items in the box are too recent, they should not have been posted.

By changing posted to "3" you no longer check that this is too soon to post.

What is your intent here though?

PR #1988, opened 2022-02-15
Add MQTT Gateway Support

timvw74:
Starter for MQTT Gateway support.

The Consumer Client keeps a local `Systems.Collections.Generic.Queue` , due to MQTT's "push" technology.

Not implemented: 
`IAmAMessageConsumer.Acknowledge`
`IAmAMessageConsumer.Reject`
`IAmAMessageConsumer.Requeue`


src/Paramore.Brighter.MessagingGateway.MQTT/MQTTMessagingGatewayConfiguration.cs
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+
+    public class MQTTMessagingGatewayProducerConfiguration : MQTTMessagingGatewayConfiguration
+    {
+
+    }
+    public class MQTTMessagingGatewayConsumerConfiguration : MQTTMessagingGatewayConfiguration
+    {
+        public Queue<Message> queue { get; set; }
```
iancooper, line 23: I'm not sure about passing the queue in on the config. It feels as though the interface is not clean as a result. Is this for testing?

iancooper (commented):
Some notes 

- We try to document all public methods with a view to what Intellisense will give you; we don't document internal or private ones

iancooper:
Minor remarks though.  Once we get this cleaned up, as you are the only person working on it, we can pretty much wave your PRs through if they are just for this transport, with feedback so you can iterate

src/Paramore.Brighter.MessagingGateway.MQTT/MQTTMessagingGatewayConfiguration.cs
timvw74: This was added to be able to pass in an external Queue. I have removed it as it can be added later if needed.

iancooper (approved):

src/Paramore.Brighter.MessagingGateway.MQTT/MQTTMessagingGatewayConfiguration.cs
iancooper: Let's think about our alternatives here, as it's a "surprise" coming from other transports.

iancooper: Not a blocker to going forward though.

iancooper:
@timvw74 I can merge if you would like. We can ignore the failing AWS tests for SNS/SQS

iancooper:
@timvw74 I converted to a draft whilst we sort out your build

iancooper:
Just a note that we tend to disable test parallelization for XUnit by putting tests into a collection.

PR #1989, opened 2022-02-15
Fix Unit Tests for Sweeper

preardon:


iancooper (approved):
Sometimes we flag categorize these if they keep failing on the CI build due to delays, but let's give it a try without that

PR #1993, opened 2022-02-16
Support Fluent Configuration of Our Inbox

iancooper:
The improvements to MessageStore - now Outbox - have really helped to surface this important tool. But when we use an Outbox we may also need an Inbox, to remove duplication (an outbox is guaranteed **at least once**). However, our Inbox configuration has lagged the Outbox.

This PR makes Inbox feel more first class by adding an Use***Inbox set of methods (internal and external). It also surfaces the global inbox settings better, so folks know how to turn it on for all their handlers.

Finally, it does a little refactoring to make some of the code clearer where it deals with multiple configuration options and has to decide how to build the external bus.
 
It alters the IAmAnInbox interface, breaking out IAmAnInboxSync and IAmAnInboxAsync to mirror the approach used by Outbox. This results in a breaking change to that interface.

### DIFF
--- a/src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
@@ -99,16 +99,19 @@ namespace Paramore.Brighter.Extensions.DependencyInjection
          /// <param name="inbox">The external inbox to use</param>
          /// <param name="inboxConfiguration">If this is null, configure by hand, if not, will auto-add inbox to handlers</param>
          /// <returns></returns>
-         public static IBrighterBuilder UseExternalInbox(this IBrighterBuilder brighterBuilder, IAmAnInbox inbox, InboxConfiguration inboxConfiguration = null)
+         public static IBrighterBuilder UseExternalInbox(
+             this IBrighterBuilder brighterBuilder, 
+             IAmAnInbox inbox, InboxConfiguration inboxConfiguration = null, 
+             ServiceLifetime serviceLifetime = ServiceLifetime.Scoped)
          {
              if (inbox is IAmAnInboxSync)
              {
-                 brighterBuilder.Services.Add(new ServiceDescriptor(typeof(IAmAnInboxSync), _ => inbox, ServiceLifetime.Singleton));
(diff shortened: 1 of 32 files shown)


src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -76,26 +75,56 @@ public static IBrighterBuilder UseExternalOutbox(this IBrighterBuilder brighterB
...
+         /// model; an internal inbox is useful for testing but outside of single consumer scenarios won't work as intended
+         /// If not null, registers singletons with the service collecion :-
+         ///  - IAmAnInboxSync - what messages have we received
+         ///  - IAmAnInboxAsync - what messages have we received (async pipeline compatible)
+         /// </summary>
+         /// <param name="brighterBuilder">Extension method to support a fluent interface</param>
+         /// <param name="inbox">The external inbox to use</param>
+         /// <param name="inboxConfiguration">If this is null, configure by hand, if not, will auto-add inbox to handlers</param>
+         /// <returns></returns>
+         public static IBrighterBuilder UseExternalInbox(this IBrighterBuilder brighterBuilder, IAmAnInbox inbox, InboxConfiguration inboxConfiguration = null)
```
preardon, line 102: While I see we have UseExternalOutbox form what I can see in the code its not used anyway and instead we appear to be using the Specific ones (UseInMemoryOutbox / UseMsSqlOutbox / UsePostgreSqlOutbox / ect.) 

I am a fan of having the specific ones as its a little easier to read, however I see That when I put in UseMsSqlOutbox I added a reference from Paramore.Brighter.Outbox.MsSql to Paramore.Brighter.DependencyInjection meaning that if you want to use MsSql outbox and an Microsoft DI library you are going to have to pull in the Microsoft DI DLL.

preardon (approved):
Looks good to me, just the one comment on the Use Statement and my preference for specific ones instead of the generic one

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
iancooper: In this case in particular it reflects that in sample code I use Sqlite for development and MySQL for Production. Now, I know some folks don't like that, but... that kind of scenario is much easier if you don't have a specific override

PR #1995, opened 2022-02-18
Fix Threading issue on Implicit Clear

preardon:
This change fixes the following
  - Bulk Clear was using ProducerRegistry.LookupBy not LookupByDefailt
  - In the Implicit clearing Threads the Monitor was never releasing so it would never allow entry after the first run
  - TimedOutboxSweeper would Fail on Async as it can't be RunSynchronously ad the thread was usually already exited

### DIFF
--- a/Brighter.sln
+++ b/Brighter.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.28922.388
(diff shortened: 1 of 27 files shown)


preardon:
Also I have removed the (now) unused OutboxSweeper.SweepAsync

iancooper (approved):
I guess we don't always come across the issue with monitor having thread affinity and async, so it's interesting to see it pop up for us in the wild. But this seems to be the preferred alternative so...

PR #2002, opened 2022-02-22
Web and EF Core Example

iancooper:
Work on an fully-featured Web and EF Core example with RMQ for the transport.

Designed to help folks "get started" and provide a useful "test" for us that we work.

Still on .NET 5.0, will go to NET 6.0 later, once we believe audience has moved


EDuToit:
Hi, I noticed the Dockerfile and docker-compose files as well as the build.sh does not match the new assembly names.

PR #2005, opened 2022-02-25
Add Chained Credentials for Azure

preardon:
This PR adds a chained Credential for Azure to both MsSql and ASB


iancooper (approved):
Looks fine. I can't recall, does this lead to the SqlClient version issues getting fixed?

preardon:
@iancooper that's another one, this just allows you to have a list of different token providers that you may want to use

PR #2006, opened 2022-02-26
fix: logger instantiation following application logging pattern

EDuToit:
I am sold on the idea of brighter and want to see if I can contribute where I can. For the beginning, I reviewed what is currently ongoing and as such "reviewed" the progress of the redis_streams, while addressing some ide suggestions.

I am not sure what the take on Newtonsoft is from Brighter's perspective, looking at some previous changes, I am assuming that it is moving towards System. Text's implementation.

I also noticed the ApplicationLogger being used elsewhere.

I am slowly going to try and familiarize myself with the project.


iancooper:
Hi @EDuToit thanks for the PR, which I will merge. The streams producer is a WiP, which got parked whilst we got v9 out, and is still probably sitting behind some of the docs. Happy to have help, I might need to review where we are and update the issue so that it's clearer what is outstanding on it.

PR #2007, opened 2022-02-26
Add the option to turn on or of encryption on MsSql Server Connection

preardon:
@iancooper This is what is required to upgrade to Microsoft.Data.Sql > 4.0.0

- Removes a few unnecessary references 
- Add an option to specify if sql connection encryption should be on or not

### DIFF
--- a/src/Paramore.Brighter.Inbox.MsSql/Paramore.Brighter.Inbox.MsSql.csproj
+++ b/src/Paramore.Brighter.Inbox.MsSql/Paramore.Brighter.Inbox.MsSql.csproj
@@ -15,9 +15,4 @@
     <ProjectReference Include="..\Paramore.Brighter.MsSql\Paramore.Brighter.MsSql.csproj" />
     <ProjectReference Include="..\Paramore.Brighter\Paramore.Brighter.csproj" />
   </ItemGroup>
-  <ItemGroup>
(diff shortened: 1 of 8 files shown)


iancooper (approved):
Got it, thanks for fixing this

preardon:
@iancooper this is a weird one for me, Microsoft have changed the default behaviour of this library from 4.0.0 to default encryption to on (previously off), this issue with this of course is that unless you have your servers setup with trusted certificates then this will fail.

The real question here is should we offer this override to connection string that is passed in to turn this on/off. There are two other options here

1. Do nothing, and advise our users that if they are not connecting to a SqlServer with a trusted connection add "Encrypt=false;" to their connection strings, which feels like a bit of a breaking change to me (however I imagine this will effect the rest of their Sql connections as we have just forced them to upgrade to 4.1.0 (Entity Framework Core only required > 2.1.4)
2. Completely change the MsSqlConfiguration Class so that you don't pass in connection string, but rather the elements and place the ConnectionStringBuilder in there, however this is a bit of a breaking change as well

I suppose what I've done here feels like the least breaking change, but I don't love the fact that if you are already passing in a Connection String with Encrypt=True and upgrade to this version we will turn it off unless you specify that it should be on

I would appreciate your thoughts :) I think after typing this novella, I'm leaning towards throwing most of the PR away and going with the Do Nothing approach.

PR #2054, opened 2022-03-29
Update documentation for EFCore_WebApi example

omelianlevkovych:
# Description
Updated the documentation readme file for the example project.

## Reasoning behind the change
While doing onboarding I have faced few obstacles. 
It was considered that documentation update will be helpful. Therefore newcomers will not have to face same issues.

## Additional changes
I have added table of content and fixed some typos.

### DIFF
--- a/samples/WebAPI_EFCore/README.md
+++ b/samples/WebAPI_EFCore/README.md
@@ -79,6 +79,21 @@ Queues are created by consumers. This is because publishers don't know who consu
 Generally, the rule of thumb is: start the consumer and *then* start the producer.
 
 You can spot this by looking in the [RabbitMQ Management console](http://localhost:1567) and noting that no queue is bound to the routing key in th exchange.
+You can use default credentials for the RabbitMQ Management console:
+```sh
+user :guest
+passowrd: guest
+```
+### Connection issue with the RabbitMQ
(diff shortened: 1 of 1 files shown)


omelianlevkovych:
Unfortunately, I haven't really used much github yet. In case I do need to link an issue somehow here, please, let me know.

Also I have a few additional questions.
* We do have docker-compose file in the example folder, however I failed to run it. What I am doing wrong, and what is the initial purpose of that file (we do have docker compose files in root folder already).
* In case it the docker compose file somehow can be used, can we add phpmyadmin as well. The GUI can be really helpful for folks who haven't used the MySQL a lot yet.
* We also have a build.sh file. Do we want powershell scrip as well for the Windows users. Is it even a forbidden question in the open source community :laughing: ? I am not sure whether it is possible to run build.sh script on Windows OS.

iancooper (approved):
These look really helpful

iancooper:
>We do have docker-compose file in the example folder, however I failed to run it.
OK, I'll add it to my list of things to investigate, alongside the failing global inbox switch

>In case it the docker compose file somehow can be used, can we add phpmyadmin as well.
Not used that before - so something for me to investigate (I tend to use DataGrip)

>We also have a build.sh file. Do we want powershell scrip as well for the Windows users. 
We can, I am just on a Mac so it would be hard to test. Some of the other maintainers are on Windows machines though. We don't have a preference as to the OS :-D

preardon (approved):
Looks, good to me, thanks for putting the effort in :)

PR #2076, opened 2022-04-16
Webapi marten example DRAFT

omelianlevkovych:
This is draft pull request to get earlier opinions, thoughts whether I am going into the right direction.

Currently, only the published (GreetingsAPI) is almost completed.
Still luck RMQ setup, outbox patterun support, consumer, migration to net6


omelianlevkovych:
It seems I should do rebase or squash before even creating the PR so the commit history won't be so messy.

samples/WebAPI_Marten/GreetingsEntities/Person.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System.Collections.Generic;
+
+namespace GreetingsEntities
+{
+    public class Person
+    {
+        public int Id { get; set; }
+        public ICollection<Greeting> Greetings { get; set; }
```
omelianlevkovych, line 8: In case I would like to use private setters here so I can follow some DDD/OOP practices; do I have to use reflection somehow so Marten can work with serialization?

holytshirt:
> It seems I should do rebase or squash before even creating the PR so the commit history won't be so messy.

Don't work you can always clean up your commits later `git rebase master -i` and you can  squish them.
We'll most likely squish the branch when we merge it into master so don't worry

samples/WebAPI_Marten/GreetingsEntities/Person.cs
omelianlevkovych: Thanks, this one helped https://martendb.io/v3/documentation/documents/json/newtonsoft

iancooper:
We are going to park this one; but please re-open if you are still working on it. You might need to target V10 at this point.

Issue #2077, opened 2022-04-18
MySql Outbox - Issue with Connection Pooling


honkuan86:
### Describe the bug
When maxpoolsize property is set in MySql connection string, outbox sweeper will hit exception.
Tested with MySql Server 5.7.

### To Reproduce
Run any sample with MySql Outbox.

### Exceptions (if any)
MySqlConnector.MySqlException (0x80004005): Connect Timeout expired. All pooled connections are in use.
   at MySqlConnector.MySqlConnection.CreateSessionAsync(ConnectionPool pool, Int32 startTickCount, Nullable`1 ioBehavior, CancellationToken cancellationToken) in /_/src/MySqlConnector/MySqlConnection.cs:line 940
   at MySqlConnector.MySqlConnection.OpenAsync(Nullable`1 ioBehavior, CancellationToken cancellationToken) in /_/src/MySqlConnector/MySqlConnection.cs:line 445
   at MySqlConnector.MySqlConnection.Open() in /_/src/MySqlConnector/MySqlConnection.cs:line 369
   at Paramore.Brighter.Outbox.MySql.MySqlOutboxSync.OutstandingMessages(Double millSecondsSinceSent, Int32 pageSize, Int32 pageNumber, Dictionary`2 args) in /_/src/Paramore.Brighter.Outbox.MySql/MySqlOutboxSync.cs:line 351
   at Paramore.Brighter.OutboxSweeper.Sweep(Double milliSecondsSinceSent, IAmAnOutboxViewer`1 outbox, IAmACommandProcessor commandProcessor) in /_/src/Paramore.Brighter/OutboxSweeper.cs:line 38

Connection should be close after each operation is done.

iancooper:
We are going to close the issue for MySQL with #2106 , but either I or @preardon should review other providers. The issue is that when we finish with a connection we need to understand who owns it. If we do, we can close it, if we don't we can't and should leave it to whoever opened it to close it down.

Broadly if you look at #2106 you will see this needs a try...finally block that checks for ownership and closes if we own, and not a using statement

honkuan86:
Hi @iancooper, @holytshirt, @preardon , do you think we can release this fix by this week? because we are pending on this fix in one of our project and is releasing soon.

preardon:
@honkuan86 I should be able to get my parts out for you, and assuming @iancooper agrees I'll do a release

holytshirt:
If @iancooper is ready should be good to go

[...]

PR #2089, opened 2022-04-20
Fix MySql Connection Pooling issue

honkuan86:
This PR contains fixes related to #2077.


iancooper:
Thanks @honkuan86 I will take a look. I'll see if @preardon can give it the once over too

src/Paramore.Brighter.MessagingGateway.AWSSQS/SqsMessageProducer.cs
```diff
@@ -60,7 +60,7 @@ public bool ConfirmTopicExists(string topic = null)
                    _publication.MakeChannels);
            }
 
-           return string.IsNullOrEmpty(ChannelTopicArn);
+           return !string.IsNullOrEmpty(ChannelTopicArn);
```
holytshirt, line 63: @preardon  Can you review this please as it will effect you maybe?

iancooper: This is good, looks like a bug on the return value

src/Paramore.Brighter.Outbox.MySql/MySqlOutboxSync.cs
```diff
@@ -128,31 +137,38 @@ public void Add(Message message, int outBoxTimeout = -1, IAmABoxTransactionConne
             if (transactionConnectionProvider != null && transactionConnectionProvider is IMySqlTransactionConnectionProvider provider)
                 connectionProvider = provider;
 
-            var connection = connectionProvider.GetConnection();
-
-
-            if (connection.State != ConnectionState.Open) await connection.OpenAsync(cancellationToken).ConfigureAwait(ContinueOnCapturedContext);
-            using (var command = connection.CreateCommand())
+            using (var connection = connectionProvider.GetConnection())
```
iancooper, line 140: I think the question here (checking with @preardon ) is whether you can close this connection - because you may not own it. We ought to determine if you own the connection, and only close if it you own it

iancooper (commented):
We need to review the lifetime issues here with disposal of the connection as we may not own the connection, in which case disposal is problematic. We may need a try...finally with a check as to whether you own the connection

src/Paramore.Brighter.Outbox.MySql/MySqlOutboxSync.cs
iancooper: Right so @honkuan86 if I check this file: https://github.com/BrighterCommand/Brighter/blob/master/src/Paramore.Brighter.Outbox.MsSql/MsSqlOutbox.cs I can see that approach, use a try...finally and check for ownership has been used. It looks like a fix from @preardon that did not roll into other *SQL outbox code. That is the code that I think you need. //cc @holytshirt

preardon:
I will have a look at this on the weekend

preardon:
Firstly, sorry its taken so long to come back to you on this one.

There is some additional logic that I can see missing from what's in main, the crux of it is that if you're using a TransactionConnectionProvider you are participating in something bigger than yourself and need to be aware of things like (Is there an already ongoing Transaction & did I open this connection or does it belong to someone else.

I'll get a PR in for in over the next hour

preardon:
@honkuan86 Please look at #2106 And let me know what you think

@iancooper  & @holytshirt  if you could put an eye over it as well for sense checks it would be appreciated.

iancooper:
Closing this as replace by another PR

PR #2106, opened 2022-05-08
Cleaned MsSqlOutbox and fix Connection Management for MySql

preardon:
This should fix the issues in MySql

also cleaned up MsSql a little

### DIFF
--- a/src/Paramore.Brighter.Outbox.MsSql/MsSqlOutbox.cs
+++ b/src/Paramore.Brighter.Outbox.MsSql/MsSqlOutbox.cs
@@ -196,8 +196,7 @@ namespace Paramore.Brighter.Outbox.MsSql
                 CreatePagedDispatchedCommand(command, millisecondsDispatchedSince, pageSize, pageNumber);
 
                 if(connection.State!= ConnectionState.Open) connection.Open();
-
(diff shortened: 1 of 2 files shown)


iancooper (changes requested):
Exactly, @honkuan86 hopefully you can see the use of try...finally here to ensure that the correct action is taken, depending on who owns the connection or transaction.

@preardon Did we check postgres and sqlite, ought to be part of this PR as I suspect it is an issue you fixed in MSSQL but did not port to those.

honkuan86:
@iancooper , @preardon , thanks for pointing out the missing pieces, I think we can go ahead to merge this PR ? or do you want to wait the same fixes port to other outboxes as well @iancooper ?

preardon:
I am currently on Holidays in Australia, I don't know if I'll be able to get the Time to do the fixes in the next week @honkuan86 how urgently do you need these?

honkuan86:
@iancooper , are you ok with this PR to be merge first? then we fix the other outboxes on another PR. @preardon please enjoy your holidays :)

iancooper (approved):

iancooper:
@honkuan86 @preardon Sure, let's merge this one. I'll raise an issue so that we don't lose track of the others.

PR #2123, opened 2022-05-20
Web Dapper example

iancooper:
An example using Dapper with Brighter and Outbox


iancooper:
I'm going to merge this, and then re-open for Postrgres and SQL. The reason is that I need to switch track to tackle DynamoDb first

PR #2171, opened 2022-06-29
Outbox Transaction Fixed

preardon:
- Removed the transaction around Dispatch in MySql and MsSql
- Added the required elements for dealing with shared connections to sqlite
- Added transaction participation for add messages to postgres


iancooper:
@preardon I merged this one, I was worried we were not understanding that I felt it was good to go, so thought it was the fastest solution

Issue #2194, opened 2022-07-19
MsSql outbox Isn't operating in UTC Time


preardon:
MsSql outbox 

https://github.com/BrighterCommand/Brighter/blob/454599d09f39acfafe6059c03a1f51c24c097a70/src/Paramore.Brighter.Outbox.MsSql/MsSqlOutbox.cs#L533

https://github.com/BrighterCommand/Brighter/blob/454599d09f39acfafe6059c03a1f51c24c097a70/src/Paramore.Brighter.Outbox.MsSql/MsSqlOutbox.cs#L502

iancooper:
@preardon is this done?

preardon:
@iancooper 

@catalintoma fixed this for us

iancooper:
Thanks @catalintoma will close

PR #2196, opened 2022-07-20
#2194 [CT] Using UTC in MSSQL / MYSQL outbox, as this is the timezone…

catalintoma:
Using UTC in MSSQL / MYSQL outbox for dispatched/outgoing message query, as UTC is used when a message is stored.

### DIFF
--- a/src/Paramore.Brighter.Outbox.MsSql/MsSqlOutbox.cs
+++ b/src/Paramore.Brighter.Outbox.MsSql/MsSqlOutbox.cs
@@ -373,7 +373,7 @@ namespace Paramore.Brighter.Outbox.MsSql
             var connection = await _connectionProvider.GetConnectionAsync(cancellationToken);
            
             if(connection.State!= ConnectionState.Open) await connection.OpenAsync(cancellationToken).ConfigureAwait(ContinueOnCapturedContext);
-            using (var command = InitMarkDispatchedCommand(connection, id, dispatchedAt))
+            using (var command = InitMarkDispatchedCommand(connection, id, dispatchedAt ?? DateTime.UtcNow))
             {
                 if (_connectionProvider.HasOpenTransaction) command.Transaction = _connectionProvider.GetTransaction();
                 await command.ExecuteNonQueryAsync(cancellationToken).ConfigureAwait(ContinueOnCapturedContext);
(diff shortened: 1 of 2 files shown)


preardon (approved):

src/Paramore.Brighter.Outbox.MsSql/MsSqlOutbox.cs
```diff
@@ -499,7 +499,7 @@ public void MarkDispatched(Guid id, DateTime? dispatchedAt = null, Dictionary<st
         
         private void CreatePagedDispatchedCommand(SqlCommand command, double millisecondsDispatchedSince, int pageSize, int pageNumber)
         {
-            var pagingSqlFormat = "SELECT * FROM (SELECT ROW_NUMBER() OVER(ORDER BY Timestamp DESC) AS NUMBER, * FROM {0}) AS TBL WHERE DISPATCHED IS NOT NULL AND DISPATCHED < DATEADD(millisecond, @OutStandingSince, getdate()) AND NUMBER BETWEEN ((@PageNumber-1)*@PageSize+1) AND (@PageNumber*@PageSize) ORDER BY Timestamp DESC";
+            var pagingSqlFormat = "SELECT * FROM (SELECT ROW_NUMBER() OVER(ORDER BY Timestamp DESC) AS NUMBER, * FROM {0}) AS TBL WHERE DISPATCHED IS NOT NULL AND DISPATCHED < DATEADD(millisecond, @OutStandingSince, getutcdate()) AND NUMBER BETWEEN ((@PageNumber-1)*@PageSize+1) AND (@PageNumber*@PageSize) ORDER BY Timestamp DESC";
```
iancooper, line 502: If I look at the docs for MySQL 8.0 it says to use UTC_DATE() (see https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_utc-date). It is interesting because getdate() is not listed but curdate() as the alternative, which leads to the 'why did this ever work' question, which could be because it supports the SQL syntax??

So should we be using utc_date??

iancooper (commented):
Can you look at the MySQL question? Feel free to separate the PRs for MySQL and MSSQL if that helps.

src/Paramore.Brighter.Outbox.MsSql/MsSqlOutbox.cs
catalintoma: Thanks for sharing the docs, we should not be using **utc_date()** because it doesn't have enough precision.

MySQL workbench doesn't even recognize this functions, I have no idea why it works in our integration tests.
FYI, the same issue occurs when we are using DATEADD, whereas the official functions is DATE_ADD.

Just to be safe, I moved to the official functions: DATE_ADD and UTC_TIMESTAMP and pushed the changes.

I also noticed the implementation for MarkDispatched was wrong, as MSSQL and MySQL outbox weren't actually defaulting to UTC Now (it would just leave it NULL if that's what was passed as input).

iancooper (approved):
OK, this looks much better

PR #2199, opened 2022-07-21
Large Message Payloads

iancooper:
This PR is around support for large message payloads, using external storage such as a database or a distributed file system. To effect this, it proposes a pipeline for the marshaling of messages to and from Brighter, allowing arbitrary steps, including large payloads to be added (but also setting us up for things like CloudEvents support).

Two ADRs describe this PR:

* doc/adr/0003-use-claim-check-for-large-messages.md
* doc/adr/0004-use-an-envelope-wrapper-with-transports.md


src/Paramore.Brighter/Transforms/Attributes/ClaimCheck.cs
```diff
@@ -0,0 +1,62 @@
...
+using System;
+using Paramore.Brighter.Transforms.Transformers;
+
+namespace Paramore.Brighter.Transforms.Attributes
+{
+    /// <summary>
+    /// Configures large payloads that may not be possible on certain middleware platforms, or perform badly on others. A claim check allows us to
+    /// offload the message body to distributed storage, from where it can be retrieved later.
+    /// </summary>
+    public class ClaimCheck : WrapWithAttribute
```
preardon, line 34: Not a biggie, but I do not love this name "ClaimCheck" to me doesn't implicitly tell me what it is I would go for something closer to "OffloadMessageBodyCheck" and "RetrieveOffloadedMessageBody"

preardon: Or possibly something around Storing the body

src/Paramore.Brighter/Transforms/Storage/IAmAStorageProviderAsync.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// <param name="cancellationToken">The cancellation token</param>
+        Task<bool> HasClaimAsync(string claimCheck, CancellationToken cancellationToken);
+
+        /// <summary>
+        /// Puts luggage into the store and provides a claim check for that luggage
+        /// </summary>
+        /// <param name="stream">A stream representing the luggage to check</param>
+        /// <param name="cancellationToken">The cancellation token</param>
+        /// <returns>A claim check for the luggage stored</returns>
+        Task<string> UploadAsync(Stream stream, CancellationToken cancellationToken);
```
preardon, line 67: Not a big one, do we want to use Upload and Download or Store and Retrieve, I suppose in the distributed world we are always going to be "uploading" and there wouldn't be a reason to put it on local disk ect. so probably not a problem,

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -634,11 +633,7 @@ public async Task PostAsync<T>(T request, bool continueOnCapturedContext = false
             if (!_bus.HasAsyncOutbox())
                 throw new InvalidOperationException("No async outbox defined.");
 
-            var messageMapper = _mapperRegistry.Get<T>();
-            if (messageMapper == null)
-                throw new ArgumentOutOfRangeException($"No message mapper registered for messages of type: {typeof(T)}");
-
-            var message = messageMapper.MapToMessage(request);
+            var message = await _transformPipelineBuilder.BuildWrapPipeline<T>().WrapAsync(request);
```
preardon, line 636: Please Pass in Cancellation token

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -676,18 +671,28 @@ public async Task PostAsync<T>(T request, bool continueOnCapturedContext = false
...
             {
-                var message = messageMapper.MapToMessage(r);
+                var wrapPipeline = _transformPipelineBuilder.BuildWrapPipeline<T>(); 
+                var message = wrapPipeline.WrapAsync(r).GetAwaiter().GetResult();
                 AddTelemetryToMessage<T>(message);
                 return message;
             }).ToList();
         }
+        
+        private async Task<List<Message>> BulkMapMessagesAsync<T>(IEnumerable<T> requests) where T : class, IRequest
```
preardon, line 683: Should we be passing a cancellation token

src/Paramore.Brighter/TransformPipelineBuilder.cs
```diff
@@ -0,0 +1,262 @@
...
+                else
+                {
+                    transforms.Add(transformer);
+                }
+            });
+
+            return transforms;
+        }
+
+        public static void ClearPipelineCache()
```
preardon, line 184: This doesn't appear used, given that this is somewhat internal, I suppose being static it could be useful

src/Paramore.Brighter.Tranformers.AWS/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,66 @@
...
+namespace Paramore.Brighter.Tranformers.AWS
+{
+    public static class ServiceCollectionExtensions
+    {
+        /// <summary>
+        /// Will add an S3 Luggage Store into the Service Collection
+        /// </summary>
+        /// <param name="services"></param>
+        /// <param name="configure"></param>
+        public static void AddS3LuggageStore(this IServiceCollection services, Action<S3LuggageOptions> configure)
```
preardon, line 40: Should we be adding and retrieving these things from the ServiceCollection?

Given that we usually ensure that you can BYO DI, I feel this should be in the Brighter Builder

src/Paramore.Brighter/CommandProcessorBuilder.cs
```diff
@@ -245,7 +247,8 @@ public CommandProcessor Build()
                     outboxTimeout: _outboxWriteTimeout,
                     featureSwitchRegistry: _featureSwitchRegistry,
                     boxTransactionConnectionProvider: _overridingBoxTransactionConnectionProvider,
-                    outboxBulkChunkSize: _outboxBulkChunkSize
+                    outboxBulkChunkSize: _outboxBulkChunkSize,
+                    messageTransformerFactory: _transformerFactory
```
preardon, line 251: Does this mean that we only want this to work with External Bus without Request Reply Queues?

src/Paramore.Brighter.Tranformers.AWS/ServiceCollectionExtensions.cs
preardon: My Mistake, I see this can work with BYO DI (you just have to implement the Transform Factory for this), Something still feels wrong to me around it being off of Service Collection, but maybe it just needs to sit with me a little long :)

preardon:
I see that we are currently dealing with Message Mapping in the Command Processor, I wonder if we should look to move this into the External bus as it feels like Message are a concern of going out of Proc and that is what the External Bus Service has morphed into in my mind.

Does not need to be resolved now, just something that stuck with me through this review

src/Paramore.Brighter/Transforms/Attributes/ClaimCheck.cs
iancooper: The reason for using Claim Check is that it is the name of the pattern:  https://www.enterpriseintegrationpatterns.com/patterns/messaging/StoreInLibrary.html see usage here: https://learn.microsoft.com/en-us/azure/architecture/patterns/claim-check

src/Paramore.Brighter/TransformPipelineBuilder.cs
iancooper: It is for tests, though it has not been needed at this point.

src/Paramore.Brighter.Tranformers.AWS/ServiceCollectionExtensions.cs
iancooper: Well this is really an optimization for folks using ServiceCollection, which we have elsewhere. In essence folks on other IoC containers need to implement an equivalent

src/Paramore.Brighter/CommandProcessorBuilder.cs
iancooper: Not the intent.

iancooper:
> I see that we are currently dealing with Message Mapping in the Command Processor, I wonder if we should look to move this into the External bus as it feels like Message are a concern of going out of Proc and that is what the External Bus Service has morphed into in my mind.
> 
> Does not need to be resolved now, just something that stuck with me through this review

Yeah, you are right. I just updated in place over moving with this work. Let me see if that would be too much of a change

Issue #2210, opened 2022-08-03
Using DI to create handlers does not work in v9+


wozzo:
### Describe the bug
We had v8.1.1399 installed.
Just tried both 9.0.1 and 9.1.16, with the same results on each.
Whenever the `commandProcessor.Send(command);` method is called we get an exception with the message "No handler factory defined.".
I've checked and the subscriber registry does contain an appropriate mapping for the command to handler.
No code changes between the attempts, only the package version.
I've been unable to find a change log, or release notes that describe any breaking changes, and a review of the documentation leads me to believe we're doing the right thing still.

### To Reproduce
[Broken version](https://github.com/wozzo/BrighterIssue)
With this version you'll get an exception as described below thrown on the `Send` method call.
If you change the version of Brighter to 8.1.1399 then you'll get a `NotImplementedException` in the handler factory but this is because I've not implemented the handler factory here as it is not needed to demonstrate the bug. If the code gets to this point then it has found the handler type.

### Exceptions (if any)
System.InvalidOperationException
  HResult=0x80131509
  Message=No handler factory defined.
  Source=Paramore.Brighter
  StackTrace:
   at Paramore.Brighter.CommandProcessor.Send[T](T command)
   at Program.<>c.<<Main>$>b__0_0(IAmACommandProcessor commandProcessor) in C:\Projects\BrighterTest\WebApplication1\Program.cs:line 21
   at Microsoft.AspNetCore.Http.RequestDelegateFactory.<>c__DisplayClass36_0.<Create>b__0(HttpContext httpContext)
   at Microsoft.AspNetCore.Routing.EndpointMiddleware.Invoke(HttpContext httpContext)

### Further technical details
- Brighter version 9.0.1+
- Include the output of `dotnet --info`
.NET SDK (reflecting any global.json):
 Version:   6.0.302
 Commit:    c857713418

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22000
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\6.0.302\

global.json file:
  Not found

Host:
  Version:      6.0.7
  Architecture: x64
  Commit:       0ec02c8c96

.NET SDKs installed:
  6.0.300 [C:\Program Files\dotnet\sdk]
  6.0.301 [C:\Program Files\dotnet\sdk]
  6.0.302 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
[...]

PR #2212, opened 2022-08-06
fix(#2210) | Adds check for correct handler factory

DevJonny:
Fix for issue #2210.

Changes made:

1. Change language version to C# 9 on the `Paramore.Brighter.csproj`. I can't remember where we were at with language versions. If we're not happy to move to 9 I can change my code to work with 8, it just won't be as succint.

2. Adds a check in the `CommandProcessor` for making sure that the `handlerFactory` is either a `IAmAHandlerFactorySync` or a `IAmAHandlerFactoryAsync` and throws an exception with a sensible message if not. This is so when moving from Brighter 8.x -> 9.x you don't get tripped up by using `IAmAHandlerFactory`. See Issue #2210 for details.

3. Removed now redundant tests that checked for errors when a null handler factory is passed


iancooper (commented):
So this is a good general fix - we definitely need to give a lot more feedback in configuration that lets you explicitly know what you misconfigured.

Up to now though we have not moved to C# 9 because it requires runtime support that is not in netstandard 2.0. Some language elements - those that don't need the language runtime work - but others don't

iancooper:
@preardon @holytshirt Any thoughts on C#8 vs C#9. I think the language runtime support requirements make this a little bit hairy. Some C#9 features will work, other's won't. I'm inclined to stick with C#8 until we feel netstandard 2.0 is no longer required, but that feels like v10 not v9.X, if then? But think this is a majority call.

preardon:
It's an interesting one for me, as we are requiring latest versions of all downstream packages.

Personally I like to maintain new lang versions, but maintaing framework compatibility is important given lots of people are still on it

iancooper:
@preardon We should not upgrade to packages that don't support .netstandard2.0 without multi-targeting. That advice based on discussions we had with the DotNet team.

preardon:
@iancooper completely agree

Was just worried that Lang ver upgrades might break framework support as well

iancooper:
@preardon Yeah, that is the issue. To use C#9, we can only use a subset of the features--those that don't require runtime support. So we could use some parts, but not others. My concern is that this is harder to get right than a simple "don't"

In essence we can use 7.3 and 8 and above only with run-time support, see: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version

I believe that Mads says, use of nullable types is tested to work in 7.3 and recommended but not others.

BTW for samples, just use .NET 6 and latest language. The issue is just the libs

iancooper:
BTW, for reference, if we leave netstandard2.0 for netstandard2.1 (we have had to for some libs where there is no maintained 2.0 support) we lose .NET framework 4.8 support and only have .NET Core 3.1 and above. It's difficult to know how much of our customer base uses .NET Framework 4.8. I think if we wanted to drop support it would be a major version change.

DevJonny:
Thanks for the confirmation @iancooper / @preardon, I couldn't remember where we landed. 

I've reverted back to C# 8, and updated the check accordingly. Should be all good now.

DevJonny:
Urggh I need to fix a bunch of tests now as they're passing in `IAmAHandlerFactory` instead of `IAmAHandlerFactory(A)sync`. I'll fix up.

holytshirt:
> Urggh I need to fix a bunch of tests now as they're passing in `IAmAHandlerFactory` instead of `IAmAHandlerFactory(A)sync`. I'll fix up.

You mean tdd works ;)

DevJonny:
Good news is a few tests needed removing (because they're checking the same thing, but at the point where the handler is created which is too late), I've still got a couple of failing tests I need to sort, but I'll do that tonight.

DevJonny:
Build failed because it took over 30 minutes to run :(

holytshirt (approved):

PR #2255, opened 2022-09-04
feat(DI): Use services.Try... as recommend for libraries

holytshirt:
- This means it is only added if it doesn't exist so first registration wins. Otherwise it's last registration wins
- https://docs.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#service-registration-methods


iancooper (approved):
Useful, as it does allow folks to override our defaults, as well as picking up the double-registered issue

preardon (approved):
Looks good to me

Issue #2251, opened 2022-08-31
[Feature] Support Bulk Deposit into the outbox


preardon:
I have a requirement to dispatch a batch of commands, i.e. for Pay Interest, all accounts are batched into chunks and then I sent thousands of PayInterestToAccountsCommand 

**Describe the solution you'd like**
On the command processor I would like DepositPost to take an array of message and then persist them to the outbox in a single (or smaller number of) statements

**Describe alternatives you've considered**
At the minute I've written things to go around Brighter as going through brighter was far too much of a performance hit

preardon:
@iancooper let me know if you're ok for me to go ahead with this,

Also would you like it to be DepositPost or would you like it to stick out i.e. BulkDepositPost, The return type would have to change to an array of persisted messages as well

iancooper:
@preardon  Technically, if we change the interface in a breaking way, it would be a bump to v10. So I think that would mean we need a BulkDeposit in v9. We could choose to collapse them together in V10. I guess there might be some advantages in keeping them separate, bbulk ecause you are going to need to create bulk insert statements for a lot of storage tools, so it would then be possible to identify with each release which outboxes now support a bulk write.

iancooper:
PS It is still a minor version change, to alter the API in a backward compatible way, not a patch, of course.

preardon:
@iancooper not suggesting a breaking change, just an overload, happy to do it either way.

I was thinking that we may want a new Interface for IAmABulkOutboxAsync

preardon:
This has been delivered is 9.2.3

PR #2264, opened 2022-09-11
Intial Plumb in of Bulk Deposit Async

preardon:
Added IAmABulkOutboxAysnc
Plumbed in a Bulk Deposit Post
Implimented IAmABulkOutboxAsync for MsSql and InMemory

#2251 

@iancooper  @holytshirt  @DevJonny Let me know what you think about this approach


src/Paramore.Brighter.Outbox.Sqlite/SqliteOutboxSync.cs
```diff
@@ -174,6 +174,104 @@ public void Add(Message message, int outBoxTimeout = -1, IAmABoxTransactionConne
...
+                {
+                    if(!connectionProvider.IsSharedConnection) connection.Dispose();
+                    else if (!connectionProvider.HasOpenTransaction) connection.Close();
+                }
+            }
+        }
+
+        /// <summary>
+        /// Interface IAmABulkOutboxAsync
+        /// In order to provide reliability for messages sent over a <a href="http://parlab.eecs.berkeley.edu/wiki/_media/patterns/taskqueue.pdf">Task Queue</a> we
```
iancooper, line 227: We can probably drop the task queue from comments and replace with external bus. The task queue wording dates from early periods of Brighter when that was the major use case over microservice communication

iancooper (approved):
looks good, the comment is just around a stale header comment ( which may not be the only place) and is pedantic rather than blocking

iancooper (approved):

iancooper:
@preardon Where are we with this one?

preardon:
@iancooper if you're happy with it I'm happy to merge

iancooper:
@preardon Go for it

PR #2270, opened 2022-09-16
Update Names

iancooper:
We moved to Subscriptions from Connections to clarify naming, which was confusing alongside how we connect to a broker. We missed the change to the fluent interface. This phases that change in. We do change the name of a public interface, but it is not used outside of our code, so I believe this risk is acceptable.


preardon (approved):
Looks good to me

Issue #301, opened 2018-08-16
Support OpenTelemetry


iancooper:
It needs some design thinking in this thread, but we should decide if it makes sense to support OpenTracing

http://opentracing.io/documentation/pages/instrumentation/instrumenting-frameworks

iancooper:
We could probably stab at an initial version using Jaeger

https://github.com/yurishkuro/opentracing-tutorial

iancooper:
Oh and as it may not be obvious. We would look to implement this as an attribute, similar to the existing Monitoring attribute which you add to a handler to trace handling individual messages. We could choose to instrument parts of the infrastructure too, but it is harder to tie that to tracing the path of a request through a distributed system, over integrating with say Prometheus to provide operations metrics, which may be better. 

We might want to use the Control Bus to push metrics that Prometheus could monitor from a a server collating control bus data and pushing to Prometheus, but not the scope of this work.

holytshirt:
Now call  Open Telemerty  https://opentelemetry.io/

iancooper:
Moved this to OpenTelemetry, which shows how long it has been open :-D

Driedas:
I am interested in this, have taken a look at the source code for an attempt to implement this but I am not sure that the extensibility points needed to do so are in place.
First, how should this actually behave (I think): `opentelemetry-dotnet` is based on `System.Diagnostics.Activity` so there is no need to take a dependency on that library. I believe that on the sending/publishing end, the current activity id should be appended to all outgoing message headers and on the receiving end an activity should automatically get created, it's parent id set to the id from the message header and started (plus adding a couple of useful activity tags like message id and type, topic...). This activity should be then stopped as the very last step in the pipeline, after all the user defined pipeline steps have been completed. This I think is similar to how the HttpClient works, and [Jimmy Bogard's similar extension to NServiceBus](https://github.com/jbogard/NServiceBus.Extensions.Diagnostics).
Now, why I think that the necessary extensibility points are missing. The execution pipeline in Brighter is configured using a series of handlers that act as middleware. This is however already too late at the receiving end as at this stage you no longer have access to the raw message to be able to get the activity id out of the headers, and I don't think you would want to pollute the actual request objects with an ActivityId. The only places where you have the message headers in the raw form (as far as I can see) is in the message mapper and the message pump. This logic does not belong to the message mapper, and the message pump does not expose an extension point around the `DispatchRequest` method.
[...]

PR #2285, opened 2022-09-22
Add initial OpenTelemetry (Tracing) Support

preardon:
A first look at open telemetry inside of brighter

relates to issue: #301


holytshirt:
@Driedas you may be interested in this work

Driedas:
Hey Toby, thanks for the ping, much appreciated. Happy to see this implemented, it looks quite similar to what I had in mind :-)

src/Paramore.Brighter.ServiceActivator/MessagePump.cs
```diff
@@ -119,11 +123,26 @@ public void Run()
                 }
 
                 // Serviceable message
+                Activity span = null;
                 try
                 {
+                    message.UpdateCloudEventsFromHeaders();//ToDo: Discuss this as a temp measure
```
iancooper, line 129: I would suspect we want to establish headers that are populated as a result of translating a message i.e. to message headers from the transport, and then test for the presence of those headers i.e. MessageHeaders has some specific properties, and then we can figure out how to populate MessageHeaders from the transport, whether cloud events are present or not. binary vs structured etc

src/Paramore.Brighter.ServiceActivator/MessagePump.cs
```diff
@@ -132,17 +151,21 @@ public void Run()
                         Channel.Name, Thread.CurrentThread.ManagedThreadId);
 
                     RejectMessage(message);
+                    span?.SetStatus(ActivityStatusCode.Error,
```
iancooper, line 154: I wonder if this would be part of a span? Is the health of this serviceactivator part of a trace?

src/Paramore.Brighter.ServiceActivator/MessagePump.cs
```diff
@@ -132,17 +151,21 @@ public void Run()
...
                 catch (DeferMessageAction)
                 {
+                    span?.SetStatus(ActivityStatusCode.Error, "Deferring message for later action");
                     if (RequeueMessage(message)) continue;
                 }
                 catch (AggregateException aggregateException)
                 {
                     var (stop, requeue) = HandleProcessingException(aggregateException);
 
+                    span?.SetStatus(ActivityStatusCode.Error, $"Error while dispatching, Stopping {stop}, re-queueing {requeue}");
```
iancooper, line 168: Again, not sure what our action on stop vs requeue should be. We probably want a requeue, but unsure about stop. Do we just kill the span on a stop i.e. we add nothing because something else will re-process you?

src/Paramore.Brighter/Command.cs
```diff
@@ -45,5 +46,10 @@ public Command(Guid id)
         {
             Id = id;
         }
+        
+        /// <summary>
+        /// Gets or sets the span that this operation live within
+        /// </summary>
+        public Activity Span { get; set; }
```
iancooper, line 53: Should this be on Request?

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -471,6 +543,8 @@ public async Task PostAsync<T>(T request, bool continueOnCapturedContext = false
 
             var message = messageMapper.MapToMessage(request);
 
+            AddCloudEventsToMessage<T>(message);
```
iancooper, line 546: Again, I think add to the Message Header

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -508,11 +582,23 @@ public async Task PostAsync<T>(T request, bool continueOnCapturedContext = false
...
```
iancooper, line 585: As above

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -508,11 +582,23 @@ public async Task PostAsync<T>(T request, bool continueOnCapturedContext = false
...
+
             await _bus.AddToOutboxAsync(request, continueOnCapturedContext, cancellationToken, message, connectionProvider);
 
             return message.Id;
         }
 
+        private void AddCloudEventsToMessage<T>(Message message)
```
iancooper, line 592: See earlier comments. I don't think we know in the CP whether you are using CE or not. That lives with the transport (and possibly enveloper wrapper later). All this should know is that MessageHeader has some appropriate properties

src/Paramore.Brighter/IRequest.cs
```diff
@@ -37,5 +38,10 @@ public interface IRequest
         /// </summary>
         /// <value>The identifier.</value>
         Guid Id { get; set; }
+
```
iancooper, line 41: OK, so it is here, and you can ignore the Command comment above

src/Paramore.Brighter/Message.cs
```diff
@@ -48,6 +55,11 @@ public class Message : IEquatable<Message>
         /// <value>The body.</value>
         public MessageBody Body { get; private set; }
 
+        /// <summary>
+        /// Gets the details for the Cloud Events specification
+        /// </summary>
+        public MessageCloudEvents CloudEvents { get; private set; }
```
iancooper, line 61: I think these are just part of MessageHeader, and CE is 'transport specific')

src/Paramore.Brighter/Message.cs
```diff
@@ -76,6 +88,42 @@ public Message(MessageHeader header, MessageBody body)
...
             Body = body;
+            
+            UpdateCloudEventsFromHeaders();
+        }
+
+        /// <summary>
+        /// Populate the Cloud Events from the HeaderBag
+        /// </summary>
+        [Obsolete("Looking to remove this in v10 in favour of storing the Cloud Events in their own property in the outbox")]
+        public void UpdateCloudEventsFromHeaders()
```
iancooper, line 99: Same theme, there are just header properties for tracing, we can do CE support later

src/Paramore.Brighter/MessageCloudEvents.cs
```diff
@@ -0,0 +1,64 @@
+﻿using System;
+
+namespace Paramore.Brighter
+{
+    public class MessageCloudEvents
```
iancooper, line 5: Just create new MessageHeader telemetry properties

iancooper:
Main feedback: let's ignore CloudEvents here. CE needs to get figured out within the transport. We can probably give better support when I build the Wrap-Unwrap pipeline. We should just treat the Brighter internal representation as properties we need for telemetry on MessageHeader. The transport can populate them there, regardless of how they were supplied. That makes your core code here independent of what standard we use, or how we let folks choose that.

src/Paramore.Brighter/MessageCloudEvents.cs
preardon: I'll move this into headers and rename it to Telemetry, this got created early on and got the naming from the distributed Tracing stuff

src/Paramore.Brighter.ServiceActivator/MessagePump.cs
preardon: This won't necessarily "Kill" the whole span, the activity is created as a child of the activity in the producer. I will do a little more testing here to ensure.

At the minute I'm likening Service activator the asp.net as it's effectively a framework around requests, if something unexpected goes wrong in returns 500, if the upstream has a retry policy and the next attempt works then the parent will still return the operation as successful but when you look in the traces it'll have a failed attempt.

I'll setup a sample to fail and report back

src/Paramore.Brighter.ServiceActivator/MessagePump.cs
preardon: Did we want to add these to the transport in 9.x as I'm happy to do that? 

I was working on the assumption of have them in the header Bag for v9.x then move them out with v10

The reason this is here is due to how the rmq transport reconstructed the message Headers and then was meant to be a shim until we went full implementation

src/Paramore.Brighter/CommandProcessor.cs
preardon: I think this is terribly named on my part. I think it's should be called something closer to AddTelemetryToMessage or EnsureMessageTelemetry

src/Paramore.Brighter/Message.cs
preardon: Renamed and moved

src/Paramore.Brighter/Message.cs
preardon: Renamed

src/Paramore.Brighter/MessageCloudEvents.cs
preardon: Renamed and Moved

iancooper: Yeah, this looks much better

src/Paramore.Brighter/Message.cs
iancooper: Clearer

src/Paramore.Brighter/CommandProcessor.cs
iancooper: Seen resolution

src/Paramore.Brighter.ServiceActivator/MessagePump.cs
iancooper: Thanks

src/Paramore.Brighter.ServiceActivator/MessagePump.cs
iancooper: As with the other comment. let's see what this looks like. I guess it is conceptual, is a trace intended to 'monitor' health, or just to show 'flow' and events in the flow. Is a 'stop' event, just in the flow, perhaps, but not sure that 'shutdown' subsequent is for that trace.

preardon: As discussed, this just shows as a failure of processing of that one part of the trace

src/Paramore.Brighter.ServiceActivator/MessagePump.cs
preardon: As discussed, this just shows as a failure of processing of that one part of the trace
image

iancooper (approved):

Issue #2269, opened 2022-09-15
Kafka Consumer Offset not committed, only committed after service restarts


honkuan86:
### Discussed in https://github.com/BrighterCommand/Brighter/discussions/2263

<div type='discussions-op-text'>

<sup>Originally posted by **honkuan86** September  9, 2022</sup>
Hi all,

We have noticed an issue when consuming message from Kafka using Brighter, where the offsets are not committed until a service restarts.

I'm using the settings bellow:
`new KafkaSubscription<UserSessionEvent>(
                    new SubscriptionName("user.session.subscription"),
                    channelName: new ChannelName("user.session.channel"),
                    routingKey: new RoutingKey(messageBusConfig.SubscriptionTopics.UserSessionEvents),
                    groupId: "task-service.group",
                    timeoutInMilliseconds: 100,
                    offsetDefault: AutoOffsetReset.Earliest,
                    isAsync: true,
                    commitBatchSize:15,
                    noOfPerformers: 3,
                    sweepUncommittedOffsetsIntervalMs: 5000)`

Note: The service is hosted in AWS EKS and consuming from a AWS MSK Cluster

The image below shows the logs after the service restarts, from the image it shows that all uncommitted offsets are committed at the same time, and after awhile it will starts slowing down, until a second restart, then it will commit all uncommitted offsets again.

![image](https://user-images.githubusercontent.com/45580206/189310972-96a6b7f9-e6c5-4b29-b22d-3853d436199b.png)
</div>

honkuan86:
Hi @iancooper , may I know if you know about this behavior ? Or did I doing the configuration wrong ?

iancooper:
Hi @honkuan86. That is a little disturbing.
:
The offsets should be committed when:

- You hit the batch size
- An timeout triggers a "no offsets committed, let's commit what we have"
- Partition changes
- Shutdown

So, a failure to commit them prior to that is worrying.

You could try lowering the batch size, to see what happens i.e. lower it to about 5. (Lowering to 1 maybe counter productive).

A couple of things to try (individually) to see their impact:

- Batch Size: Lower this to 5 for a while, see if a reduced batch size triggers the commit earlier
- NoOfPerformers: Drop this to 1. I need to think about what increasing the noofperformers means for Brighter with Kafka; we would have to allocate each performer a different partitions, and have the part the same consumer group

BTW
[...]

PR #2294, opened 2022-09-29
Kafka threading issue meant offsets not being committed.

iancooper:
Failing to commit offsets #2269 

The KakfaConsumer was using a Monitor to protect us against concurrent attempts to flush. The strategy was:

- Enter the monitor
- Run the flush in a new thread, and exit the monitor when the flush completes

However, a monitor does not work across threads, so this was failing. For a cross-thread lock we need to use a Semaphore (in this case we can use the lightweight SemaphoreSlim).

In addition, when we did a sweep we were sweeping up to batch size of items. We have changed that, and now sweep the current length of offset storage (obviously this will only be up to the point of the sweep, new items will be arriving).

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -7,6 +7,9 @@ This section lists features in master, available by [AppVeyor](https://ci.appvey
 
 ## Master ##
 
+- Bug with Kafka Consumer failing to commit offsets fixed. Caused by Monitor being used for a lock on one thread and released on another, which does not work. Replaced with SemaphoreSlim.
+- Behavior of Kafka Consumer offset sweep changed. It now runs every x seconds, and not every x seconds since a flush. This will cause it to run more frequently, but it is easier to reason about.
+
 ## Release 9.1.14 ##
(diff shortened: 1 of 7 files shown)


iancooper:
PS Sweeper now runs at interval from last sweep. We ignore the last run of a batch commit. This is easier to reason about, every _x_ seconds, you flush anything not flushed yet.

preardon (approved):
Looks good to me,

However did you want to put an else on your _flushTokenWait(0) with some logging so that you can tell that it's been skipped due to one already being in progress (apologies if there already is I did this on a phone)

iancooper:
> Looks good to me,
> 
> However did you want to put an else on your _flushTokenWait(0) with some logging so that you can tell that it's been skipped due to one already being in progress (apologies if there already is I did this on a phone)

Not about shout. I expect I'll make it a debug log, as most of the time folks won't want to see it, but it could be handy for trouble shooting

PR #2332, opened 2022-10-19
Pass through the args for outbox to support Dynamo and Event Store

iancooper:
We don't pass through the args required by Dynamo or Event Store to manage an Outbox. This adds the option to pass these through. Should be non-breaking as it is an optional argument (unless you are implementing your own CommandProcessor).

This makes non-breaking changes to interfaces:

* CommandProcessor -> The Outbox Sweeper methods (they need a rename still) now take the args dictionary needed to pass through to these outboxes
* Producer -> The Producer now has an OutboxBag for these arguments
* Publication -> The Publication now has an OutboxBag, which is passed to the Producer on creation.

We may want to review how this is done in V10, when we can make breaking changes to these interfaces.

### DIFF
--- a/src/Paramore.Brighter.Outbox.DynamoDB/DynamoDbOutbox.cs
+++ b/src/Paramore.Brighter.Outbox.DynamoDB/DynamoDbOutbox.cs
@@ -267,12 +267,14 @@ namespace Paramore.Brighter.Outbox.DynamoDB
          int pageNumber = 1, 
          Dictionary<string, object> args = null)
         {
+            var now = DateTime.UtcNow;
+            
             if (args == null)
             {
                 throw new ArgumentException("Missing required argument", nameof(args));
             }
-            
-            var createdTime = DateTime.UtcNow.Subtract(TimeSpan.FromMilliseconds(millisecondsDispatchedSince));
+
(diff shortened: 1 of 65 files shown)


iancooper:
Issues with a failing test around outstanding messages fixed by correcting the type in the filter expression

src/Paramore.Brighter.Outbox.DynamoDB/DynamoDbOutbox.cs
```diff
@@ -266,12 +272,14 @@ public void MarkDispatched(Guid id, DateTime? dispatchedAt = null, Dictionary<st
...
+            var now = DateTime.UtcNow;
+            
             if (args == null)
             {
                 throw new ArgumentException("Missing required argument", nameof(args));
             }
-            
-            var sinceTime = DateTime.UtcNow.Subtract(TimeSpan.FromMilliseconds(millisecondsDispatchedSince));
+
+            var dispatchedTime = now.Subtract(TimeSpan.FromMilliseconds(millisecondsDispatchedSince));
```
preardon, line 282: This should be Inserted/Created time, as if it has a dispatched time it should not be returned in this call as it is not longer outstanding

preardon:
Looks good

preardon (approved):

PR #2342, opened 2022-10-25
Dynamo Outbox constructor fixes

Thijmen:
This PR makes sure that an instance of DynamoDbConfiguration can be created without an instance of AWSCredentials and RegionEndpoint. These arguments are being set as properties inside the class - however, upon review not being used. I marked the properties as obsolete as the Outbox is using the connection from the IAmazonDynamoDB.


iancooper (approved):
Thanks for this, we'll get a point release out

PR #2348, opened 2022-11-04
Use Overwrite table name for outbox for DynamoDB.

Thijmen:
This PR makes sure that in all instances the overwrite table name is being used, if provided.


iancooper (approved):
Thanks for that

PR #2358, opened 2022-11-08
Downgrade Sqlite

preardon:
Microsoft.Data.Sqlite downgraded from 7.0.0 to  6.0.11

@iancooper  This fixes the Build


iancooper:
I believe that @holytshirt has an alternate solution

PR #2396, opened 2022-12-02
Set correct partition key (kafka key) for Kafka messages 

Thijmen:


iancooper (approved):
Discussed with @Thijmen

PR #2397, opened 2022-12-03
Set correct span status for Send and SendAsync

easyfy-fredrik:
### OpenTelemetry instrumentation
Span status (Error/OK) gets flipped for Send() and SendAsync() while being correct for Publish() and PublishAsync(). 
EndSpan() has a success parameter which sets the span status to Error if true which is a bit confusing. 

This change sets Error status directly on exception and assumes success if no other status is set, this should fix the flipped status and remove the need for the success parameter.

### DIFF
--- a/src/Paramore.Brighter/CommandProcessor.cs
+++ b/src/Paramore.Brighter/CommandProcessor.cs
@@ -280,7 +280,6 @@ namespace Paramore.Brighter
 
             using (var builder = new PipelineBuilder<T>(_subscriberRegistry, _handlerFactorySync, _inboxConfiguration))
             {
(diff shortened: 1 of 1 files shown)


iancooper:
Thanks @easyfy-fredrik I am going to pull in @preardon to review, as he wrote our original telemetry support

iancooper (approved):

iancooper:
@easyfy-fredrik We have a couple of other fixes in the pipe, if they are there in the next 24 hours I will fold into a release, otherwise I will get this out then

iancooper:
PS Thanks for the contribution, I'll add you to our list

PR #2406, opened 2022-12-11
Start to refactor the MsSql Outbox to be a little cleaner

preardon:
The MSSQL Outbox has had a little bit of spaghetti code and quite a lot of duplications, the Aim of this PR is to clean the code and ensure that we are not duplicating code and that the same queries and methods are being used for complimentary paths

### DIFF
--- a/src/Paramore.Brighter.Outbox.MsSql/MsSqlOutbox.cs
+++ b/src/Paramore.Brighter.Outbox.MsSql/MsSqlOutbox.cs
@@ -535,7 +535,7 @@ namespace Paramore.Brighter.Outbox.MsSql
         {
             var command = connection.CreateCommand();
             command.CommandText = $"SELECT * FROM {_configuration.OutBoxTableName} WHERE MessageId = @MessageId";
-            command.CommandTimeout = outBoxTimeout;
+            command.CommandTimeout = outBoxTimeout <0 ? 0 : outBoxTimeout;
             command.Parameters.Add(CreateSqlParameter("MessageId", messageId));
             return command;
         }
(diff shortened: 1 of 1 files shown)


preardon:
Managed to reduce the lines of code but a little over 200, also we now have more structure and less duplicated code.

Just a thought, we could probably Abstract class a decent portion of this and get it to service at least MySQL and Progress as well

let me know what you think

src/Paramore.Brighter.Outbox.MsSql/MsSqlOutbox.cs
```diff
@@ -555,177 +295,207 @@ public void MarkDispatched(Guid id, DateTime? dispatchedAt = null, Dictionary<st
...
-            var sql = string.Format(pagingSqlFormat, _configuration.OutBoxTableName);
 
-            command.CommandText = sql;
-            command.Parameters.AddRange(parameters);
-        }
+        #endregion
 
-        private void CreatePagedReadCommand(SqlCommand command, int pageSize, int pageNumber)
+        #region Things that Actually call SQL
+        private void RunNonQuery(IAmABoxTransactionConnectionProvider transactionConnectionProvider, Func<SqlConnection, SqlCommand> commandFunc, Action loggingAction)
```
iancooper, line 311: Isn't NonQuery a Command. This seems like a difficult negative naming choice

iancooper (commented):
Looks a LOT better. I am slightly weirded by us surfacing NonQuery, it feels like we need a better name. Write vs. Read if no other option.

src/Paramore.Brighter.Outbox.MsSql/MsSqlOutbox.cs
preardon: Good spot, I can get that renamed

preardon: I've renamed them ReadFromStore and WriteToStore Thought @ICooper ?

iancooper (approved):

PR #2419, opened 2022-12-17
feat: Adds FeatureSwitchAsync attribute and handler

DevJonny:
When I wrote the original `FeatureSwitch[Attribute|Handler]` I didn't add an async version for using with handlers that extend `RequestHandlerAsync`, this PR adds that functionality.


iancooper (approved):
Looks good

PR #2430, opened 2022-12-20
Use Kafka with Schema Registry

iancooper:
Whilst use of MessageMapper should allow us to use Kafka with the Schema Registry, there are a couple of issues:

- Confluent's DotNet client uses Json.NET and NJsonSchema
- Ignoring a property, requires Json.NET to avoid serialize/deserialize
- Ignoring a property requires NJsonSchema to avoid adding it to the schema against which we validate
- The OTel property Activity is one we want to ignore in JSON serialization but we only do this for System.Text.Json

This change allows us to support serializing using Json.NET and using NJsonSchema

### DIFF
--- a/src/Paramore.Brighter/Command.cs
+++ b/src/Paramore.Brighter/Command.cs
@@ -44,7 +44,7 @@ namespace Paramore.Brighter
         /// Initializes a new instance of the <see cref="Command"/> class.
         /// </summary>
         /// <param name="id">The identifier.</param>
-        protected Command(Guid id)
+        public Command(Guid id)
         {
             Id = id;
         }
--- a/src/Paramore.Brighter/Event.cs
+++ b/src/Paramore.Brighter/Event.cs
@@ -46,7 +46,7 @@ namespace Paramore.Brighter
         /// Initializes a new instance of the <see cref="Event"/> class.
         /// </summary>
         /// <param name="id">The identifier.</param>
-        protected Event(Guid id)
+        public Event(Guid id)
         {
             Id = id;
         }


iancooper:
So this PR fixes the issues associated with using the schema registry. In essence, we just need to add attributes to the Command and Event classes to support the use of Json.NET and NJsonSchema by Confluent, as opposed to System.Text.Json

https://github.com/BrighterCommand/Brighter/pull/2430

You can see a message mapper here: https://github.com/BrighterCommand/Brighter/blob/98d0328aeda02c270778d1dac715225ad[…]aRegistry/Greetings/Ports/Mappers/GreetingEventMessageMapper.cs

Note how we use the JsonSerializer from Confluent. We provide some defaults as a helper when creating your config and settings that should work to produce a good schema. You don't have to use them, but it is a "black art" to configure NJsonSchema.

Note also that the MessageMapper takes a dependency on ISchemaRegistryClient.

You need to register your schema registry with the IoC container to fulfill this dependency:

https://github.com/BrighterCommand/Brighter/blob/98d0328aeda02c270778d1dac715225ad6618be5/samples/KafkaSchemaRegistry/GreetingsSender/Program.cs

var schemaRegistryConfig = new SchemaRegistryConfig { Url = "http://localhost:8081"};
var cachedSchemaRegistryClient = new CachedSchemaRegistryClient(schemaRegistryConfig);
                    services.AddSingleton<ISchemaRegistryClient>(cachedSchemaRegistryClient);

      
To test this we added confluent schema registry and control centre to our kafka docker compose file.

PR #2447, opened 2022-12-29
Schema Registry Support

iancooper:
To make it possible to use the Confluent Schema Registry with serialization from System.Text.Json, this transform will:

1: Register the schema of an outgoing schema with the registry (initial implementation provided for the Confluent Schema Registry)
2: Validate the schema of the outgoing message against the schema stored in the registry.

As we practice a 'tolerant reader' strategy there is no corresponding unwrap to validate against the schema on receipt.

For performance/availability implementations of the schema registry will need to cache entries locally once read/written.

We need to be able to version schemas in the registry.


iancooper:
I am going to close this PR as we would want to start again here.

PR #2468, opened 2023-01-08
Outbox archiver

preardon:
The idea behind this is to have a service that can move old messages from the outbox to another provider

### DIFF
--- a/src/Paramore.Brighter.Archive.Azure/AzureBlobArchiveProvider.cs
+++ b/src/Paramore.Brighter.Archive.Azure/AzureBlobArchiveProvider.cs
@@ -1,0 +1,39 @@
+using Azure.Storage.Blobs;
+
+namespace Paramore.Brighter.Storage.Azure;
+
+public class AzureBlobArchiveProvider : IAmAnArchiveProvider
+{
+    private BlobContainerClient _containerClient;
+    
+    public AzureBlobArchiveProvider(AzureBlobArchiveProviderOptions options)
+    {
+        _containerClient = new BlobContainerClient(options.BlobContainerUri, options.TokenCredential);
+    }
+
+    public void ArchiveMessage(Message message)
+    {
+        var blobClient = _containerClient.GetBlobClient(message.Id.ToString());
+        
+        blobClient.Upload(BinaryData.FromBytes(message.Body.Bytes));
+    }
+
+    public async Task ArchiveMessageAsync(Message message, CancellationToken cancellationToken)
+    {
+        var blobClient = _containerClient.GetBlobClient(message.Id.ToString());
+
+        // var options = new Dictionary<string, string>()
+        //     {
+        //         { "topic", message.Header.Topic },
+        //         { "correlationId", message.Header.CorrelationId.ToString() },
+        //         { "message_type", message.Header.MessageType.ToString() },
+        //         { "timestamp", message.Header.TimeStamp.ToString() },
+        //         { "content_type", message.Header.ContentType }
+        //     };
+        //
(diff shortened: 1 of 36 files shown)


src/Paramore.Brighter.Archive.Azure/AzureBlobArchiveProvider.cs
```diff
@@ -0,0 +1,39 @@
+using Azure.Storage.Blobs;
+
+namespace Paramore.Brighter.Storage.Azure;
+
+public class AzureBlobArchiveProvider : IAmAnArchiveProvider
```
iancooper, line 5: I think the main difference between this and the luggage store is that with the luggage store we only store the payload, not the whole message, hence our use of a stream

src/Paramore.Brighter.Extensions.Hosting/HostedServiceCollectionExtensions.cs
```diff
@@ -23,5 +23,19 @@ public static class HostedServiceCollectionExtensions
             brighterBuilder.Services.AddHostedService<TimedOutboxSweeper>();
             return brighterBuilder;
         }
+
+        public static IBrighterBuilder UseOutboxArchiver(this IBrighterBuilder brighterBuilder,
```
iancooper, line 27: I assume we think most folks will run this stand alone, perhaps with their sweeper?

src/Paramore.Brighter/NullOutboxArchiveProvider.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Paramore.Brighter.Logging;
+
+namespace Paramore.Brighter
+{
+    /// <summary>
+    /// Use this archiver will result in messages just being deleted from the outbox and not stored
+    /// </summary>
+    public class NullOutboxArchiveProvider : IAmAnArchiveProvider
```
iancooper, line 11: LOL, send to null

iancooper (approved):

src/Paramore.Brighter.Archive.Azure/AzureBlobArchiveProvider.cs
preardon: Completely agree, it was an optional extra here but I thought it was a Nice Touch however it looks like azure doesn't want to give me access to do this 🤣

src/Paramore.Brighter.Extensions.Hosting/HostedServiceCollectionExtensions.cs
preardon: That's the plan, I'm looking at putting this in the same process as outbox sweeper so I've got an "Outbox management" service

src/Paramore.Brighter/NullOutboxArchiveProvider.cs
preardon: Well I had to ways to go, either write code in to give you an "Action" or default register you a don't write store. On the fence weather I like it, just know I don't currently hate it, thoughts @iancooper

src/Paramore.Brighter.Extensions.Hosting/TimedOutboxArchiverOptions.cs
```diff
@@ -0,0 +1,20 @@
+namespace Paramore.Brighter.Extensions.Hosting
+{
+    public class TimedOutboxArchiverOptions
+    {
+        /// <summary>
+        /// The timer interval in hours.
```
preardon, line 6: Must confirm or change this to second

Issue #2402, opened 2022-12-06
[Feature] Customization of Casing of Kafka Message's Header Names


siewchinglim:
I would like to create a Kafka message with headers named with camel case. However, Brighter has fixed the header names in the following code. I would like to request a feature to provide flexibility for us to configure the casing of header names.

> namespace Paramore.Brighter.MessagingGateway.Kafka
{
    public class HeaderNames
    {
        /// <summary>
        /// The message type
        /// </summary>
        public const string MESSAGE_TYPE = "MessageType";
        /// <summary>
        /// The message identifier
        /// </summary>
        public const string MESSAGE_ID = "MessageId";
        /// <summary>
        /// The correlation id
        /// </summary>
        public const string CORRELATION_ID = "CorrelationId";
        /// <summary>
        /// The time that message was sent
        /// </summary>
        public const string TIMESTAMP = "TimeStamp";
        /// <summary>
        /// The topic
        /// </summary>
        public const string TOPIC = "Topic";
        /// <summary>
        /// The key used to partition this message
        /// </summary>
        public static string PARTITIONKEY = "PartitionKey";
        /// <summary>
        /// What is the offset into the partition of the message
        /// </summary>
        public static string PARTITION_OFFSET = "TopicPartitionOffset";
    }
}
`

iancooper:
Hi @siewchinglim, those defaults work because we support multiple transports, but we could look at an option that would allow you to choose camel case for the headers

Thijmen:
@iancooper a nice solution could be that users can provide their own header implementation, how would you feel about that?

iancooper:
@Thijmen @siewchinglim So I suspect that we need to use our JsonSerialisationOptions to control how this works. Looking at other transports I can see that we use this to control how header serialisation occurs. I'll put something together on this

iancooper:
@honkuan86 Go for it.

honkuan86:
hi @iancooper, may I know when we can get this release ?

iancooper:
We are just discussing it. Shortly in a pre-release, and then a bit later on a final, once all the current set of features are documented and ***@***.*** 1 Feb 2023, at 13:42, Yong Hon Kuan ***@***.***> wrote:﻿
hi @iancooper, may I know when we can get this release ?

PR #2487, opened 2023-01-19
Kafka message header builder implementation

honkuan86:
This PR include changes to allow custom implementation of a Kafka message header builder. The current implementation does not allowed to customize the header key and values before sending to a Kafka topic. This changes should allow customization of the message header such as camelCasing.

Related issue: #2402


iancooper:
Hi @honkuan86,

I just wanted to understand intent here. The goal is to enable you to customize how the headers are produced. But at first glance it seems that to do that you offer the facility to inject a whole new class to build headers. That seems as though it is putting a lot of work on the implementer, if their goal is to simply shift to camel casing.

Another option would be a set of configuration options that you can pass in (think of the way you can customize System.Text.Json's json serializer by using JsonSerializerOptions: https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/configure-options?pivots=dotnet-7-0)

Is there a particular reason your design went down this path?

honkuan86:
hi @iancooper ,

Yes, apart from changing to camel case, we also want to have a different value on the timestamp, which is a ISO8601 compliant format timestamp, the default one is a unix timestamp, so I think by providing an option for the implementer to customize the way to build the header would be flexible. 

But I think most people would just go for the default one, only those who has a needs to customize the headers they'll have the option to fully customize it.

iancooper:
@honkuan86 Understood. I'll merge. We may return to it at some point, if we think that configuration might be a better way to handle it, but let's fix the problem for you for now

iancooper (approved):

iancooper:
@honkuan86 Do you feel up to adding some information on this to the Docs repo? I'll check in on the other maintainers and what we have scheduled to determine next available release

iancooper:
@honkuan86 Out in 9.4.0, sorry for the delay

PR #2497, opened 2023-02-06
2459 multi type bulk

preardon:
Its not the Prettiest, but it does this job


iancooper (approved):

iancooper:
Looks like it will solve the issue. Nice work.

Issue #2514, opened 2023-02-17
Exception in new version when `DOTNET_ENVIRONMENT` environment variable is set to "Development"


dracan:
### Describe the bug

Exception is thrown in new version when `DOTNET_ENVIRONMENT` environment variable is set to "Development".

### To Reproduce

* Create a new dotnet project... `dotnet new worker`
* Do a bare-minimal Brighter "hello world" example...

```csharp
var host = Host.CreateDefaultBuilder()
    .ConfigureServices(services =>
    {
        services.AddBrighter().AutoFromAssemblies();
    })
    .UseConsoleLifetime()
    .Build();

var commandProcessor = host.Services.GetService<IAmACommandProcessor>();
commandProcessor.Send(new MyCommand("Bob"));

host.WaitForShutdown();
```

```csharp
internal class MyCommand : Command
{
    public MyCommand(string name) : base(Guid.NewGuid())
    {
        Name = name;
    }

    public string Name { get; }
}
```

```csharp
internal class MyHandler : RequestHandler<MyCommand>
{
    [RequestLogging(step: 1, timing: HandlerTiming.Before)]
    public override MyCommand Handle(MyCommand command)
    {
        Console.WriteLine($"Hello {command.Name}");
        return base.Handle(command);
    }
}
```

* Run it, and notice the following exception...

```
System.AggregateException: Some services are not able to be constructed (Error while validating the service descriptor 'ServiceType: Paramore.Brighter.Transforms.Transformers.ClaimCheckTransformer Lifetime: Singleton ImplementationType: Paramore.Brighter.Transforms.Transformers.ClaimCheckTransformer': Unable to resolve service for type 'Paramore.Brighter.Transforms.Storage.IAmAStorageProviderAsync' while attempting to activate 'Paramore.Brighter.Transforms.Transformers.ClaimCheckTransformer'.)
 ---> System.InvalidOperationException: Error while validating the service descriptor 'ServiceType: Paramore.Brighter.Transforms.Transformers.ClaimCheckTransformer Lifetime: Singleton ImplementationType: Paramore.Brighter.Transforms.Transformers.ClaimCheckTransformer': Unable to resolve service for type 'Paramore.Brighter.Transforms.Storage.IAmAStorageProviderAsync' while attempting to activate 'Paramore.Brighter.Transforms.Transformers.ClaimCheckTransformer'.
[...]

PR #2516, opened 2023-02-20
Missing Dependencies for Built In Types break with validation

iancooper:
This PR addresses #2514 where when the ServiceCollection is validated when it is built (the DefaultServiceProviderFactory call to CreateServiceProvider when your options have been set to validate) and we have missing dependencies for built-in types that we have registered.

This is valuable behaviour but not for built in types that you are not using.


iancooper:
@dracan for information

preardon (approved):

PR #2537, opened 2023-02-28
Make Kafka Consumer Reading Message Headers Case Insensitive

honkuan86:
This PR contains changes to make the Kafka consumer to read the message header case insensitive. 

This changes comes after we have added the custom header builder which allow coders to customize the message header to camel case and causes the consumer to ignore the headers due to case sensitivity. So the intention of this PR is to make the consumer read the headers case insensitive.


iancooper (approved):

iancooper:
@honkuan86 Thanks. We should be getting a release out soon, so I have merged

iancooper:
@honkuan86 Should be in 9.4.1

Issue #2512, opened 2023-02-16
KafkaPublication.NumPartitions not working


sheepy-jy:
### Describe the bug
Kafka topic's partition is always set to default 1, even if I set KafkaPublication.NumPartitions to 3.
To make sure its a issue on brighter side, I have used the below ways to create topic, and they all are working:
1. Create topic with kafka-ui
2. Create topic with kafka cli
3. Create topic by directly using Confluent.Kafka library instead of via brighter

### To Reproduce
1. Normal configuration documented in https://brightercommand.gitbook.io/paramore-brighter-documentation/brighter-configuration/kafkaconfiguration#publication
2. Set NumPartitions to any number other than the default

### Exceptions (if any)
N/A

### Further technical details
- Brighter version 9.3.6
- The OS (Windows)

iancooper:
Thanks @sheepy-jy we will take a look

iancooper:
@sheepy-jy I can recreate. It's a curious one as we are using the Confluent Admin Client and passing the value through, but it does not seem to be working. We'll need to dig into the Confluent Admin Client's code to see if that helps, but at some point is just calls librdkafka and is a black box.

Let you know when I know more

iancooper:
Interestingly, I get an error back saying that the topic already exists, so it appears to be some sort of double creation happening. which I need to track down.

iancooper:
It may be that we need to be sure that we have disabled auto-creation of a topic, so that it is not being created by the server. In theory this ought not to happen because we should create the topic before using it, but it may be useful to ensure this is "off" to remove the issue. I'll check on that

sheepy-jy:
Hi @iancooper , can I say that this will fix the double creation issue by checking on the number of partitions and the replication factor, but we should still turn off the auto-creation configuration in kafka broker?

iancooper:
@sheepy-jy So I believe the issue is the configuration of Kafka. If you have auto-create topics turned on, then it will create the topic with a single partition when we look for it. If that setting is not enabled, we can create it, with the relevant number of partitions.

I have updated our Kafka docker compose file with the environment setting:

KAFKA_AUTO_CREATE_TOPICS_ENABLE: "false"

And then I create the desired number of partitions.

I have cleaned up the error handling in our code that searches for a valid topic, so it's possible it is a combination of both, but the clean up to the code on its own, does not fix this.

iancooper:
[...]

PR #2548, opened 2023-03-07
[Bug] Kafka Topic Partitions not Matching Publication

iancooper:
As per #2512 the admin client is not setting the num of partitions correctly. This fixes that.

-- It creates a stronger check when ensuring a topic that not only is it found by name, but that the  number of partitions and the replication factor match the publication; if not it will error.


iancooper:
Confusing errors caused by docker container setting auto creation of topics, which will create before we can

PR #2560, opened 2023-03-20
Binary payloads

iancooper:
We want to support binary payloads for messages. When an Outbox stores a binary payload as a UTF-8 string this can be lossy, so we would prefer to support native binary columns. However, as some preference exists for readable JSON for debugging, and for backwards compatibility, we are making support for binary payloads an option.

Because binary payloads may be large, we also take advantage of any Db support for handling large binary fields where appropriate.


src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -493,13 +453,41 @@ await handlerChain.First().HandleAsync(command, cancellationToken)
...
         /// <param name="request">The request.</param>
+        /// <param name="transactionProvider">The transaction provider</param>
+        /// <typeparam name="TRequest">The type of request</typeparam>
+        /// <typeparam name="TTransaction">The type of transaction used by the Outbox</typeparam>
         /// <exception cref="System.ArgumentOutOfRangeException"></exception>
-        public void Post<T>(T request) where T : class, IRequest
+        public void Post<TRequest, TTransaction>(TRequest request, IAmABoxTransactionProvider<TTransaction> transactionProvider) where TRequest : class, IRequest
         {
-            ClearOutbox(DepositPost(request, null));
+            ClearOutbox(DepositPost(request, transactionProvider));
```
preardon, line 463: If you write the IRequest into a transaction then one two following things should happen
  - when we go to clear if we don't select inside of the txn then there will be no message to dispatch
  - if we do select inside of the transaction then if we roll back the txn the message will already be sent

src/Paramore.Brighter/IAmACommandProcessor.cs
```diff
@@ -42,53 +42,103 @@ public interface IAmACommandProcessor
...
+        void Post<TRequest>(TRequest request) where TRequest : class, IRequest;
+        
+        /// <summary>
+        /// Posts the specified request.
+        /// </summary>
+        /// <typeparam name="TRequest">The type of the request</typeparam>
+        /// <typeparam name="TTransaction">The type of any transaction used by the request</typeparam>
+        /// <param name="request">The request.</param>
+        /// <param name="transactionProvider">If you wish to use an outbox with this request, the transaction provider for that outbox</param>
+        void Post<TRequest, TTransaction>(
```
preardon, line 94: See comment on implementation around this

src/Paramore.Brighter.MsSql/MsSqlSqlAuthUnitOfWork.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Data.SqlClient;
+
+namespace Paramore.Brighter.MsSql
+{
+    /// <summary>
+    /// A connection provider for Sqlite 
+    /// </summary>
+    public class MsSqlSqlAuthUnitOfWork : RelationalDbTransactionProvider
```
preardon, line 13: Should this be Sql Auth Unit of work?

iancooper: I was trying to change it to MsSqlUnitOfWork as we only have the one

src/Paramore.Brighter/CommandProcessor.cs
iancooper: Will review. I was trying to: provide consistent interfaces for transactions rather than force casting, in the handler; eliminate generics from the command processor interface that could result from that. It may be that it can't be null because we have to infer a transaction type, but it may be that we can just use the new default CommittableTransaction type, which can effectively act as a "do nothing"

preardon (approved):

PR #2583, opened 2023-04-01
remove some redundant parens

SimonCropp:


iancooper (changes requested):
Not sure, stylistically I like the parentheses because they remind me that this is a parameter

PR #2584, opened 2023-04-01
use method groups

SimonCropp:


iancooper (changes requested):
I don't tend to prefer method groups. I'd reject this unless there was a compelling reason

SimonCropp:
for me it is simply less code. but it is a personal preference. so feel free to close. no feelings hurt 😜

iancooper:
@SimonCropp Broadly, a lambda is more performant than a method group: 

https://www.alwaysdeveloping.net/dailydrop/2023/01/24-lambda-vs-method/ 

I'll close.

PR #2591, opened 2023-04-02
fix typo in TimeoutInMilliseconds

SimonCropp:


src/Paramore.Brighter/Subscription.cs
```diff
@@ -117,7 +117,15 @@ public class Subscription
...
-        public int TimeoutInMiliseconds { get; }
+        [Obsolete("Use TimeoutInMilliseconds")]
+        public int TimeoutInMiliseconds { get => TimeoutInMilliseconds; }
+
+        /// <summary>
+        /// Gets the timeout in milliseconds that we use to infer that nothing could be read from the channel i.e. is empty
+        /// or busy
+        /// </summary>
+        /// <value>The timeout in mi;liseconds.</value>
+        public int TimeoutInMilliseconds { get; }
```
iancooper, line 128: Because this is a public method on a class exposed to consumers, this forms a breaking change, and we would need to version. To fix this, you need a new property with the correct name, and you need the old property to use that new property for it's implementation. Then you need to mark the old property as obsolete, note that folks should use the correctly spelled one and indicate that we will switch over in v10

iancooper (changes requested):
We like to only version if we were in beta, or we up a whole version, which we prefer to do once a year.

src/Paramore.Brighter/Subscription.cs
SimonCropp: thats what i did

iancooper: Not sure how I missed that :-D

Well, I guess you did it our preferred way then...

iancooper (approved):

PR #2597, opened 2023-04-04
cleanup Paramore.Brighter.Outbox.PostgreSql

SimonCropp:


iancooper (approved):

src/Paramore.Brighter.Outbox.PostgreSql/PostgreSqlOutboxSync.cs
```diff
@@ -254,10 +238,8 @@ public void MarkDispatched(Guid id, DateTime? dispatchedAt = null, Dictionary<st
 
             try
             {
-                using (var command = InitMarkDispatchedCommand(connection, id, dispatchedAt))
-                {
-                    command.ExecuteNonQuery();
-                }
+                using var command = InitMarkDispatchedCommand(connection, id, dispatchedAt);
```
iancooper, line 241: You can't use Using Declarations as we are C# 7.3 due to supporting .NET Standard 2.0 if you look in the file section of GitHub you should be able to see the build errors against the files.

(We support netstandard20 as this gives us framework and core support. Dropping framework support is something we might consider in V10, but the principle of a library being as compatible as important remains important to us)

iancooper (changes requested):

iancooper:
We have changes that will impact this one, as we are moving PostgreSql to the same model as other Outboxes. I am going to close this one as a result.

SimonCropp:
@iancooper no worries. do you want more of these type of prs?

iancooper:
Happy to keep taking PRs. This one just collided with a major restructuring of that code.

PR #2598, opened 2023-04-04
remove duplicate dictionary lookups

SimonCropp:


iancooper (approved):
Thanks, I don't think we ever went back to this after the api for the dictionary added this.

PR #2604, opened 2023-04-09
remove empty return comments

SimonCropp:


iancooper (changes requested):
The issue here is that most of these need completion, not removal. The XML-docs are useful because they provide help via intelisense. So we should either update them, or close this PR

PR #2617, opened 2023-04-12
Fix for Dockerfile smell DL3008

grosa1:
Hi!
The Dockerfile placed at "Docker/RabbitMQ/Dockerfile" contains the best practice violation [DL3008](https://github.com/hadolint/hadolint/wiki/DL3008) detected by the [hadolint](https://github.com/hadolint/hadolint) tool.

The smell DL3008 occurs when the version pinning for the installed packages with apt is not specified. This could lead to unexpected behavior when building the Dockerfile.
In this pull request, we propose a fix for that smell generated by our fixing tool. We have verified that the patch is correct before opening the pull request.
To fix this smell, specifically, we use a heuristic approach that selects the most probable version tag for a given apt package corresponding to the latest version at the current date. The package versions are retrieved using the Canonical Launchpad APIs.

This change is only aimed at fixing that specific smell. If the fix is not valid or useful, please briefly indicate the reason and suggestions for possible improvements.

Thanks in advance

### DIFF
--- a/Docker/RabbitMQ/Dockerfile
+++ b/Docker/RabbitMQ/Dockerfile
@@ -2,7 +2,7 @@ FROM rabbitmq:3-management
 
 RUN apt-get update
 
-RUN apt-get install -y curl && apt-get install -y zip
+RUN apt-get install -y curl=7.68.* && apt-get install -y zip=3.0-*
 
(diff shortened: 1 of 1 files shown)


iancooper:
Thanks for this. This Dockerfile is used to support testing, and we want to test against the current version. For this reason we don't pin the version. As a result I am going to close this issue.

Issue #2622, opened 2023-04-17
[Bug] The UsePolicyAsync support for policy wrap is not implemented


iancooper:
### Describe the bug
Given the decision to restrict UsePolicy and UsePolicyAsync to a single use attribute (due to IoC issues), we need to support the array of policies for UsePolicyAsync. They are in UsePolicy and were omitted by mistake.

PR #2627, opened 2023-04-17
Support multiple policy async

iancooper:
We moved support for multiple polices to the attribute, so as to avoid issues when we tried to create multiple instances of the handler with Scoped or Singleton lifetimes.

We forgot async as part of that.


iancooper:
See the associated issue #2622

preardon (approved):

PR #2677, opened 2023-05-23
Fix SQL Archiving

preardon:
Looks like this Ignored the - and as such does not respect the minimum age of messages


DevJonny (approved):
LGTM

PR #2691, opened 2023-06-11
Fix UnwrapWith not setting base step value

### LINKED ISSUE #2616: [Bug]Missing base step value in UnwrapWithAttribute
https://github.com/BrighterCommand/Brighter/blob/729889958326fbdd8dea94ee338dc32848690293/src/Paramore.Brighter/UnwrapWithAttribute.cs#L40

UnwrapWithAttribute does not set the step value to base step (TransformAttribute)

image

image

KieranBond:
Update UnwrapWith to set base Step property

Resolves #2616

### DIFF
--- a/src/Paramore.Brighter/UnwrapWithAttribute.cs
+++ b/src/Paramore.Brighter/UnwrapWithAttribute.cs
@@ -24,7 +24,7 @@ THE SOFTWARE. */
 using System;
 
 namespace Paramore.Brighter
-{         
+{
     /// <summary>
     /// class UnwrapWithAttribute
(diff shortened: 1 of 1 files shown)


iancooper (approved):

KieranBond:
Thanks for your approval @iancooper. Looks like the two failing checks are for reasons unassociated with the changes.

Can we skip them? Or do we need to retry the actions?

preardon (approved):
Looks good, I can merge this in if you would like

KieranBond:
If you're happy, I'm happy 😅

Issue #2757, opened 2023-08-04
[Bug] In Service Activator an unhandled exception is not been logged as an error


holytshirt:
### Describe the bug
 In Service Activator an unhandled exception is not been logged as an error when sendsync or publishasync.
 The AggregateException does not log an error and the message is acked

PR #2758, opened 2023-08-05
fix(SeviceActivator): Logs unhandled exceptions in handlers as error …

holytshirt:
…on publishasync (#2757)

### DIFF
--- a/src/Paramore.Brighter.ServiceActivator/MessagePump.cs
+++ b/src/Paramore.Brighter.ServiceActivator/MessagePump.cs
@@ -176,59 +176,76 @@ namespace Paramore.Brighter.ServiceActivator
 
                     span?.SetStatus(ActivityStatusCode.Ok);
                 }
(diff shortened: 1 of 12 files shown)


holytshirt:
I in-lined the method as it was hard to compare and think about the exception handling for a Exception and AggregateException

iancooper (approved):
Looks good. I don't think the handling of aggregate exceptions is so long that we need to necessarily wonder about inlining. I do wonder if at some point we should look at commonality though between individual vs aggregate and factor out the action as a method. Not for now though.

holytshirt:
> handling

It used to be in it's own method, I took out it's method and put it in place. I think that is why the logging was missing as it was hard to reason that the aggregate exception and normal exception should be similar

holytshirt:
I managed to get the logging tests working and now covered by tests. Renamed the requeue ones to what it is testing.

Issue #2810, opened 2023-09-05
DynamoDB Outbox Hot Partition


jtsalva:
### Describe the bug
`Paramore.Brighter.Outbox.DynamoDB` throttling on `DynamoDbOutbox.OutstandingMessages`. This is despite the table and GSIs being 'on-demand/pay per request'. According to AWS docs 'A hot partition can cause throttling if the partition limits of **3000 RCU and 1000 WCU per second are exceeded**.'

The schema for the Outstanding index is:
```
hash_key_name: Topic
hash_key_type: STRING
range_key_name: CreatedTime
range_key_type: NUMBER
```

We encounter a hot partition where we're publishing only to a single topic at a rate exceeding the limit imposed by AWS Dynamo itself.

### Exceptions
```
System.AggregateException: One or more errors occurred. (Throughput exceeds the current capacity for one or more global secondary indexes. DynamoDB is automatically scaling your index so please try again shortly.)
 ---> Amazon.DynamoDBv2.AmazonDynamoDBException: Throughput exceeds the current capacity for one or more global secondary indexes. DynamoDB is automatically scaling your index so please try again shortly.
 ---> Amazon.Runtime.Internal.HttpErrorResponseException: Exception of type 'Amazon.Runtime.Internal.HttpErrorResponseException' was thrown.
   at Amazon.Runtime.HttpWebRequestMessage.GetResponseAsync(CancellationToken cancellationToken)
   at Amazon.Runtime.Internal.HttpHandler`1.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.Unmarshaller.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.ErrorHandler.InvokeAsync[T](IExecutionContext executionContext)
   --- End of inner exception stack trace ---
   at Amazon.Runtime.Internal.HttpErrorResponseExceptionHandler.HandleExceptionStream(IRequestContext requestContext, IWebResponseData httpErrorResponse, HttpErrorResponseException exception, Stream responseStream)
   at Amazon.Runtime.Internal.HttpErrorResponseExceptionHandler.HandleExceptionAsync(IExecutionContext executionContext, HttpErrorResponseException exception)
   at Amazon.Runtime.Internal.ExceptionHandler`1.HandleAsync(IExecutionContext executionContext, Exception exception)
[...]

PR #2813, opened 2023-09-07
Add random sharding for the DynamoDb outbox

jtsalva:
Addresses #2810

- Add configuration for number of shards (default 3, max 20)
- Add optional configuration for TTL (default null - so forever)
- Applies random sharding when writing items into the dynamo outbox for the 'Outstanding' index PK
- Query all shards when getting all outstanding messages

### DIFF
--- a/src/Paramore.Brighter.Outbox.DynamoDB/DynamoDbOutbox.cs
+++ b/src/Paramore.Brighter.Outbox.DynamoDB/DynamoDbOutbox.cs
@@ -26,6 +26,7 @@ THE SOFTWARE. */
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Security.Cryptography;
 using System.Threading;
 using System.Threading.Tasks;
 using Amazon.DynamoDBv2;
@@ -55,7 +56,11 @@ namespace Paramore.Brighter.Outbox.DynamoDB
         {
             _configuration = configuration;
             _context = new DynamoDBContext(client);
-            _dynamoOverwriteTableConfig = new DynamoDBOperationConfig { OverrideTableName = _configuration.TableName };
+            _dynamoOverwriteTableConfig = new DynamoDBOperationConfig
+            {
+                OverrideTableName = _configuration.TableName,
+                ConsistentRead = true
+            };
         }
 
         /// <summary>
@@ -90,7 +95,9 @@ namespace Paramore.Brighter.Outbox.DynamoDB
         /// <param name="cancellationToken">Allows the sender to cancel the request pipeline. Optional</param>        
         public async Task AddAsync(Message message, int outBoxTimeout = -1, CancellationToken cancellationToken = default, IAmABoxTransactionConnectionProvider transactionConnectionProvider = null)
         {
-            var messageToStore = new MessageItem(message);
+            var shard = GetShardNumber();
+            var expiresAt = GetExpirationTime();
+            var messageToStore = new MessageItem(message, shard, expiresAt);
 
             if (transactionConnectionProvider != null)
             {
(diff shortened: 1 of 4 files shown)


jtsalva:
@iancooper I'm not sure how to trigger CI

src/Paramore.Brighter.Outbox.DynamoDB/DynamoDbOutbox.cs
```diff
@@ -413,6 +398,49 @@ private async Task<IEnumerable<MessageItem>> PageAllMessagesAsync(QueryOperation
 
             return messages;
         }
+        
+        private async Task<IEnumerable<MessageItem>> QueryAllOutstandingShardsAsync(string topic, DateTime minimumAge, CancellationToken cancellationToken = default)
+        {
+            using var semaphore = new SemaphoreSlim(20);
```
jtsalva, line 404: No reasoning behind this being 20, maybe should be configurable

src/Paramore.Brighter.Outbox.DynamoDB/DynamoDbOutbox.cs
```diff
@@ -55,7 +56,11 @@ public DynamoDbOutbox(IAmazonDynamoDB client, DynamoDbConfiguration configuratio
         {
             _configuration = configuration;
             _context = new DynamoDBContext(client);
-            _dynamoOverwriteTableConfig = new DynamoDBOperationConfig { OverrideTableName = _configuration.TableName };
+            _dynamoOverwriteTableConfig = new DynamoDBOperationConfig
+            {
+                OverrideTableName = _configuration.TableName,
+                ConsistentRead = true
```
jtsalva, line 62: Setting `ConsistentRead = true` as default for non-GSI lookups to mitigate against `NullReferenceException`s we've been seeing when calling `PostAsync`

iancooper:
> @iancooper I'm not sure how to trigger CI

GitHub feature, one of us has to approve you when you have not triggered one before. It's to stop you uploading something that will start bitcoin mining or the like

src/Paramore.Brighter.Outbox.DynamoDB/DynamoDbOutbox.cs
iancooper: What is the reason for the semaphore?

jtsalva: I now realise maybe doing a `Parallel.ForEachAsync` is more suitable? I was thinking we need to set `MaxDegreeOfParallelism` as max number of shards is unbounded

iancooper: I wonder if we could sensibly limit the number of shards. At some point we only want so many threads, so many open sockets etc. If someone exceeds that number in order for their outbox to work, I wonder if an outbox is the right solution to transactional messaging for them.

It is a good question though

iancooper: i think we should set an upper limit when you configure the number of partitions. Let's say 20 for now and throw an error if you use more than that.

I wonder if you need more than 20 partitions whether you should first think about using DAX over increasing the number of partitions further.

jtsalva: That makes sense, will simplify to remove semaphores and just set max shards to 20. Can always be another PR if we find 20 isn't enough.

Speaking of DAX, I'm not sure how that would solve the hot partitioning issue, I think it would alleviate maybe some reads and reduce latency but underlying issue would still be there? (from someone who's only just read about DAX on the surface)

iancooper: Looks like we have reached agreement to limit at config, which is more explicit.

In principle a write through-cache will limit a hotspot as our partition would tend to be held in memory not on disk so access would not hit an RCU or WCU limit. For "hot spots" DAX is often a good solution, particularly in this case where you write and then read shortly after what you just read, and can evict older elements (which would be dispatched messages for us). I would guess that a write through cache would resolve this issue. So if our partitioning strategy gets too many partitions and thus requires a lot of threads on your Sweeper process it may be more efficient to move to an in-memory cache.

jtsalva: Thanks for explaining Ian

Issue #2837, opened 2023-09-24
Add DynamoDbInbox Configurability


jtsalva:
There can be multiple applications deployed to the same AWS account requiring their own separate inbox. [DynamoDbInboxConfiguration](https://github.com/BrighterCommand/Brighter/blob/master/src/Paramore.Brighter.Inbox.DynamoDB/DynamoDbInboxConfiguration.cs) exists but there doesn't seem to be any references. 

*Proposed solution*
Follow the same pattern as [DynamoDbConfiguration](https://github.com/BrighterCommand/Brighter/blob/master/src/Paramore.Brighter.Outbox.DynamoDB/DynamoDbConfiguration.cs) being used by the [DynamoDbOutbox](https://github.com/BrighterCommand/Brighter/blob/master/src/Paramore.Brighter.Outbox.DynamoDB/DynamoDbOutbox.cs)

PR #2838, opened 2023-09-24
Add DynamoDbInbox configurability

jtsalva:
Addresses #2837


iancooper (approved):

iancooper:
Thanks for this, makes sense to fix up the Inbox too.

PR #2860, opened 2023-10-08
[Bug] Ensured flush token is released if offset commit fails.

lsphillips:
When flushing offsets, the commit can fail in the background task created. This results in the flush token not being released preventing all future sweeps from never happening causing the offset for the partition to never be committed in cases where auto commit is disabled. This is a tiny change that should ensure the consumer recovers gracefully in the event of a commit error.


iancooper (approved):

iancooper:
Thanks @lsphillips. Good spot!

PR #2861, opened 2023-10-09
Azure Blob Archive - Storage Pattern

preardon:
Add the ability to specify the location and name of the Blobs inside of the container


iancooper (approved):

iancooper:
@preardon approved

PR #2936, opened 2023-12-01
Allow for Async Message Mapper Pipelines

iancooper:
Previous versions had async transformers and sync message mappers, to avoid breaking changes in V9. With V10 we want to support a sync transform and mapper pipeline and an async transformer and mapper pipeline. This allows for I/O such as a claim check or schema registry.


iancooper:
Merged. Will release post bug fix for GUID serialization and updating the documentation

PR #2943, opened 2023-12-09
Feature 2942 Support NetStandard 2.0 for SQLite projects

SteveBush:
Add NetStandard2.0 support for Sqlite projects.  Make asynchronous calls on NetStandard2.1, synchronous and without cancellation support on NetStandard 2.0.

I plan to use SQLite support on a Net462 project without only supports NetStandard2.0.


iancooper:
Going to add to release/9 so that we can merge in to v10

iancooper:
@SteveBush This change should have been applied via V9. Do you want to check, and then I will close this PR?

SteveBush:
> @SteveBush This change should have been applied via V9. Do you want to check, and then I will close this PR?

Yes.  Thanks.  I have verified b14a82c8171d26b745ebd0159cf718cac45c176a fixes the issue.

Issue #2944, opened 2023-12-10
[Bug] Sqlite Queries don't support millisecond time comparisons


SteveBush:
### Describe the bug
In the SQLite queries for PageDispatchedCommand and PagedOutstandingCommand, strftime is called with '%s' which returns seconds since 1970-01-01.  This value is then multiplied by 1000 to get milliseconds and compared to OutstandingSince.  

### To Reproduce
Running the SQLite When_there_are_multiple_messages_and_some_are_received_and_Dispatched_bulk_Async test on a fast machine will fail unless you add a one-second delay between MarkDispatchedAsync and OutstandingMessagesAsync.  I have verified waiting one second solves the comparison resolution issue.

In other words, the resolution of OutstandingSince is in milliseconds and the comparison is in seconds converted to milliseconds.

The fix is to change the SQL below to compare now - Dispatched  to @OutstandingSince with millisecond resolution.
```
         public string PagedDispatchedCommand { get; } = "SELECT * FROM {0} WHERE DISPATCHED IS NOT NULL AND (strftime('%s', 'now') - strftime('%s', Dispatched)) * 1000 < @OutstandingSince ORDER BY Timestamp ASC LIMIT @PageSize OFFSET (@PageNumber-1) * @PageSize";
        public string PagedOutstandingCommand { get; } = "SELECT * FROM {0} WHERE DISPATCHED IS NULL AND (strftime('%s', 'now') - strftime('%s', TimeStamp)) * 1000 > @OutstandingSince ORDER BY Timestamp ASC LIMIT @PageSize OFFSET (@PageNumber-1) * @PageSize";

```

### Further technical details
- Brighter version
- Include the output of `dotnet --info`
- The OS (Window, Linux, MacOs, etc)

SteveBush:
Verified code in 9.7.6.  What are the plans to release 10.0.0-preview.2 with 9.7.6 merged fixes?

PR #2941, opened 2023-12-08
Add 1 MS deplay to CleanUpDb to allow time for test{guid}.db to fully…

SteveBush:
Pull request for bug #2940 

Added code to CleanUpDb to wait for one millisecond to allow test{guid}.db to fully close.  In the process made CleanUpDb async.  I decided not to rename CleanUpDbAsync which I would normally have done.

PR #2945, opened 2023-12-10
Convert a DateTime to a string representation with millisecond accuracy

SteveBush:
#2944 .  In Sqlite implementation, the timestamp has only second precision.  as it was using "s".  Convert it to a dateime format with millisecond accuracy.

Note, I chose not to convert it here to UniversalTime().  Dispatch converts to universal time so maybe we should be consistent here.

DispatchedAt?.ToUniversalTime()

### DIFF
--- a/src/Paramore.Brighter.Outbox.Sqlite/SqliteOutbox.cs
+++ b/src/Paramore.Brighter.Outbox.Sqlite/SqliteOutbox.cs
@@ -27,6 +27,7 @@ using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Data.Common;
+using System.Globalization;
 using System.Text.Json;
 using System.Threading;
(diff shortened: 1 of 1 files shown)


SteveBush:
Note, you will need to merge in #2941 before the tests will succceed.

iancooper:
As this changes the format of a datetime string, I will probably push this to V10 as we can document the change

SteveBush:
> As this changes the format of a datetime string, I will probably push this to V10 as we can document the change

Sounds good.  To be specific, the consumer format of datetime does not change just the precision and format that is stored in the database internally.  DateTime has an accuracy of 10ms to 30ms depending on OS and computer.  Unless the consumer is looking at the database directly, they will only see a change in the accuracy of DateTime.

iancooper (approved):

PR #2951, opened 2023-12-11
fix: Allows setting of TransactionConnectionProvider on RPC mode

DevJonny:
This fixes an issue where you are unable to set the implementation of the `IAmABoxTransactionConnectionProvider` when using `useRequestResponseQueues = true`.

It looks like this was just an oversight as the `boxTransactionConnectionProvider` is set to the value of `_overridingBoxTransactionConnectionProvider` later on.

### DIFF
--- a/src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
@@ -356,7 +356,8 @@ namespace Paramore.Brighter.Extensions.DependencyInjection
 
             if (externalBusType == ExternalBusType.None)
                 return messagingBuilder.NoExternalBus();
-            else if (externalBusType == ExternalBusType.FireAndForget)
+            
+            if (externalBusType == ExternalBusType.FireAndForget)
                 return messagingBuilder.ExternalBus(
                     new ExternalBusConfiguration(
                         producerRegistry, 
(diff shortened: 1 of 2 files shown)


iancooper (approved):

Jonny-Freemarket:
> Looks good, this might be different in V10 :-D

I will take another look, but I believe in v10 it was already fixed / not an issue because of the explicit setting of it in the configuration of the `UseExternalBus`. 

I have purposely targeted `release/9` instead of `master` as it's just a v9 fix, but not sure if that's correct?

iancooper:
> I will take another look, but I believe in v10 it was already fixed / not an issue because of the explicit setting of it in the configuration of the `UseExternalBus`.

yes, my assumption too

 
> I have purposely targeted `release/9` instead of `master` as it's just a v9 fix, but not sure if that's correct?

Yes, this is the way.

Feel free to merge. I can get a release out for you after

Jonny-Freemarket:
Thanks for the clarification @iancooper!

Unfortunately I don't have permissions to merge 😃

iancooper:
> Unfortunately I don't have permissions to merge 😃
Done for you. Let me see if I can fix that.

Jonny-Freemarket:
Thank you, and thank you!

iancooper:
> Thank you, and thank you!

You are listed as an Admin on Brighter. Hmmm.

Jonny-Freemarket:
Ah probably because this was created with my work GH account rather than my personal one!

iancooper:
>Ah probably because this was created with my work GH account rather than my personal one!
You should be able to update your GitHub account so that it knows you are either; can't recall how I did that

PR #2993, opened 2024-01-25
feature: Central Package Management

holytshirt:
Moving the management of packages to CPM
https://learn.microsoft.com/en-us/nuget/consume-packages/Central-Package-Management


iancooper (approved):
I had zero idea that we could do this. It will make this a whole lot easier. Thanks.

PR #3011, opened 2024-02-06
Fix MessagePumpAsync to catch error and create MessageMapp…

holytshirt:
…er errors


iancooper (commented):
I think that the catch for these should be in the TranslateMessage method above, to mirror the blocking pump, and keep our synchonizatoncontext free of the catch code

iancooper (approved):

PR #3012, opened 2024-02-07
fix(test): fix issue with aws tests

holytshirt:


tests/Paramore.Brighter.AWS.Tests/TestDoubles/MyLargeCommandMessageMapper.cs
```diff
@@ -21,3 +22,21 @@ public MyLargeCommand MapToRequest(Message message)
...
+        return Task.FromResult(new Message(
+            new MessageHeader(request.Id, "transform.event", MessageType.MT_COMMAND, DateTime.UtcNow),
+            new MessageBody(JsonSerializer.Serialize(request, new JsonSerializerOptions(JsonSerializerDefaults.General))))
+        );
+    }
+
+    [RetrieveClaim(0, retain:false)]
+    public Task<MyLargeCommand> MapToRequest(Message message)
+    {
+        return Task.FromResult(JsonSerializer.Deserialize<MyLargeCommand>(message.Body.Bytes));
```
iancooper, line 40: Not sure if there is a perf advantage, considering a large message, to using async deserialize here i.e. create a memory stream from body bytes, async deserialize from stream, return MyLargeCommand

holytshirt: I can do that, I was testing that this would fix the issue. I can't run the aws tests local atm

iancooper: There should be AWS login information in the tests for when they run on GitHub. But it's a little harder if you don't have your kit setup for that.

holytshirt: @iancooper  made the whole mapper more async-y, also realised we are missing the cancelation token.
Should the `MapToRequest(Message message) ` be `MapToRequestAsync(Message message, CancellationToken cancellationToken = default)` ?

iancooper: Good shout, we should probably keep to our naming convention here.

iancooper: I suspect that both the methods on that interface need renaming

holytshirt: @iancooper I'll do it as a separate PR

PR #3027, opened 2024-02-20
Fix Warnings

iancooper:
We have various warnings in the codebase, mostly due to handling of null, or async that is not. This clouds our warnings with a lot of noise, and means that committers tend to ignore warnings.

This PR tries to fix a lot of these warnings, through fixing the problem of suppressing where the warning is not needed.

We don't fix the warnings around UpdateTelemetry as that needs a separate PR


iancooper:
@holytshirt Comments appreciated. it is a bit of a smorgasboard, in order to get rid of the obsoletes and warnings.

preardon (approved):

holytshirt:
> @holytshirt Comments appreciated. it is a bit of a smorgasboard, in order to get rid of the obsoletes and warnings.

only 131 files changes ....

holytshirt (approved):

iancooper:
> > @holytshirt Comments appreciated. it is a bit of a smorgasboard, in order to get rid of the obsoletes and warnings.
> 
> only 131 files changes ....

Could have been worse, I could have included Otel...

Issue #3056, opened 2024-04-16
Error when number of partition changed using Cooperative rebalance protocol


jsyauideagen:
### Describe the bug
After we increase the number of partition of a topic (e.g. from 2 -> 5), the consumer will hit into rebalance related error.

### To Reproduce
Increase any existing topic's number of partition and you will hit into this error.

### Exceptions (if any)

> %4|1713152471.109|ASSIGN|icsa.notification#consumer-6| [thrd:main]: Group "ICSA.Notification.Services.NotificationProcessor": application *assign() call failed: Changes to the current assignment must be made using incremental_assign() or incremental_unassign() when rebalance protocol type is COOPERATIVE
> Unhandled exception. Confluent.Kafka.KafkaException: Local: Erroneous state
>    at Confluent.Kafka.Impl.SafeKafkaHandle.AssignImpl(IEnumerable`1 partitions, Func`3 assignMethodErr, Func`3 assignMethodError)
>    at Confluent.Kafka.Impl.SafeKafkaHandle.Assign(IEnumerable`1 partitions)
>    at Confluent.Kafka.Consumer`2.Unassign()
>    at Confluent.Kafka.Consumer`2.RebalanceCallback(IntPtr rk, ErrorCode err, IntPtr partitionsPtr, IntPtr opaque)

### Further technical details
- Brighter version - 9.7.6
- .NET SDK:
 Version:           8.0.202
 Commit:            25674bb2f4
 Workload version:  8.0.200-manifests.8cf8de6d

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.22621
 OS Platform: Windows
 RID:         win-x64
 Base Path:   C:\Program Files\dotnet\sdk\8.0.202\

.NET workloads installed:
There are no installed workloads to display.

Host:
  Version:      8.0.3
  Architecture: x64
  Commit:       9f4b1f5d66

.NET SDKs installed:
  2.1.818 [C:\Program Files\dotnet\sdk]
  3.1.426 [C:\Program Files\dotnet\sdk]
  5.0.408 [C:\Program Files\dotnet\sdk]
  6.0.203 [C:\Program Files\dotnet\sdk]
  8.0.103 [C:\Program Files\dotnet\sdk]
  8.0.202 [C:\Program Files\dotnet\sdk]

.NET runtimes installed:
  Microsoft.AspNetCore.All 2.1.30 [C:\Program Files\dotnet\shared\Microsoft.AspNetCore.All]
[...]

PR #3059, opened 2024-04-16
Kafka  Bug Fixes

iancooper:
Fixes from V9

- Drops support for Co-operative Sticky rebalancing for now. (Defaults to Round Robin and gives an exception if you try to set co-operative).  There is an issue around this you can track: https://github.com/BrighterCommand/Brighter/issues/3056 . It may take some thought because we need to know if we are rebalancing when we try to commit an offset.
- Removed a spurious call to StoreOffset. As we have our own offset store this was not being used.


MaorDavidzon:
@iancooper It is a big limitation, 
Is supporting comparative sticky with manual commit on the roadmap? This is a common use case and it’s already implemented in other clients.

PR #3067, opened 2024-04-25
Add Cloud Events Attributes to Message Header and Publication.

iancooper:
Our next step on the path to Cloud Events is to amend the header and publication to contain Cloud Events attributes. As part of this work we are also going to:

* Write a Cloud Events Transformer i.e. [CloudEvents] that puts Cloud Events metadata from the Publication or Attribute parameters into the Message Header. This makes it easy to write tests around the properties being set, forcing them to be created.
* Source Telemetry information to populate the traceparent, tracestate and baggage fields

Whilst we will amend transport code so that it compiles, this PR won't actually read-write over a given transport, or read-write to a given Outbox. We will raise a set of PRs for those, one per technology, after.


iancooper:
Because of the difficulties with merging this code, I have merged into master with the RequestContext. I will then re-open this PR for the open telemetry changes.

PR #3077, opened 2024-05-11
Use Time Provider to fix time related tests

iancooper:
.NET 8 adds TimeProvider which can be used to provide system time; it also supports FakeTimeProvider which can be used to fake an advancing clock, provide a fake timer etc. There is a backport of the interface and library to netstandard2.0

We have tests that rely on time, specifically where we:

* Clear a cache following expiry of a time period (in memory box)
* Determine if items in an Outbox are outstanding.

Many of these come together in Paramore.Brighter.InMemory.Tests. We stopped running these because they were flaky in contested CI environments as they relied on delaying action/assertion until time had passed.

With this change we use TimeProvider with our InMemoryBox (used by default if you don't provide an out of process alternative). This allows us to advance time for use with the tests around this functionality. The Advance method to set the clock forward works well, and helps with our tests where we want something to expire. 

Note the Delay method doesn’t work as expected where we need to yield to allow another thread to complete. So we continue to use Task.Delay where we need a background thread to run. (It also doesn't seem to play nice with async xunit tests, which seem to spin).


preardon (approved):
Looks good, good to know that Microsoft do this out of the box now

PR #3090, opened 2024-05-15
Archiver Dependencies

iancooper:
The archiver has dependencies on Outbox interfaces, and understands how to archive from them. This does not match the Sweeper, where the Sweeper's role is timed calling of methods on the Command Processor to Clear messages. We want to move the archive functionality into the service bus (which deals with our Outboxes from clients) and have the archive sweeper call that.

The Command Processor tests folder is getting pretty busy, so we will deal with that with folders; we will add tests of the archive functionality under there.


iancooper:
I looked at breaking up External Service Bus as part of this as it has grown large. I abandoned that because it became more complex. The External Service Bus manages the interaction of producer and outbox, which are already separate classes. When you try breaking this up into Outbox and Producer components, you realize that you are just creating an abstraction over that inter-relationship, which doesn't gain anything as it is already abstracted into producer and outbox.

So there still does not seem to be a refactoring route that helps here

preardon (approved):
looks good to me

Issue #3075, opened 2024-05-07
[Bug] The IAmAnOutbox interface is required by the archiver, but not always registered with IoC


iancooper:
### Describe the bug
When the TimedOutboxArchiver hosted service is initialised, it requests a IAmAnOutbox instance from the IoC container. If a DynamoDb outbox is being used via the UseDynamoDbOutbox extension method, then while IAmAnOutboxSync and IAmAnOutboxAsync both extend the IAmAnOutbox interface, IAmAnOutbox is never registered in the IoC container and so the archiver service fails to initialise, crashing the application

### Further technical details
- Brighter version 9.7.8

### Suggested fix
Register the IAmAnOutbox interface in the DynamoDb Outbox registration

### V10 differences

V10 has a lot of improvements around Outbox configuration, but we should confirm as part of this bug fix, that the IAmAnOutbox interface is registered in V10 for archiver scenarios

preardon:
I think this is an Odd one, as IAmAnOutbox technically has 0 methods on it, but I agree we should do something

iancooper:
@preardon The issue comes out of TimedOutboxArchiver taking a dependency on IAmAnOutbox<Message>. We then pass that to OutboxArchiver which queries for both interfaces. The alternative would be to have both of the Archiver classes take both IAmAnOutbox* types as dependencies. This was an issue for us as we did not have support for both sync and async for all of our Outboxes at one point. Not sure if that is still true.

preardon:
@iancooper We could, have Sync and Async TimedOutboxArchivers.

One thing (out of scope) that I could use with this as well so some locking (need it for sweeper more than Archiver) but both would be good

iancooper:
Yeah, came up. For now most folks are using Db locking, but we might want to look at a proper distributed lock

preardon:
I need one at work specifically for the outbox sweeper and archiver let's have a chat about these and the best places to put them on we already have an process lock for sweeping and it might be better to do it at that level instead of the timed service level, but very willing to discuss as I see benefits both Ways

preardon:
Also, I know I didn't say it before but do we want to be registering IAmAnOutbox as we would have to choose which outbox to register against it

iancooper:
> Also, I know I didn't say it before but do we want to be registering IAmAnOutbox as we would have to choose which outbox to register against it

Easy route there is to do what we do now, query to see if you can cast.

iancooper:
OK, first stab at this will go with:

[...]

PR #3092, opened 2024-05-15
Initial suggestion for Global locking for Timed Services

preardon:
Issue #3075

This is an example of providers to allow locking across resources, included is

- In Memory Lock (for when we only want a lock in process)
- Azure Blob Lock


iancooper (approved):
This looks spot on to me

preardon:
We should talk about merging some of the packages, given the size of the archiver and the and locking it could be worth just having an azure package

PR #3093, opened 2024-05-16
Initial suggestion for Global locking for Timed Services for v9

preardon:
issue #3075


iancooper (approved):
With the Azure Storage Lock (showing my ignorance of Azure here) how do we ensure that if you crash without doing a Release, someone else can grab the lock?

preardon:
@iancooper The way this works is using something called a Blob Lease which can be between 15 and 60 seconds, after that the lease expires. so if the process was to crash (or unfortuneatly take over the specified lock time) the lease would expire and would be available for locking again

PR #3096, opened 2024-05-17
Add better OTel information into Brighter

iancooper:
We created semantic conventions for Brighter: https://github.com/BrighterCommand/Brighter/blob/master/docs/adr/0010-brighter-semantic-conventions.md

This PR is about adding those conventions into Brighter.


iancooper:
We will not be able to get links on a publish or bulk clear until the ability to add links to an existing span is added to Activity. This is the relevant issue: https://github.com/dotnet/runtime/issues/97680 and matching PR: https://github.com/dotnet/runtime/pull/101381 so we are just waiting release of this.

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -564,34 +602,59 @@ public string[] DepositPost<TRequest>(IEnumerable<TRequest> requests, RequestCon
...
+            var span = _tracer?.CreateBatchSpan<TRequest>(requestContext?.Span, options: _instrumentationOptions);
             var context = InitRequestContext(span, requestContext);
-
+            
             try
             {
                 var successfullySentMessage = new List<string>();
 
-                foreach (var batch in SplitRequestBatchIntoTypes(requests))
+                foreach (var request in requests)
```
preardon, line 612: This is a significant change in Functionality, Previously we split into types (so that we could pass through the correct request type and the provider factory could correctly identity the request type) and then put them as a bulk operation onto the bus as well as marking them as disposed as a single Bulk Operation.  This will sent them to the bus one by one and then dispatch them one by one

iancooper: So the binding to the type happens now in the call to DepositPost. This means that the DepositPost has the correct request type and can be identified by the mapper factory correctly. LINQ was then used to dispatch. Were we batching up writes to the Outbox? We tend to pay the cost for the pipeline transform call per message.  Batch Clear is a separate issue, as it does not involve the message transform pipeline, so we don't have the same issues.

docs/adr/0011-brighter-support-for-bulk-messaging-operations.md
```diff
@@ -0,0 +1,86 @@
...
+* Feed the resulting set of messages to a bulk version of the `Outbox` via `AddToOutbox`
+
+The main issue with this approach is that the flow for `DepositPost` differs between bulk and single versions, which makes it more complex to modify. 
+
+## Decision
+
+It is worth noting the following:
+
+* The `BulkMessageMap` is a bulk, iterative operation; it's main purpose is late-binding of the type from an `IEnumerable<IRequest>`
+* Producing a `Message` via a `Producer` is not a bulk operation in Brighter
```
preardon, line 55: Brighter v9 does support Bulk clearing via the `IAmABulkMessageProducerAsync` interface

docs/adr/0011-brighter-support-for-bulk-messaging-operations.md
```diff
@@ -0,0 +1,86 @@
...
+
+The only bulk operations we have in the flow are the read and write from the `Outbox`
```
preardon, line 57: In v9 there is also a Bulk Mark Dispatches via the MarkDispatchedAsync(IEnumerable<string> ids, ...) method on the relational outbox

preardon:
@iancooper Talking about bulk I see a couple of Paths here,  Generally I use implicit clearing of the outbox so in my application code I am only ever doing a Bulk outbox add (where necessary). The most important flow in my mind is the Bulk dispatch of messages, at the moment if we consider a batch of 100 messages that are to be dispatched:

Without Bulk Dispatch
- Call CommandProcessor.ClearOutbox
- Get Messages to Dispatch
  - For Each Message Id
    - Open SQL Connection
    - Run Query
    - Close SQL Connection
- For each Message
  - Publish Message
  - Mark As Dispatched
    - Open SQL Connection
    - Run SQL Command
    - Close Connection

In this instance we have
  - 200 SQL Connections Opened and Closed
  - 100 call to the Provider

This is very slow, in my experience dispatching from the outbox is slower than consuming (from memory it was around a message or two a second) so no active messages end up on the queue

The same example above with Bulk (as is currently functioning in v9 for MsSql and ASB)
- Call CommandProcessor.ClearOutbox
- Get Messages to Dispatch
    - Open SQL Connection
    - Run Query
    - Close SQL Connection
- Publish Messages
  - Mark As Dispatched
    - Open SQL Connection
    - Run SQL Command
    - Close Connection
   
In this instance we have
  - 2 SQL Connections Opened and Closed
  - 1 call to the Provider

Using this method I have successfully been able to dispatch a very large weight of messages (>100k) in well under a minute.

Without the ability to bulk dispatch my system will not work (we dispatch millions of messages per day)

docs/adr/0011-brighter-support-for-bulk-messaging-operations.md
iancooper: I'll take a look, although we have not made changes to clear along these lines yet

docs/adr/0011-brighter-support-for-bulk-messaging-operations.md
iancooper: Again, we won't have changed that

iancooper:
> @iancooper The most important flow in my mind is the Bulk dispatch of messages, at the moment if we consider a batch of 100 messages that are to be dispatched:

Not altered here. I'll pick up batches after. It would be good to support both batch writes to the outbox and batch read and write to producer.

There is a V10 change because we need the publication for the pipeline, so that we can use it for CloudEvents (and its there for generic mappers) so we do publication lookup when we run the transform/mapping pipeline.

That means we would need to do something a little different for deposit post as I moved the late binding earlier.

But the clear, I am still looking at

preardon:
> > @iancooper The most important flow in my mind is the Bulk dispatch of messages, at the moment if we consider a batch of 100 messages that are to be dispatched:
> 
> Not altered here. I'll pick up batches after. It would be good to support both batch writes to the outbox and batch read and write to producer.
> 
> There is a V10 change because we need the publication for the pipeline, so that we can use it for CloudEvents (and its there for generic mappers) so we do publication lookup when we run the transform/mapping pipeline.
> 
> That means we would need to do something a little different for deposit post as I moved the late binding earlier.
> 
> But the clear, I am still looking at

Excellent, as long as in the Deposit Flow we can do a single Command for insert this will be excellent! let me know if there are any parts you want me to look at

iancooper:
@preardon As discussed going to merge. It is not complete, but it should work.

* Clear is still be bulk. I have renamed the 'bulk' method to ClearOutstandingFromOutbox for clarity as to purpose
* I'll look into the bulk deposit later
* No error handling in activities yet
* Only InMemory Producer and Inbox/Outbox currently instrumented; will need to pick up the concrete ones

Those will be in further PRs

PR #3158, opened 2024-06-22
Add In-Memory Consumer 

iancooper:
This creates an In-Memory Consumer for use with the bus. Why?

- This allow us to finish the tests around the new tracing by having an in-memory consumer to use
- This allows us to replace the A.Fake<IAmAMessageConsumer> implementations, which better fits our testing style, by using a genuine in-memory class over a mock or fake
- This could be used with an Internal Bus and In-Memory Producer as part of a modular monolith to give asynchronous conversations between a producer and consumer


iancooper:
There are some TODOs around observing the channel being disposed. I did think observability around errors might help. But I also wonder if it implies we want to know the channel state, and mark it is CLOSED or OPEN.

No intent to fix in this PR. Should be discussed and fixed in another

preardon (approved):

PR #3176, opened 2024-06-29
Add Batching into the External Bus

preardon:
Ensure that Messages can be comitted to the outBox as a batch if they are subimtted as one


preardon:
@iancooper 

I did origionally add `BatchAdd` into the `OutboxDbOperation` but then removed it, we could show this as a different operation, but I'm not sure we need it, thoughts?

iancooper (approved):
Looks good.

PR #3178, opened 2024-06-30
Add Sweeper to WebApi_ samples

iancooper:
We have a plethora of samples. Our goal is to keep:

(1) Examples for Send and Publish
(2) A Task Queue example for each transport
(3) WebAPI_ examples for EFCore, Dapper, and DynamoDb (and in theory CosmosDb etc). But each one of these can be configured for any of our transports

As part of getting the otel work done, I need to do archive and so this has triggered me to add the Sweeper from the other samples into WebAPI_ so that we do not need the Sweeper sample. It will also show of our lock mechanism


src/Paramore.Brighter/ApplicationTelemetry.cs
```diff
@@ -23,11 +23,13 @@
 #endregion
 
 
+using System;
 using System.Diagnostics;
 using System.Reflection;
 
 namespace Paramore.Brighter
 {
+    [Obsolete("Use BrighterTracer instead")]
```
preardon, line 32: If this is going into master should we just Remove this class (For v10)

preardon (approved):

src/Paramore.Brighter/ApplicationTelemetry.cs
iancooper: Will happen in a later PR as part of cleanup

preardon (approved):

PR #3182, opened 2024-07-02
[Feature] Parallel Publish

iancooper:
Change the approach on a CommandProcessor Publish to use parallelism: a parallel for on a sync publish and a Task.WhenAll on an async publish.


preardon (approved):
Looks good to me.  As discussed I will need to think about how to Handle DI scopes in the Publish flow as at the minute I believe they share one

iancooper:
> Looks good to me. As discussed I will need to think about how to Handle DI scopes in the Publish flow as at the minute I believe they share one

Indeed. Let's address that in another PR, as I think it needs some thought (and an ADR)

Issue #3177, opened 2024-06-29
[Bug] MySQL Outbox - All pooled connections are in use


romtur:
### Describe the bug
When doing **DepositPostAsync** and then clear messages explicitly by calling **ClearOutboxAsync** the connection pool becomes full very quick. It looks like the connections are not properly disposed. With each call to **ClearOutboxAsync** there is +1 connection in a connection pool.
 
This bug was already fixed in the past: https://github.com/BrighterCommand/Brighter/issues/2077 but probably reintroduced in version 10.

I tried to extract the Outbox code and called 
`connection.Dispose();
` 
after closing the connection but it didn't help.

It seems that in v9 we were able to configure the lifetime of an Outbox, but in v10 it's strictly Singleton.
@iancooper @preardon Please advise if there could be a workaround for this issue? 

### Exceptions (if any)
```
MySqlConnector.MySqlException (0x80004005): Connect Timeout expired. All pooled connections are in use.
   at MySqlConnector.MySqlConnection.CreateSessionAsync(ConnectionPool pool, Int64 startingTimestamp, Activity activity, Nullable`1 ioBehavior, CancellationToken cancellationToken) in /_/src/MySqlConnector/MySqlConnection.cs:line 946
   at MySqlConnector.MySqlConnection.OpenAsync(Nullable`1 ioBehavior, CancellationToken cancellationToken) in /_/src/MySqlConnector/MySqlConnection.cs:line 419
```

### Further technical details
- Brighter version: 1.0.0-preview.3
- The OS: Windows

iancooper:
Thanks @romtur we will take a look and see why it regressed from https://github.com/BrighterCommand/Brighter/issues/2077 @preardon

romtur:
I think I found where the issue is. I was concentrating on the ReadFromStore method, but it happens in the WriteToStore method when we are marking the message as dispatched.

We need to close the connection there in the way it's done in MsSqlOutbox: 
`if (transactionProvider != null)
    transactionProvider.Close();
else
    connection.Close();`

preardon:
I'll have a look, I thought this was done in the relationalOutbox Abstraction

iancooper:
> I'll have a look, I thought this was done in the relationalOutbox Abstraction

It is possible that I broke something when I changed some of this for V10.

preardon:
PR Raised, @iancooper  let me know what you think @romtur  sorry it took me so long to get to this and thanks for the traige

[...]

PR #3185, opened 2024-07-05
Fix MySQL Outbox to close connections if there is no transactions

preardon:
Task #3177


iancooper (approved):
Are we likely to have this on other rdbms boxes?

preardon:
@iancooper I'll double check, I'll also see if I can abstract this into the parent

PR #3186, opened 2024-07-06
Look at Refactoring Outbox Write Actions

preardon:
@iancooper as discussed I wanted to look into further abstracting the Relations Outbox,  If you're happy with the approach I'll make these changes on the rest of them


iancooper (approved):
Yes, this looks better

Issue #3179, opened 2024-07-01
[FEATURE] Service Bus Transport : Refactor for efficencies of setup


preardon:
It appears that everytime the `IServiceBusClientProvider` is used to `GetServiceBusClient` of `GetServiceBusAdministrationClient` a new instance of the clients are newed up.

**Describe the solution you'd like**
I would like this refactored so that the clients are a singleton and use their inbuilt methods to get the child `Get()` properties to create the required clients at a topic level/

iancooper:
@preardon Is this still in progress?

iancooper:
Done, raise new tickets for V11 if needed

PR #3195, opened 2024-07-09
Ensure that we dont recreate Service Bus clients on every get

preardon:
This PR ensures that (inline with recomendation) we dont create a new instance of the Service Bus clients on every call. Also removed some unused interfaces

#3179


preardon:
@iancooper I've started a PR to do this and a full refactor for v10, but this should solve the fact that our clients aren't singletons in v9

iancooper (approved):
Make sense

PR #3217, opened 2024-07-19
Backport Dynamo DB distributed locking provider to v9

dhickie:
This backports the implementation of a Dynamo DB distributed locking provider to to the v9 branch.

In addition to the changes made for the v10 version, this also adds an extension method for registering the Dynamo DB locking provider, in line with the pattern used elsewhere, for example when registering the Dynamo DB outbox.


iancooper (approved):

iancooper:
Thanks @dhickie

PR #3219, opened 2024-07-19
Fixes unlock resource on Postgres 

lillo42:
Fixes unlock resource on Postgres


iancooper:
Thanks @lillo42. As the Postgres tests pass, I will merge.

PR #3243, opened 2024-08-07
Fix v9 Send breaking Open Telemetry

preardon:
When running Send Async and there is an error this can cause a Span the be compelted that should not be and then leaks into the  Activity in the higher context which will cause thigns to be related that are not

### DIFF
--- a/src/Paramore.Brighter/CommandProcessor.cs
+++ b/src/Paramore.Brighter/CommandProcessor.cs
@@ -345,12 +345,14 @@ namespace Paramore.Brighter
                 }
                 catch (Exception)
                 {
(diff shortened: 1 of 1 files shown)


catalintoma:
The reason for this is that disposing an activity also resets Activity.Current to the previous activity (or null for root).
But because Activity.Current is defined as AsyncLocal, it means that when control gets back to MessagePump it is re-assigned.
Since it's stopped by command processor, nothing happens above.

On the next message the pump handles, we expect a null current Activity and it is not. That means that all requests will now be a child of the previous one.

https://medium.com/net-under-the-hood/hidden-workings-of-execution-context-in-net-43b491726c65

iancooper (approved):

PR #3256, opened 2024-08-13
Only configure publisher confirmation callbacks once [v9]

### LINKED ISSUE #3246: [Bug] Infinitely increasing memory usage when publish confirmation is used with a transaction provider
### Describe the bug
By default, the `CommandProcessor` is registered in the IoC container with a `transient` lifetime. Every time it's constructed, it goes through the producers that have been added to the producer registry, and for any that implement the `ISupportPublishConfirmation` interface (currently RMQ and Kafka), it subscribes a delegate to that producer's `OnMessagePublished` event ([link](https://github.com/BrighterCommand/Brighter/blob/bec0523d07c35c53e8b186367665e6be28572ff9/src/Paramore.Brighter/ExternalBusService.cs#L815-L834)).

However, the producer registry is registered to the IoC container as a `singleton`, with no option for customisation. This means that every time a `CommandProcessor` instance is requested from the IoC container (for example when publishing a message), it adds a new subscriber to each producer's `OnMessagePublished` event, resulting in an ever growing list of subscribers. Each of these subscribers, on receipt of a publish confirmation, will then compete over marking the message as dispatched in whatever outbox implementation has been registered.

As the list of event subscribers grows, so too does memory consumption, along with CPU usage for each publish as all of the registered subscribers attempt to mark the message as dispatched.

One way to get around this would be to use the [BrighterOptions](https://github.com/BrighterCommand/Brighter/blob/master/src/Paramore.Brighter.Extensions.DependencyInjection/BrighterOptions.cs) to set the `CommandProcessor` to run as a singleton, and therefore only register the event subscriber delegates once. However, when the `CommandProcessor` is constructed, it also takes an implementation of the `IAmABoxTransactionConnectionProvider` if one has been registered to support transactional writes to the outbox. Any implementations of this interface need to be registered with a `Scoped` lifetime to prevent multiple operations sharing the same transaction incorrectly. As a result, the `CommandProcessor` cannot be registered as a singleton.

### Possible fix

Every instance of the `CommandProcessor` shares a static instance of `ExternalBusServices`, which is where the publish confirmation delegates are actually registered. An easy fix would be to add a `bool` member variable to `ExternalBusServices` which keeps track of whether producers have already had their publish confirmation delegates registered, and only register them if they haven't been already. This would, however, prevent the user from adding new RMQ or Kafka producers to the producer registry at runtime, as their confirmation delegates would never be registered. In order to ensure that each individual producer has had their delegate registered every time a `CommandProcessor` is constructed, some form of producer ID would need to be added to the `IAmAMessageProducer` interface.

### Further technical details
- Brighter version: 9.9.0

dhickie:
This fixes #3246 by ensuring that publish confirmation delegates are only registered once. It does this in a slightly different way than the one that was suggested in the issue, as I noticed that v10 fixes this by moving the callback registration to inside the `ExternalBusServices` constructor. In v9 `ExternalBusServices` doesn't have an explicit constructor, so I've moved the call to `ConfigureCallbacks` to inside the lock where the static instance is initialised in `CommandProcessor`.

### DIFF
--- a/src/Paramore.Brighter/CommandProcessor.cs
+++ b/src/Paramore.Brighter/CommandProcessor.cs
@@ -172,8 +172,6 @@ namespace Paramore.Brighter
             _transformPipelineBuilder = new TransformPipelineBuilder(mapperRegistry, messageTransformerFactory);
 
             InitExtServiceBus(policyRegistry, outBox, outboxTimeout, producerRegistry, outboxBulkChunkSize);
-
-            ConfigureCallbacks(producerRegistry);
         }
 
         /// <summary>
@@ -220,8 +218,6 @@ namespace Paramore.Brighter
             _transformPipelineBuilder = new TransformPipelineBuilder(mapperRegistry, messageTransformerFactory);
 
             InitExtServiceBus(policyRegistry, outBox, outboxTimeout, producerRegistry, outboxBulkChunkSize);
-
-            ConfigureCallbacks(producerRegistry);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


iancooper (approved):
Makes sense to use the one-time initialization of the bus to do a one-time initialization of callbacks.

preardon (approved):

Issue #3242, opened 2024-08-06
[Bug] Using async methods to clear outbox with SNS publishers throws exceptions


dhickie:
### Describe the bug
The [SqsMessageProducer](https://github.com/BrighterCommand/Brighter/blob/master/src/Paramore.Brighter.MessagingGateway.AWSSQS/SqsMessageProducer.cs) only implements the `IAmAMessageProducerSync` interface, meaning it can only publish messages synchronously and has no `async` method.

Now that we can combine producer registries when configuring external buses, this means that if any SNS publications are included, then async methods cannot be used to clear the outbox. If they are, then an `InvalidOperationException` exception is thrown when trying to publish a message to SNS, and no further messages are published.

In practice, this means that:

- `IAmACommandProcessor.ClearOutboxAsync()` cannot be used
- When configuring the outbox, the `UseBulk` option cannot be set to `true`

These limitations are not obvious or straight forward for a user to understand, making them an easy rake to step on.

In this instance, my suggestion would be to implement the `IAmAMessageProducerAsync` interface in `SqsMessageProducer` (in fact, it already uses the `async` method from the AWS SDK, so this is easy to implement), allowing `async` methods to be used to clear the outbox.

### Further technical details
- Brighter version: 9.9.0

iancooper:
This should be fixed

PR #3269, opened 2024-08-20
bug: Support async sns/sqs message publishing v9

jtsalva:
Addresses #3242

### DIFF
--- a/src/Paramore.Brighter.MessagingGateway.AWSSQS/AWSMessagingGateway.cs
+++ b/src/Paramore.Brighter.MessagingGateway.AWSSQS/AWSMessagingGateway.cs
@@ -24,7 +24,7 @@ THE SOFTWARE. */
 
 using System;
 using System.Collections.Generic;
-using System.Threading.Tasks;
+using Amazon.SimpleNotificationService;
 using Amazon.SimpleNotificationService.Model;
 using Microsoft.Extensions.Logging;
 using Paramore.Brighter.Logging;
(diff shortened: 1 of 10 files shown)


src/Paramore.Brighter.MessagingGateway.AWSSQS/AWSMessagingGateway.cs
```diff
@@ -45,16 +45,16 @@ public AWSMessagingGateway(AWSMessagingGatewayConnection awsConnection)
             _awsClientFactory = new AWSClientFactory(awsConnection);
         }
 
-        protected string EnsureTopic(RoutingKey topic, SnsAttributes attributes, TopicFindBy topicFindBy, OnMissingChannel makeTopic)
+        protected async Task<string> EnsureTopic(RoutingKey topic, SnsAttributes attributes, TopicFindBy topicFindBy, OnMissingChannel makeTopic)
```
dhickie, line 48: Since this is a v9 PR, you need to be careful about breaking changes. Changing the signature of a non-private method on a non-private class is a breaking change for consumers updating their packages - they'd now have a bunch of tasks flying around in their code they're not going anything with.

Breaking changes should be reserved for major version upgrades, so v10 would be an OK place to make changes like this. That said, it would need to be part of a broader conversation around the conventions when it comes to supporting async methods. I would suggest we keep it as the addition of the async publish methods, and leave the auxiliary methods as they are now.

src/Paramore.Brighter.MessagingGateway.AWSSQS/SqsMessagePublisher.cs
```diff
@@ -42,7 +43,7 @@ public SqsMessagePublisher(string topicArn, AmazonSimpleNotificationServiceClien
             _client = client;
         }
 
-        public string Publish(Message message)
+        public async Task<string> Publish(Message message)
```
dhickie, line 46: As with the comment above about breaking changes, since this is a public method on a public class, we should be adding a new method for the async version, even if the sync version is just calling `GetAwaiter().GetResult()` on the new method.

tests/Paramore.Brighter.AWS.Tests/MessagingGateway/When_posting_a_message_via_the_messaging_gateway.cs
```diff
@@ -61,7 +61,7 @@ public SqsMessageProducerSendTests()
         public async Task When_posting_a_message_via_the_producer()
         {
             //arrange
-            _messageProducer.Send(_message);
```
dhickie, line 64: We should probably have one version of the test that calls the sync method, and one version that calls the async method.

iancooper (approved):

PR #3270, opened 2024-08-21
feature: Support the Subject property when publishing to SNS

jtsalva:
Use `Message.Header.Subject` as SNS subject

### DIFF
--- a/src/Paramore.Brighter.MessagingGateway.AWSSQS/SqsMessagePublisher.cs
+++ b/src/Paramore.Brighter.MessagingGateway.AWSSQS/SqsMessagePublisher.cs
@@ -23,11 +23,9 @@ THE SOFTWARE. */
 
 using System;
 using System.Collections.Generic;
-using System.Net.Mime;
 using System.Text.Json;
 using Amazon.SimpleNotificationService;
 using Amazon.SimpleNotificationService.Model;
-using Paramore.Brighter.Transforms.Transformers;
 
 namespace Paramore.Brighter.MessagingGateway.AWSSQS
 {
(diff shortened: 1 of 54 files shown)


iancooper (commented):
I have a few questions here.

Typically we compose the body of a message from the properties of the request that is being serialized. We don't tend to add metadata in via the publish process, but via that message mapping pipeline (either the sink or intermediary middleware). 

The publication is accessible via middleware, so it might make sense to add a subject, particularly if we are adding metadata into the body of the message, via that middleware and not via the serialization of the request in the sink. This allows you to not have to remember to add in subject in every mapper, just add the middleware (especially as V10 wants to add default mappers).

As subject is meant for an email header, it would feel that you are doing something specific here i.e. composing a message that will be sent as an email, so adding middleware that supports an SNS subject may be the right approach.

We may need to think about our goal here. I am not sure this is the right approach. I can see why you would opt for it, without knowing about the mapping pipeline however. So I do get that.

jtsalva:
Summary of chat with @iancooper

Ideally, Brighter would avoid this work because the SNS `Subject` is intended for email and is arguably being misused by JustSaying. Nevertheless we agreed on the following:
- `Message.Subject`, currently only used for cloud events will be reused to carry the SNS subject in v10. In v9, the SNS subject will be contained in the message bag. The message producer will pass any provided subject onto SNS and the consumer will make available the SNS subject in `Message`
- Brighter library users are recommended to use tranformers to map the subject from and to `Message` but also have the option of just doing that in the mapper

src/Paramore.Brighter.MessagingGateway.AWSSQS/SqsMessagePublisher.cs
```diff
@@ -45,7 +45,7 @@ public SqsMessagePublisher(string topicArn, AmazonSimpleNotificationServiceClien
         public string Publish(Message message)
         {
             var messageString = message.Body.Value;
-            var publishRequest = new PublishRequest(_topicArn, messageString);
+            var publishRequest = new PublishRequest(_topicArn, messageString, message.Header.Subject);
```
iancooper, line 48: Do we know what happens if the subject is null or empty? We probably need a test for that, as modifying the existing test won't tell us if it is breaking for existing consumers. It just needs a conditional if it breaks anyone.

tests/Paramore.Brighter.AWS.Tests/MessagingGateway/When_posting_a_message_via_the_messaging_gateway.cs
```diff
@@ -23,26 +23,29 @@ public class SqsMessageProducerSendTests : IDisposable
         private readonly string _replyTo;
         private readonly string _contentType;
         private readonly string _topicName;
+        private readonly string _subject;
 
         public SqsMessageProducerSendTests()
         {
-            _myCommand = new MyCommand{Value = "Test"};
+            _myCommand = new MyCommand{Value = "Testttttttt"};
```
iancooper, line 30: As above, we need a test for when there is no subject.

iancooper (changes requested):

iancooper (approved):

iancooper:
@jtsalva merged

PR #3272, opened 2024-08-23
feature: Backport #3270

### LINKED ISSUE #3268: [Feature] Support the `Subject` property on SNS events
Messages published to SNS can have an optional `Subject` parameter specified, which is then populated at the top level. The intent behind the parameter is for populating the subject line when SNS messages are routed to an email endpoint. However, there are other libraries, such as [JustSaying](https://github.com/justeattakeaway/JustSaying), which rely on this parameter as a routing mechanism when consuming messages from SNS.

We would like Brighter to have support for the `Subject` parameter when publishing to SNS so that users routing to email endpoints or using libraries which rely on the parameter can consume messages from SNS that have been published using Brighter.

The [SnsPublication](https://github.com/BrighterCommand/Brighter/blob/master/src/Paramore.Brighter.MessagingGateway.AWSSQS/SnsPublication.cs) class offers a point at which SNS specific configuration can be provided. This class could be extended with an extra optional parameter which takes a `Func<Message,String>` delegate that returns the subject for a given message. 

This delegate can then ultimately be passed to the [SqsMessagePublisher](https://github.com/BrighterCommand/Brighter/blob/master/src/Paramore.Brighter.MessagingGateway.AWSSQS/SqsMessagePublisher.cs), and invoked (if not null) to get the subject of the message and pass it to the SNS `PublishRequest`.

Rather than taking a delegate that determines the subject for a message, the `SnsPublication` could simply take a subject string that applies to all messages for that publication. This would solve the issue for our use case, since each publication corresponds to one topic, and each topic always uses the same subject, but this may not be the case for all users. Users of email endpoints in particular could benefit from more dynamic subject generation depending on the content of the message.

jtsalva:
Use Message.Header.Subject as SNS subject

Backport #3270 

Fixes #3268

### DIFF
--- a/src/Paramore.Brighter.MessagingGateway.AWSSQS/SqsMessagePublisher.cs
+++ b/src/Paramore.Brighter.MessagingGateway.AWSSQS/SqsMessagePublisher.cs
@@ -23,11 +23,9 @@ THE SOFTWARE. */
 
 using System;
 using System.Collections.Generic;
-using System.Net.Mime;
 using System.Text.Json;
 using Amazon.SimpleNotificationService;
 using Amazon.SimpleNotificationService.Model;
-using Paramore.Brighter.Transforms.Transformers;
 
 namespace Paramore.Brighter.MessagingGateway.AWSSQS
 {
@@ -45,7 +43,8 @@ namespace Paramore.Brighter.MessagingGateway.AWSSQS
         public string Publish(Message message)
         {
             var messageString = message.Body.Value;
-            var publishRequest = new PublishRequest(_topicArn, messageString);
+            object subject = GetSubject(message);
+            var publishRequest = new PublishRequest(_topicArn, messageString, subject?.ToString());
 
             var messageAttributes = new Dictionary<string, MessageAttributeValue>();
             messageAttributes.Add(HeaderNames.Id, new MessageAttributeValue{StringValue = Convert.ToString(message.Header.Id), DataType = "String"});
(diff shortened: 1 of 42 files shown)


iancooper (commented):
See comments on the V10 PR. It might be worth getting agreement on the V10 PR before we raise a backport to V9.

src/Paramore.Brighter.MessagingGateway.AWSSQS/SqsMessagePublisher.cs
```diff
@@ -73,5 +72,16 @@ public string Publish(Message message)
 
              return null;
         }
+
+        private static string GetSubject(Message message)
+        {
+            var subjectExists = message.Header.Bag.TryGetValue("Subject", out var subject);
+            if (subjectExists)
+            {
+                message.Header.Bag.Remove("Subject");
```
jtsalva, line 81: I considered not removing this from the bag meaning it would be added to message attributes but that would make the test asserting the subject is produced and consumed correctly meaningless

iancooper (changes requested):
I suspect that there is an missing piece here in V9; we don't have a Subject on the Publication that we can easily map to an SNS Subject. We may need to backport it onto the SNSPublication (the base class version will get picked up in V10) to allow you to set it for the SNSPublication. We can then add it to the bag in a custom transform. We can then pick that up in the SNS Publisher and publish. Otherwise there is no way to set it into the bag. 

in V10 you can drop the transform as we just pick it up directly from the Publication and set it on the header (though we could alter the transform in V10 to enforce this).

It is most likely we need this in V9 for producers only, as we don't use it on consumption in Brighter, so there is no need to alter the consumer in that version. As we consumer the Subject anyway in V10 for Cloud Events, we may as well read it there, although we could just grab it from the header over parsing it from the JSON body.

iancooper (approved):

PR #3273, opened 2024-08-23
Implement a CombinedProducerRegistryFactory for using multiple message transports [v9]

dhickie:
This is a backport of https://github.com/BrighterCommand/Brighter/pull/3271 for v9.


tests/Paramore.Brighter.Core.Tests/TestHelpers/DummyMessageProducer.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System.Collections.Generic;
+
+namespace Paramore.Brighter.Core.Tests.TestHelpers
+{
+    public class DummyMessageProducer : IAmAMessageProducer
```
iancooper, line 5: We are trying to get away from using a dummy in a test, in favor of using a Brighter type that is in-memory; in this case there is an InMemoryProducer, which will need an InMemoryProducerFactory. So if we could add that factory into Paramore.Brighter and then create an InMemoryProducerFactory that would complete the circle here.

iancooper (changes requested):
All good, mixed up V9 and V10 in review

iancooper (approved):

PR #3294, opened 2024-09-04
Enable Nullable in Core and Service Activator

preardon:


preardon:
@iancooper As discussed I think the better example here is the second commit (Service Activator) but let me know what you think.

PR #3295, opened 2024-09-04
Implement a sparse outstanding index for the dynamo db outbox [v9]

dhickie:


iancooper (approved):
Thanks Dom. Makes sense, and thanks for picking this one up.

PR #3320, opened 2024-09-24
Fix Outbox Unit Tests

preardon:
This PR addresses the issues in the Outbox unit tests after moving to DateTimeOffset


iancooper (approved):
I think this is the main fix needed in a few places

PR #3323, opened 2024-09-26
Reduce Enterprise Service Bus Complexity

iancooper:
This PR is about factoring a `MessageDispatcher` and `Outbox` out of `ExternalServiceBus` to reduce its complexity. At this point the goal is not to remove `ExternalServiceBus` but to improve the code health of the project as indicated by CodeScene. In particular this focuses on removing the Bumpy Road warning The PR includes an ADR that describes the change in more detail.


iancooper:
Very much a WiP dump of ideas right now, no need to review yet. If you want to look at anything, the ADR is probably your most valuable item

PR #3346, opened 2024-10-09
Reduce 'sweeper still running' log level from warning to information

jtsalva:
It is a normal case for a sweeper to not obtain a lock, there is no need to draw attention to the log


iancooper (approved):
Thanks @jtsalva

iancooper:
I've squashed. I will check on anything else waiting, but otherwise get a release out asap

PR #3377, opened 2024-11-06
OpenTelemetry and Cloud Events Rollup

iancooper:
This one is about cleaning up some hangovers, leaving us with a clean otel and cloud events implementation and rollout to transports and storage


iancooper:
Drop this one, will pick up in another clean up

PR #3389, opened 2024-11-20
[Docs] Add ADR for generating metrics from traces

jtsalva:
This aims to be a good enough starting point. There are more metrics I think Brighter could provide and some ideas on how, things like `ApproximateAgeOfOldestMessage` - deferring for later discussion.


jtsalva:
@iancooper fyi a first draft

iancooper (approved):
Thanks @jtsalva. This looks great.

holytshirt (approved):
Nicely written up, thank you!

iancooper:
BTW @jtsalva 

We should follow the [Metrics conventions](https://opentelemetry.io/docs/specs/semconv/general/metrics/) for Open Telemetry, In particular, we will want to make sure that we use the [Messaging metric conventions](https://opentelemetry.io/docs/specs/semconv/messaging/messaging-metrics/), for Brighter when it is sending a message or receiving a message. It would be nice to also to use the Db Metric conventions, for our *box implementations.

For the traces, I built against the in-memory bus and *box options first, so that I can then go through and add these to the real implementations; that might work as an approach here too. It lets you break up the PRs for example.

jtsalva:
@iancooper makes sense. I've updated this ADR and raised #3453, both ready for review :)

iancooper:
> @iancooper makes sense. I've updated this ADR and raised #3453, both ready for review :)

Great, I think the ADR is nicely written, so very happy to approve

iancooper (approved):

Issue #3355, opened 2024-10-16
[Feature] Default Message Mapper


preardon:
Functionality should exist to be able to set a Default Message Mapper with the ability to specify overrides for Individual Message Types.

Also Brighter should include a Default Message Mapper with Body Type of JSON which is registered by default

PR #3407, opened 2024-12-02
Add Default Message Mapper

preardon:
First look into Default Message Mapper

#3355

### DIFF
--- a/src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
@@ -379,7 +379,9 @@ namespace Paramore.Brighter.Extensions.DependencyInjection
 
             var messageMapperRegistry = new MessageMapperRegistry(
                 new ServiceProviderMapperFactory(provider),
-                new ServiceProviderMapperFactoryAsync(provider)
+                new ServiceProviderMapperFactoryAsync(provider),
+                serviceCollectionMessageMapperRegistry.DefaultMessageMapper,
+                serviceCollectionMessageMapperRegistry.DefaultMessageMapperAsync
             );
 
             foreach (var messageMapper in serviceCollectionMessageMapperRegistry.Mappers)
@@ -391,6 +393,7 @@ namespace Paramore.Brighter.Extensions.DependencyInjection
             {
                 messageMapperRegistry.RegisterAsync(messageMapper.Key, messageMapper.Value);
             }
+            
 
             return messageMapperRegistry;
         }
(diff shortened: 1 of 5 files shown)


src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionMessageMapperRegistry.cs
```diff
@@ -28,27 +28,23 @@ THE SOFTWARE. */
...
-    public class ServiceCollectionMessageMapperRegistry
+    public class ServiceCollectionMessageMapperRegistry(
+        IServiceCollection serviceCollection,
+        ServiceLifetime lifetime = ServiceLifetime.Singleton)
     {
-        private readonly IServiceCollection _serviceCollection;
-        private readonly ServiceLifetime _lifetime;
-
         public Dictionary<Type, Type> Mappers { get; } = new Dictionary<Type, Type>();
+        public Type DefaultMessageMapper { get; private set; } = typeof(JsonMessageMapper<>);
```
iancooper, line 44: Do we need to default the default here, or do we just do that in the extensions DSLs?

src/Paramore.Brighter/MessageMappers/JsonMessageMapper.cs
```diff
@@ -0,0 +1,47 @@
...
+            IEvent => MessageType.MT_EVENT,
+            _ => throw new ArgumentException(@"This message mapper can only map Commands and Events", nameof(request))
+        };
+
+        if(publication.Topic is null)
+            throw new ArgumentException($"No Topic Defined for {publication}");
+
+        var header = new MessageHeader(messageId: request.Id, topic: publication.Topic, messageType: messageType);
+
+        var body = new MessageBody(JsonSerializer.Serialize(request, JsonSerialisationOptions.Options));
```
iancooper, line 33: In principle there is an async serializer. I am not sure if it actually has meaning. But we should investigate. The MS route here is to pass an enum flag to the common method and then just use sync or async code in the body, based on that flag as needed

iancooper: See the Flag Argument Hack here: https://learn.microsoft.com/en-us/archive/msdn-magazine/2015/july/async-programming-brownfield-async-development

iancooper (approved):
This looks really good. I can add a default Avro as well later

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -379,7 +379,9 @@ public static MessageMapperRegistry MessageMapperRegistry(IServiceProvider provi
 
             var messageMapperRegistry = new MessageMapperRegistry(
                 new ServiceProviderMapperFactory(provider),
-                new ServiceProviderMapperFactoryAsync(provider)
+                new ServiceProviderMapperFactoryAsync(provider),
```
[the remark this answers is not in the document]
preardon: @iancooper  what did we decide here?

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionMessageMapperRegistry.cs
preardon: I think we need something set if we don't want to make this type nullable, I'm open either way

Issue #3386, opened 2024-11-10
Upgrade to V7 of RabbitMQ Client


iancooper:
**Is your feature request related to a problem? Please describe.**
RabbitMQ has released version 7, which has breaking changes. This means it is not a simple dependency upgrade, but a code fix. For this reason, raising a ticket to track it.

**Describe the solution you'd like**
There is a migration guide at: https://github.com/rabbitmq/rabbitmq-dotnet-client/blob/main/v7-MIGRATION.md

**Describe alternatives you've considered**
We could chose to remain on 6.X.X for longer, but as we encapsulate the RMQ dependency for our users, there is no real benefit to doing so

**Additional context**
Possible difficulties include async code and our own event consumer

iancooper:
I suggest that we make this change for V10 only, following completion of the work in #3418

iancooper:
This means overall our strategy is: V9 => V6 RMQ, V10 => V7 RMQ. Once we see the performance implications for RMQ on V10 (mainly significant around expectations for the Reactor as opposed to Proactor pipeline).

PR #3408, opened 2024-12-04
GH-3386 Upgrade RabbitMQ client version to 7

lillo42:


iancooper:
Hi @lillo42 thanks for this. I appreciate the effort on this. It may take a little bit time for me to review it all. 

As part of that review, one thing I want to look at is the risks around doing a wait over async code here, which has a theoretical risk around thread pool starvation. I suspect it won't matter much around start up tasks like exchange and queue declaration, because it is unlikely we will have that kind of load. But I just want to look at where we do that on any tight loop that might starve under load.

I think Stephen Cleary has some tools that help us write some tests that emulate thread pool starvation and determine if we will hit issues (https://github.com/StephenCleary/AsyncEx/tree/master/doc)

We do have our own synchronization context, so it's possible we can use that in the context of any async that risks exhaustion to provide a safer way to wait for async work to complete.

A few things for me to ponder.

iancooper:
Hi @lillo42 

The current plan is to work on getting proper async support for the transport into V10 via #3409. I might do that with V6 and get you then change for V7, or do it with V7 and just take your changes for the sync version. Not quite sure, will keep you posted.

iancooper:
Going to merge this into the work in #3409

iancooper:
@lillo42 I have merged these changes into #3409 alongside the async interfaces

Issue #3309, opened 2024-09-19
[Bug] Some transports block on a delayed requeue where not natively supported


iancooper:
### Describe the bug
Our API offers a delayed requeue where it is possible that the error is transient and could be fixed by simply returning to the queue with a delay. There are three responses we have:

1. Native support for requeue with delay; we used this
2. No native support for requeue and we requeue via a timer event that posts after the elapsed time window
3. We block for the delay period then requeue

We know that 3 is never the right solution as this will block the whole pump, and pause message processing, not requeue with a delay. We should fix those transports that have the incorrect behavior. 

### Further technical details

- The internal message consumer is a good example of how to get this right with a internal Requeue method that is called via a timer if delay is requested, and just passed through if not.

iancooper:
As an example, we don't currently support SendWithDelay on AWS SNS/SQS but we should be able to do this.

I don't think we want to u delay queue, as this delays all items. We would need to set up a specific delayqueue for time required and poll that queue for items to repost. Using the visibility timeout seems better, but this might not be possible if not being used from a Requeue as you have to read a message to then increase its lock period

iancooper:
- Add an abstraction for a scheduled job
- Create an in-memory scheduler that relies on Timers and thus will not persist schedules across runs (i.e works for  in 3 seconds ok, but not 3 days, and not reliable)
- Add external schedulers to implement such as AWS Scheduler, Quartz, Hangfire, Redis, Postgres etc.

iancooper:
Fixed by #3847

PR #3409, opened 2024-12-05
Fix: Sync Over Async Improvements

iancooper:
We have some outstanding issues, to tackle in V10 (and perhaps backport to V9) that concern sync over async. For example see #3309. This has surfaced again through the V7 release of RabbitMQ, which is moving to async. This then raises a problem in PRs like #3408. It is also worth noting others identifying this problem in discussions such as [this one](https://github.com/rabbitmq/rabbitmq-dotnet-client/discussions/1738#discussioncomment-11472088). 

Some goals:

* Move the synchronization context to Brighter to make it available in producer scenarios if needed
* Avoid using .Wait, .Result, or GetAwaiter().GetResult() in hot paths (probably does not matter in configuration at startup)
* Use synchronizationcontext where appropriate to push continuations back onto a single thread to avoid thread pool starvation issues.

This is a set of fixes for some of the most problematic issues that we know of. I suspect this will be a whack-a-mole problem over time. They can be hard to spot in review, and the context for introducing an additional thread matters a lot, so it's not a one-size fits all answer.

We also need to add some tests, using tools from [AsyncEx](https://github.com/StephenCleary/AsyncEx) if needed


iancooper:
Added #3418

iancooper:
This is a complex PR to review, as it makes changes throughout the codebase. It has one main goal: make the Reactor model blocking, and the Proactor model non-blocking, top-to-bottom.

There is an ADR associated with this change, it may be worth reading that ADR to understand what has been done. A summary:

* Make Proactor async throughout, not just for user defined code
* Use a version of Stephen Cleary's AsyncEx over our old synchronizationcontext
* Use the AsyncEx variation to provide sync over async (so not Wait, Result, GetAwaiter().GetResult()
* Wrap the Proactor message pump in that context/scheduler to allow us to terminate async in the thread
* Provide async versions of all existing tests for the transports

Issues will emerge, but it may be better to merge and fix, as this PR is problematic sitting outside too long. Some issues:

* We get some issues with AsyncEx where work is queued back to a stored ExecutionContext, following the SynchronizationContext being disposed. We may be able to use some prior art to solve this. It is possible we could switch back to our old context - but a trial of that proved it handled some scenarios worse that Stephen Cleary's or switch to Visual Studio's SingleThreadedSynchronizationContext. I am inclined to understand and fix.

It is in the ADR but we should document the "native" choice where that exists i.e. Kafka is Reactor and RMQ v7 is Proactor

PR #3427, opened 2024-12-16
[Bug] Fixes issue on resolving on DynamoDb Outbox 

lillo42:
The current implementation to avoid duplication makes an infinity loop to find another `IAmAnOutbox`

Question: Should we allow only one `IAmAnOutbox`, or should I change the implementation to allow a multi-outbox pattern register but only one DynamoDB?


iancooper:
>Question: Should we allow only one IAmAnOutbox
I think we assume only one outbox right now. I don't think we have code that thinks about things like tenancy etc. That would be a bigger piece of work, say V11

Issue #3069, opened 2024-04-29
[FEATURE] Use Source Generated Logging


iancooper:
**Is your feature request related to a problem? Please describe.**
We want to improve the performance of Brighter by reducing unnecessary allocations. Whilst this is a difficult and long process, there is some low-hanging fruit, one of which is source generated logging.

**Describe the solution you'd like**
Replace our existing string templated logs with [source generated logs](https://learn.microsoft.com/en-us/dotnet/core/extensions/logging-library-authors). 

**Describe alternatives you've considered**
An alternative would be to drop a lot of our logging in favor of OpenTelemetry events, but this is mostly seamless via libraries

PR #3429, opened 2024-12-20
GH-3069 use of Logger with source code generator

lillo42:
I'm starting with one class, so we can validate that I'm going with a good path or if we should change it


lillo42:
Do you prefer folks, one big PR to change everything or multi-PRs (one per project)?

iancooper:
@lillo42 Thanks for picking this up. Probably a lot of work. With this kind of change, I tend to suggest that we work as follows:

- Add an ADR to explain what you are doing. Folks who review can read that and sample
- Just target V10. 
- Do a big PR. We can merge and fix as we verify V10. I doubt anyone will successfully trawl every statement that you write.

iancooper:
@lillo42 I am going to close this, as I think that @jtsalva picked up this work in his PR. Reopen if I missed something.

Issue #3321, opened 2024-09-24
Outbox Sweeper effectively ignores the distributed lock due to the sweeping being done on a background thread


preardon:
### Describe the bug
In the timed outbox sweeper we use a Locking provider

``` c#
            var lockId = _distributedLock.ObtainLockAsync(LockingResourceName, CancellationToken.None).Result; 
            if (lockId != null)
            {
                s_logger.LogInformation("Outbox Sweeper looking for unsent messages");

                var scope = _serviceScopeFactory.CreateScope();
                try
                {
                    IAmACommandProcessor commandProcessor = scope.ServiceProvider.GetService<IAmACommandProcessor>();

                    var outBoxSweeper = new OutboxSweeper(
                        millisecondsSinceSent: _options.MinimumMessageAge,
                        commandProcessor: commandProcessor,
                        _options.BatchSize,
                        _options.UseBulk,
                        _options.Args);

                    if (_options.UseBulk)
                        outBoxSweeper.SweepAsyncOutbox();
                    else
                        outBoxSweeper.Sweep();
                }
                catch (Exception e)
                {
                    s_logger.LogError(e, "Error while sweeping the outbox.");
                    throw;
                }
                finally
                {
                    _distributedLock.ReleaseLockAsync(LockingResourceName, lockId, CancellationToken.None).Wait();
                    scope.Dispose();
                }
            }
...

```
however inside of the External bus Service we  do the following

``` c#
if (useAsync)
            {
                if (!HasAsyncOutbox())
                    throw new InvalidOperationException("No async outbox defined.");
                
                Task.Run(() => BackgroundDispatchUsingAsync(amountToClear, minimumAge, useBulk, args), CancellationToken.None);
            }

            else
            {
                if (!HasOutbox())
                    throw new InvalidOperationException("No outbox defined.");
                
                Task.Run(() => BackgroundDispatchUsingSync(amountToClear, minimumAge, args));
            }
...
```

meaning that we release the Lock inside of the Timed Sweeper service without any assurances that the Sweep is finished

preardon:
@iancooper I've got 2 PRs here that both solve this issue, do you have a preference on which one to take forward?

iancooper:
@preardon Did we finish this one?

PR #3438, opened 2024-12-27
Enable a Blocking Sweep

preardon:
Allow the sweeper to run a sweep on the foreground thread so that the locking provider can work

#3321

### DIFF
--- a/Brighter.sln
+++ b/Brighter.sln
@@ -117,7 +117,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "WebAPI_EFCore", "WebAPI_EFC
 		samples\WebAPI\WebAPI_EFCore\tests.http = samples\WebAPI\WebAPI_EFCore\tests.http
 	EndProjectSection
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Paramore.Brighter.Extensions.Hosting", "src\Paramore.Brighter.Extensions.Hosting\Paramore.Brighter.Extensions.Hosting.csproj", "{722B348B-846B-4338-B7F5-DB9C865C3F64}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Paramore.Brighter.Outbox.Hosting", "src\Paramore.Brighter.Outbox.Hosting\Paramore.Brighter.Outbox.Hosting.csproj", "{722B348B-846B-4338-B7F5-DB9C865C3F64}"
 EndProject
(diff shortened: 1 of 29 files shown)


preardon:
@iancooper let me know your thoughts, this loosely follows what we did for v9

iancooper:
I guess that I should ask the dumb question. Why does the locking provider fail on a background thread? It would seem that the a ClearOutboxAsync ought not to block,  by it's name. That either means it is able to call async directly (which might not need a Thread Run, or can't, in which case we should wrap it in a Thread Run?

What am I missing?

preardon:
The problem is that the Locking provider has no visibility to the background thread is it unlocks when this method returns which has no bearing on the background thread at all (as its kicked off on another thread) the Locking provider is pointless using this without a blocking call. The way I see it we have 4 options

1. Throw away the Locking Provider (I don't suggest this as it was brought in to solve the multi Farm Problem)
2. Move the Locking Provider into the core - We can do this but also not necessary
3. The Solution I've put forward here where we can pick if this call blocks or not
4. Make this call always Blocking, the call is only used by the Sweeper (up and down the stack) and given the sweeper now has a locking provider it in I would imagine if we make the change I've made so far then it would never be run in the background.  As an addition to this we could move this out of the Command Processor all together and just pass a `OutboxProducerMediator` to the sweeper, this wouldn't get around us needing to make it run blocking.

I don't hate the idea of the things done in option 4 as it feels a little cleaner, but if we wanted to maintain existing functionality we would need 3.

Let me know your thoughts @iancooper

iancooper:
> The problem is that the Locking provider has no visibility to the background thread is it unlocks when this method returns which has no bearing on the background thread at all (as its kicked off on another thread) the Locking provider is pointless using this without a blocking call. 

Still a little puzzled. Where am I am I missing something?

- I start a TimedSweeper; it starts a timer that runs a sweep at an interval, it runs on the thread pool
- Once signaled, the thread pool thread running sweep tries to obtain a lock; it ought to use a unique name, but I notice TimedOutboxSweeper hardcodes this? Is that right. Should it be passed in, or just a GUID? That way multiple threads all have a unique Id.
- If it is an in-process lock, it is a semaphore slim, which is process wide (has to be for producer-consumer usage); if it is a distributed lock then it is not impacted by thread affinity either, surely. Isn't it always match id to table and release

I think I am still missing a part of this puzzle. Or perhaps it is just that we don't generate a unique sweeper name?

>The way I see it we have 4 options

>  As an addition to this we could move this out of the Command Processor all together and just pass a `OutboxProducerMediator` to the sweeper, this wouldn't get around us needing to make it run blocking.

I don't hate the idea of the Sweeper calling the OutboxProducerMediator directly, regardless of the misunderstanding here, as routing through the CommandProcessor might be waste

iancooper (approved):

PR #3447, opened 2024-12-31
Fixes for Kafka Tests following the Proactor/Reactor split

iancooper:


iancooper:
Likely to be a straight check in from PR as it is fixing broken tests

Issue #3356, opened 2024-10-16
[Feature] Ensure Scope is around Handlers and not Send or Publish call


preardon:
Currently Brighter creates a Scope around the Command Handler, we need this to be around the Request Handler

PR #3454, opened 2025-01-03
Move scopes inside

preardon:
At the minute the Factory Scope (Dependency injection) is around the command processor, this means that when Publish is called multiple Handlers could share the same scope which would be unexpected by the users.

#3356


preardon:
@iancooper let me know if you're happy with this direction

iancooper:
> @iancooper let me know if you're happy with this direction

When I publish, particularly in parallel, I would expect state to be fresh. I think the problem with reusing scope, is that's not true.

I think we need to ask for a fresh scope for each publish chain.

I like where we are, so hopefully we can do that through the lifetime somehow.

iancooper:
The only other alternative is to make them transient within the scope i.e. the publish has the scope, the publish chains are transient within that scope.

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceProviderHandlerFactory.cs
```diff
@@ -43,49 +45,76 @@ public ServiceProviderHandlerFactory(IServiceProvider serviceProvider)
...
         /// <param name="handlerType">The type of handler to request</param>
+        /// <param name="lifetime">The brighter Handler lifetime</param>
         /// <returns>An instantiated request handler</returns>
-        IHandleRequestsAsync IAmAHandlerFactoryAsync.Create(Type handlerType)
+        IHandleRequestsAsync IAmAHandlerFactoryAsync.Create(Type handlerType, IAmALifetime lifetime)
         {
-            return (IHandleRequestsAsync)_serviceProvider.GetService(handlerType);
+            if (_isSingleton)
+                return (IHandleRequestsAsync)_serviceProvider.GetService(handlerType);
+
```
iancooper, line 80: Agreed that is this handles our factory implementation for ServiceCollection, it is a good place to abstract how that works

iancooper (approved):
I think this a solid approach and follows what we would expect the scope to be.

PR #3487, opened 2025-01-20
Add support to scheduler message 

lillo42:
- [x] Remove unnecessary code
- [x] Fixes build
- [x] Add Unit test
- [x] Finish implementation
- [x] Add Sample
- [ ] ~~Implement support to HangFire~~ 
- [x] Implement support for Quartz
- [x] Implement support for AWS Scheduler
- [x] Implement support for Azure


docs/adr/0024-Scheduling.md
```diff
@@ -0,0 +1,89 @@
...
+
+
+## Decision
+
+Giving support to schedule message, it's necessary breaking on `IAmACommandProcessor` by adding these methods:
+
+```c#
+public interface IAmACommandProcessor
+{
+    string SchedulerSend<TRequest>(TimeSpan delay, TRequest request) where TRequest : class, IRequest;
```
iancooper, line 22: I wonder if we would be better off overloading Send with a time, than adding a new name here

src/Paramore.Brighter.MessagingGateway.AWSSQS/SnsMessageProducer.cs
```diff
@@ -173,8 +150,46 @@ public void SendWithDelay(Message message, TimeSpan? delay = null)
...
-    {
-        // SNS doesn't support publish with delay
-        await SendAsync(message, cancellationToken);
+    { 
+        delay ??= TimeSpan.Zero;
+        if (delay != TimeSpan.Zero)
+        {
+            if (Scheduler is IAmAMessageSchedulerAsync async)
+            {
+                await async.ScheduleAsync(message, delay.Value, cancellationToken);
```
iancooper, line 159: Probably best to avoid a keyword in a variable name, use asyncScheduler instead?

src/Paramore.Brighter.MessagingGateway.AWSSQS/SnsMessageProducer.cs
```diff
@@ -173,8 +150,46 @@ public void SendWithDelay(Message message, TimeSpan? delay = null)
...
+                return;
+            }
+     
+            if (Scheduler is IAmAMessageSchedulerSync sync)
+            {
+                sync.Schedule(message, delay.Value);
```
iancooper, line 165: Probably have to rename to strategy used above

src/Paramore.Brighter.MessagingGateway.AWSSQS/SqsMessageProducer.cs
```diff
@@ -131,11 +134,29 @@ public async Task<bool> ConfirmQueueExistsAsync(string? queue = null, Cancellati
...
+        delay ??= TimeSpan.Zero;
+        if (delay > TimeSpan.FromMinutes(15))
+        {
+            if (Scheduler is IAmAMessageSchedulerAsync async)
+            {
+                await async.ScheduleAsync(message, delay.Value, cancellationToken);
+                return;
+            }
+
+            if (Scheduler is IAmAMessageSchedulerSync sync)
```
iancooper, line 151: Same comments here on not blocking on the async path

src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageProducer.cs
```diff
@@ -192,12 +194,62 @@ public void Init()
...
+            delay ??= TimeSpan.Zero;
+            if (delay != TimeSpan.Zero)
+            {
+                if (Scheduler is IAmAMessageSchedulerSync sync)
+                {
+                    sync.Schedule(message, delay.Value);
+                    return;
+                }
+                  
+                if (Scheduler is IAmAMessageSchedulerAsync async)
```
iancooper, line 234: We can default back to sync context, but the alternative is to assume you have the right scheduler. Again avoid async as a name

src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageProducer.cs
```diff
@@ -237,106 +289,99 @@ public void Send(Message message)
...
-                s_logger.LogError(ioe,
-                    "Error sending message to Kafka servers {Servers} because {ErrorMessage} ",
-                    _producerConfig.BootstrapServers,
-                    ioe.Message
-                );
-                throw new ChannelFailureException("Error talking to the broker, see inner exception for details", ioe);
+             delay ??= TimeSpan.Zero;
+             if (delay != TimeSpan.Zero)
+             {
+                 if (Scheduler is IAmAMessageSchedulerAsync async)
```
iancooper, line 319: See other comments on names; match scheduler to path

src/Paramore.Brighter.MessagingGateway.MQTT/MQTTMessageProducer.cs
```diff
@@ -79,8 +83,31 @@ public async Task SendAsync(Message message, CancellationToken cancellationToken
         /// <param name="delay">Delay is not natively supported - don't block with Task.Delay</param>
         public void SendWithDelay(Message message, TimeSpan? delay = null)
         {
+            delay ??= TimeSpan.Zero;
+            if (delay != TimeSpan.Zero)
+            {
+                if (Scheduler is IAmAMessageSchedulerSync sync)
```
iancooper, line 89: As before: names, matching scheduler to path

src/Paramore.Brighter.MessagingGateway.MQTT/MQTTMessageProducer.cs
```diff
@@ -91,10 +118,30 @@ public void SendWithDelay(Message message, TimeSpan? delay = null)
         /// <param name="cancellationToken">Allows cancellation of the Send operation</param>
         public async Task SendWithDelayAsync(Message message, TimeSpan? delay, CancellationToken cancellationToken = default)
         {
-            // delay is not natively supported
-            await SendAsync(message, cancellationToken);
-        }
+            delay ??= TimeSpan.Zero;
+            if (delay != TimeSpan.Zero)
+            {
+                if (Scheduler is IAmAMessageSchedulerAsync async)
```
iancooper, line 124: As before: names, match scheduler to path

src/Paramore.Brighter.MessagingGateway.MsSql/MsSqlMessageProducer.cs
```diff
@@ -114,9 +111,30 @@ public async Task SendAsync(Message message, CancellationToken cancellationToken
...
         /// <param name="delay">The delay to use.</param>
         public void SendWithDelay(Message message, TimeSpan? delay = null)
-        {
-            // No delay support implemented
-            Send(message);
+        { 
+            delay ??= TimeSpan.Zero;
+            if (delay != TimeSpan.Zero)
+            {
+                if (Scheduler is IAmAMessageSchedulerSync sync)
```
iancooper, line 118: As before: names, match scheduler to path

src/Paramore.Brighter.MessagingGateway.MsSql/MsSqlMessageProducer.cs
```diff
@@ -129,8 +147,30 @@ public void SendWithDelay(Message message, TimeSpan? delay = null)
         /// <returns>A task that represents the asynchronous send operation.</returns>
         public async Task SendWithDelayAsync(Message message, TimeSpan? delay, CancellationToken cancellationToken = default)
         {
-            // No delay support implemented
-            await SendAsync(message, cancellationToken);
+            delay ??= TimeSpan.Zero;
+            if (delay != TimeSpan.Zero)
+            {
+                if (Scheduler is IAmAMessageSchedulerAsync async)
```
iancooper, line 153: As before: names, match scheduler to path

src/Paramore.Brighter.MessagingGateway.RMQ/RmqMessageProducer.cs
```diff
@@ -144,13 +149,21 @@ public async Task SendWithDelayAsync(Message message, TimeSpan? delay, Cancellat
 
             _pendingConfirmations.TryAdd(await Channel.GetNextPublishSequenceNumberAsync(cancellationToken), message.Id);
 
-            if (DelaySupported)
+            if (delay == TimeSpan.Zero || DelaySupported)
             {
                 await rmqMessagePublisher.PublishMessageAsync(message, delay.Value, cancellationToken);
             }
+            else if(Scheduler is IAmAMessageSchedulerAsync asyncScheduler)
```
iancooper, line 156: Names good, but match scheduler to path

src/Paramore.Brighter.MessagingGateway.Redis/RedisMessageProducer.cs
```diff
@@ -123,11 +102,45 @@ public void Send(Message message)
...
         {
             if (s_pool is null)
+            {
                 throw new ChannelFailureException("RedisMessageProducer: Connection pool has not been initialized");
+            }
+            
+            delay ??= TimeSpan.Zero;
+            if (delay != TimeSpan.Zero)
+            {
+                if (Scheduler is IAmAMessageSchedulerSync sync)
```
iancooper, line 128: As before: names, match scheduler to path

docs/adr/0024-Scheduling.md
iancooper: Alternatively, how about SendAt, PublishAt, PostAt etc.

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -461,6 +472,166 @@ public async Task PublishAsync<T>(
             }
         }
 
+        /// <inheritdoc />
+        public string SchedulerPost<TRequest>(TRequest request,
```
iancooper, line 476: As above. I suggest we overload Post with a method that takes a DateTimeOffset ```at``` which is the time at which you want to publish. We could have a ```TimeSpan delay``` but I assume with most schedulers that ends up as now + timespan. Perhaps a helper extension for TimeSpan of TimeSpan.FromX.Delay() where delay just adds to now? Either way I think of a scheduler as being an At, but you might want an easy way to mean that as fromNow

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -461,6 +472,166 @@ public async Task PublishAsync<T>(
...
+            s_logger.LogInformation("Scheduling a request: {RequestType} {Id}", request.GetType(), request.Id);
+            
+            var span = _tracer?.CreateSpan(CommandProcessorSpanOperation.Scheduler, request, requestContext?.Span, options: _instrumentationOptions);
+            var context = InitRequestContext(span, requestContext);
+
+            try
+            {
+                Message message = s_mediator!.CreateMessageFromRequest(request, context);
+
+                var scheduler = _messageSchedulerFactory.Create(this);
```
iancooper, line 495: Let's sync up scheduler types to the path type; if you don't have the correct one, we assert. If you need to wrap async over sync or vice-versa, do it in the wrapper for the scheduler we are abstracting

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -461,6 +472,166 @@ public async Task PublishAsync<T>(
...
+                throw;
+            }
+            finally
+            {
+                _tracer?.EndSpan(span);
+            }
+        }
+
+        /// <inheritdoc />
+        public string SchedulerPost<TRequest>(TRequest request,
```
iancooper, line 515: The same comments apply. Overload with an At parameter over a new method convention; only one scheduler type

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -461,6 +472,166 @@ public async Task PublishAsync<T>(
...
+                throw;
+            }
+            finally
+            {
+                _tracer?.EndSpan(span);
+            }
+        }
+        
+        /// <inheritdoc />
+        public async Task<string> SchedulerPostAsync<TRequest>(TRequest request,
```
iancooper, line 554: The same comments apply. Overload with an At parameter over a new method convention; only one scheduler type

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -461,6 +472,166 @@ public async Task PublishAsync<T>(
...
+                _tracer?.AddExceptionToSpan(span, [e]);
+                throw;
+            }
+            finally
+            {
+                _tracer?.EndSpan(span);
+            }
+        }
+
+        public async Task<string> SchedulerPostAsync<TRequest>(TRequest request,
```
iancooper, line 594: The same comments apply. Overload with an At parameter over a new method convention; only one scheduler type

src/Paramore.Brighter/InMemoryMessageScheduler.cs
```diff
@@ -0,0 +1,149 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Paramore.Brighter.Logging;
+using Paramore.Brighter.Scheduler.Events;
+using Paramore.Brighter.Tasks;
+using InvalidOperationException = System.InvalidOperationException;
+
+namespace Paramore.Brighter;
+
+public class InMemoryMessageScheduler(
```
iancooper, line 13: Awesome, thanks for making sure we have one of these

src/Paramore.Brighter/InMemoryMessageScheduler.cs
```diff
@@ -0,0 +1,149 @@
...
+    public async Task CancelAsync(string id, CancellationToken cancellationToken = default)
+    {
+        if (s_timers.TryRemove(id, out var timer))
+        {
+            await timer.DisposeAsync();
+        }
+    }
+
+    /// <inheritdoc cref="Dispose"/>
+    public void Dispose()
```
iancooper, line 114: Do we need to ensure we dispose any remaining timers here?

src/Paramore.Brighter.MessageScheduler.Quartz/QuartzBrighterJob.cs
```diff
@@ -0,0 +1,36 @@
...
+        if (!context.JobDetail.JobDataMap.TryGetBooleanValue("async", out var async))
+        {
+            return;
+        }
+
+        var id = context.JobDetail.Key.Name;
+        var message = JsonSerializer.Deserialize<Message>(obj!, JsonSerialisationOptions.Options)!;
+        if (async)
+        {
+            await processor.PostAsync(new FireSchedulerMessage { Id = id, Message = message });
```
iancooper, line 29: This works if our intent is always to send a message, but on a Send or Publish intent is actually just to delay Send or Publish.

src/Paramore.Brighter.MessagingGateway.Redis/RedisMessageProducer.cs
```diff
@@ -172,10 +171,54 @@ public void SendWithDelay(Message message, TimeSpan? delay = null)
...
-            await SendAsync(message, cancellationToken);
+            if (s_pool is null)
+            {
+                throw new ChannelFailureException("RedisMessageProducer: Connection pool has not been initialized");
+            }
+            
+            delay ??= TimeSpan.Zero;
+            if (delay != TimeSpan.Zero)
+            {
+                if (Scheduler is IAmAMessageSchedulerAsync async)
```
iancooper, line 186: usual comments on name and pick one

src/Paramore.Brighter/CommandProcessor.cs
iancooper: We cover Post here, but we would also want to cover Send and Publish. Those take a Request, not a message. You might find it easier to use a Request to be pushed to the Job. That means that when your QuartzJob or the like calls back, it can just replay the request. Of course, this does not help with messaging providers, who need to support a message, as they have a message by that point. But their callback is just to reproduce their message

iancooper (commented):
First, there is a tremendous amount of work here, that has broken the back of this. So thank you so much. I think my thoughts are generally:

- Let's overload existing methods to take a DateTimeOffset scheduledAt parameter (and provide a helper method to create a scheduledAt from a TimeSpan that represents a delay) over new method names
- We have different job types: Request (from CP goes back to send, post or publish, you need to know which) and Message (from a consumer implementation, used to delay posting a message, comes back to the consumer )
- That makes your scheduler a bit more gnarly; you want an interface that the 3rd party lib support implements that offers either.

But once we have that this will solve a massive problem for us with supporting a Send at a point in time, which will be a really powerful addition in V10

iancooper:
So, let's think separately about how we schedule requests and messages because they have different characteristics:

- Requests - we want to execute code in the same process, so this needs an embedded engine that lets us run our code on the callback
- Message - we can do this in process by just delaying when the message is sent, but it would also work with an out-of-process delay that either delays the external send of our message or triggers a process we run that queues the message.

In the Request case, I think CommandProcessor just needs overloads that take time to schedule this at:

```
public interface IAmACommandProcessor
{
    void Send<TRequest>(TRequest command, RequestContext? requestContext = null) where TRequest : class, IRequest;

void Send<TRequest>(DateTimeOffset at, TRequest command, RequestContext? requestContext = null) where TRequest : class, IRequest;

    Task SendAsync<TRequest>(TRequest command, RequestContext? requestContext = null, bool continueOnCapturedContext = true, CancellationToken cancellationToken = default) where TRequest : class, IRequest;

 Task SendAsync<TRequest>(DateTimeOffset at, TRequest command, RequestContext? requestContext = null, bool continueOnCapturedContext = true, CancellationToken cancellationToken = default) where TRequest : class, IRequest;

    void Publish<TRequest>(TRequest @event, RequestContext? requestContext = null) where TRequest : class, IRequest;

void Publish<TRequest>(DateTimeOffset at, TRequest @event, RequestContext? requestContext = null) where TRequest : class, IRequest;
    
Task PublishAsync<TRequest>(
        TRequest @event, 
        RequestContext? requestContext = null,
        bool continueOnCapturedContext = true, 
        CancellationToken cancellationToken = default
        ) where TRequest : class, IRequest;
    
Task PublishAsync<TRequest>(
        DateTimeOffset at,
        TRequest @event, 
        RequestContext? requestContext = null,
        bool continueOnCapturedContext = true, 
        CancellationToken cancellationToken = default
        ) where TRequest : class, IRequest;

    void Post<TRequest>(TRequest request, RequestContext? requestContext= null, Dictionary<string, object>? args = null) where TRequest : class, IRequest;

void Post<TRequest>(DateTimeOffset at, TRequest request, RequestContext? requestContext= null, Dictionary<string, object>? args = null) where TRequest : class, IRequest;

    Task PostAsync<TRequest>(
        TRequest request, 
        RequestContext? requestContext = null,
        Dictionary<string, object>? args = null,
        bool continueOnCapturedContext = true, 
        CancellationToken cancellationToken = default
    ) where TRequest : class, IRequest;

    Task PostAsync<TRequest>(
        DateTimeOffset at,
        TRequest request, 
        RequestContext? requestContext = null,
        Dictionary<string, object>? args = null,
        bool continueOnCapturedContext = true, 
        CancellationToken cancellationToken = default
    ) where TRequest : class, IRequest;

    //... We don't provide scheduled overrides for deposit and clear. The former occurs as part of a transaction, the latter either immediately or via a sweeper. We assume Post just schedules the calls to Post
}
```

it looks as though DateTimeOffset.ToOffset handles being passed a TimeSpan, which makes x seconds from 'now" easy enough

So this would mean that you end up with two interfaces IAmAMessageScheduler (and variants) and IAmARequestScheduler (and variants).
And IAmARequestSchedulerSync would probably look a little like

```
public interface IAmRequestSchedulerSync : IAmARequestScheduler, IDisposable
{
    string Schedule(Request request, DateTimeOffset at);

    string Schedule(Request request, TimeSpan delay);

    bool ReScheduler(string schedulerId, DateTimeOffset at);

    bool ReScheduler(string schedulerId, TimeSpan delay);

    void Cancel(string id);
}
```

And it would be IAmARequestScheduler that IAmACommandProcessor would use to get its work done, just scheduling the method that does not take the ```DateTimeOffset at``` parameter to run later.

But we still want IAmAMessageScheduler because our transports have messages, and so they need to deal with a message, not a request.

Now Quartz, because it is an embedded scheduler (runs in our process), can use Job just calls the relevant method with stored parameters. So it can implement IAmARequestScheduler and work for IAmACommandProcessor to support this.
But for something like AWS Scheduler, because it's an out-of-process scheduler, we can't easily exercise code, so it's probably not a great use case for making it implement IAmARequestScheduler.

Of course, AWS Scheduler can implement IAmAMessageScheduler if we can forward the message from the external scheduler or listen to an event from the scheduler and forward it ourselves. 

Of course, AWS Scheduler is out of process, so we can implement IAmAMessageScheduler only.  Quartz can implement both.

So we will need to be able to register both if an interface implementation is missing, default to in-memory. I think that in-memory remains useful to provide baseline, non-resumable scheduling.

iancooper:
- Thanks for getting spans in; that really helps!!
- Thanks for adding the overloads with the offset; having both at and delay overloads makes sense
- Generally, all public methods should have XML comments. Feel free to make classes internal, where they are not exposed (for consumption or tests). You can use <see cref=""> to indicate types in params and return values; use remarks if there is something interesting. We may document an implementation if it has details that are class-specific, not interface-specific.
- Let's chat about the design decision to have a FireScheduleRequest being Sent to the CP and then the correct CP method being invoked in the hand. You might be trying to align the message and request call-back paths. I guess the trade-off is between one pattern, and a cheaper request implementation. Might be worth updating our ADR for why.

iancooper (commented):
This looks good, apart from the ArgumentOutOfRangeException

PR #3500, opened 2025-02-07
Fix/issues with running commandprocessor samples

andrewjpoole:
I pulled down the Brighter code to explore and understand how it works, however I found some issues with the CommandProcessor samples, so here's a pr to fix them. 
One was related to the _successor field of RequestHandlerAsync.cs which was being set recursively, causing a stackoverflow exception. I am not sure if my fix is the correct one or in the right place, but at least its a start!

This screenshot illustrates the issue I was seeing
644ae2f6-1357-4070-b2c5-6a4bd6644955


iancooper:
Hi @andrewjpoole thanks. Things are in flux a bit as we push towards V10 release candidate, so the 9x branch is much more stable right now. But thanks for this PR. I'll take a look.

preardon (approved):
Looks good to me

PR #3507, opened 2025-02-12
Improve support to .NET Standard and .NET Framework

lillo42:
Many projects can have support for .NET Standard and we haven't added support to it.

I've created 3 variables:
- BrigtherTargetFrameworks -> This should be the default (netstandard2.0;net8.0;net9.0)
- BrigtherCoreTargetFrameworks -> Some libs have drop support to .NET Standard & .NET Framework like postgres (net462;net8.0;net9.0)
- BrigtherFrameworkAndCoreTargetFrameworks -> For projects that have support to target .NET Core & .NET Framework like MsSQL (net462;net8.0;net9.0)

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -121,6 +121,12 @@
     <PackageVersion Include="Pomelo.EntityFrameworkCore.MySql" Version="9.0.0-preview.1" />
     <PackageVersion Include="Npgsql" Version="9.0.2" />
   </ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net462' ">
+    <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3">
+      <PrivateAssets>all</PrivateAssets>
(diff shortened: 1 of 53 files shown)


iancooper:
@lillo42 Let me know if understand the issue right - build constraints - if so happy to merge

iancooper (approved):

lillo42:
Hey @iancooper 

> Guessing the issue here is that we can't get this to compile for netstandard20, so we need to "fake" the nestandard20 version using the underlying sync apis? Approving on that basis

Depending, on `#if`  for the 99% yes, bute there is one case where a method exists on .NET 8/9 and doesn't exist on .NET Standard on [SqliteOutbox.cs](https://github.com/BrighterCommand/Brighter/pull/3507/files#diff-9ec0280a5f530ea8ac354aaff26cadc2962c6de9e332ddb8db34b2fa3d69f802R452) 

About the [Microsoft.NETFramework.ReferenceAssemblies](https://www.nuget.org/packages/Microsoft.NETFramework.ReferenceAssemblies/) I've added it to allow build .NET Framework on non-Windows machine, it's used only for MS SQL packages

Issue #3112, opened 2024-05-22
[Bug] PostAsync throws an invalidCast exception with DynamoDB outbox


romtur:
### Describe the bug
An InvalidCastException is thrown when calling **PostAsync** with a **DynamoDB** outbox. However, it works correctly when calling DepositPostAsync while providing an IAmADynamoDbTransactionProvider

### To Reproduce
Setup a DynamoDB outbox:
```
brighterBuilder.UseExternalBus((configure) =>
      { 
          configure.ProducerRegistry = new KafkaProducerRegistryFactory(kafkaConfiguration, publications).Create();
          configure.Outbox = new DynamoDbOutbox(dynamoDBClient, new DynamoDbConfiguration());
          configure.TransactionProvider = typeof(DynamoDbUnitOfWork);
          configure.ConnectionProvider = typeof(DynamoDbUnitOfWork);
      });
```

Call **PostAsync** method

The failure occurs in CommandProcessor:
![image](https://github.com/BrighterCommand/Brighter/assets/22519624/7e1119bf-64fd-4492-ba58-c12861c7ba16)

Link to original discussion: https://github.com/BrighterCommand/Brighter/discussions/3091

### Exceptions (if any)
```
System.InvalidCastException: Unable to cast object of type 'Paramore.Brighter.ExternalBusServices2[Paramore.Brighter.Message,Amazon.DynamoDBv2.Model.TransactWriteItemsRequest]' to type 'Paramore.Brighter.ExternalBusServices2[Paramore.Brighter.Message,System.Transactions.CommittableTransaction]'
```

### Further technical details
- Brighter version: 1.0.0-preview.3
- The OS: Windows
- Kafka

romtur:
Just to let you know, this issue also affects the MySqlOutbox

dhickie:
I've spent some time digging into this. When looking at the `DepositPost(Async)` method, there are two overloads - one that accepts an implementation of `IAmABoxTransactionProvider<TTransaction>` as an argument, and one that doesn't. `Post(Async)` meanwhile, only has a single method, which does not accept a transaction provider argument.

When calling either the `DepositPost(Async)` method without a transaction provider argument, or `Post(Async)`, it assumes that `TTransaction` takes the type `CommittableTransaction`, and completely ignores any transaction types which have been registered in the IoC container through `UseExternalBus`. If the user has registered a transaction provider that uses a transaction _other_ than `ComittableTransaction`, then the cast fails and an exception is thrown.

[...]

PR #3525, opened 2025-02-21
[Issue 3112] Use the configured transaction provider when no override is supplied

dhickie:
This PR fixes issue #3112, in which an `InvalidCastException` was thrown when calling `Post` or `PostAsync` on the `CommandProcessor` if a persistent outbox was registered.

It does this by switching to a reflection based approach when calling methods on the `OutboxProducerMediator`, removing the need for the problematic cast. The transaction type for the generic method call is taken from the registered transaction provider during construction, and defaults to `CommittableTransaction` if no transaction provider is registered.

In an effort to simplify usage of the `CommandProcessor`, it also makes a couple of other changes:

- If a call is made to the overload of `DepositPost` or `DepositPostAsync` that _doesn't_ take a transaction provider, it uses the one that was registered along with the persistent outbox (if any).
- If a call is made to the overload of `DepositPost` or `DepositPostAsync` that _does_ take a transaction provider, but the transaction type for the supplied transaction provider doesn't match the type that was registered with the persistent outbox (if any), an `InvalidOperationException` is thrown.


tests/Paramore.Brighter.Core.Tests/CommandProcessors/TestDoubles/SpyOutbox.cs
```diff
@@ -0,0 +1,127 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Paramore.Brighter.Observability;
+
+namespace Paramore.Brighter.Core.Tests.CommandProcessors.TestDoubles
+{
+    internal class SpyOutbox :
```
iancooper, line 10: Could we use the InMemoryOutbox? We are trying to avoid Test Doubles where the in-memory version could substitute. If we need to query the state of the outbox, is that something that is all outboxed,  or should the in-memory outbox be exposed?

tests/Paramore.Brighter.Core.Tests/CommandProcessors/TestDoubles/SpyTransaction.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace Paramore.Brighter.Core.Tests.CommandProcessors.TestDoubles;
+
+public class SpyTransaction
```
iancooper, line 6: Similar to in-memory, does this imply that perhaps we need to swap from CommittableTransacton to our own InMemoryTransaction to use?

tests/Paramore.Brighter.Core.Tests/CommandProcessors/TestDoubles/SpyTransactionProvider.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System.Threading;
+using System.Threading.Tasks;
+
+namespace Paramore.Brighter.Core.Tests.CommandProcessors.TestDoubles;
+
+public class SpyTransactionProvider : IAmABoxTransactionProvider<SpyTransaction>
```
iancooper, line 6: Again, does this imply an InMemoryTransactionProvider?

iancooper (commented):
I would like to see if we can move away from a spy here, and use our InMemoryOutbox and possibly add an InMemoryTransaction and InMemoryTransactionProvider

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -103,7 +104,13 @@ public class CommandProcessor : IAmACommandProcessor
         /// </summary>
         private static IAmAnOutboxProducerMediator? s_mediator;
         private static readonly object s_padlock = new();
-        private static readonly ConcurrentDictionary<string, MethodInfo> s_boundDepositCalls = new(); 
+        private static readonly ConcurrentDictionary<string, MethodInfo> s_boundDepositCalls = new();
+        private static readonly ConcurrentDictionary<string, MethodInfo> s_boundDepositCallsAsync = new();
+        private static readonly ConcurrentDictionary<string, MethodInfo> s_boundBulkDepositCalls = new();
+        private static readonly ConcurrentDictionary<string, MethodInfo> s_boundBulkDepositCallsAsync = new();
+        private static readonly ConcurrentDictionary<string, MethodInfo> s_boundMediatorMethods = new();
```
lillo42, line 111: Since you are changing this, what about using `Delegate` rather than `MethodInfo.Invoke`?

```c#
private static readonly ConcurrentDictionary<string, Func<....>> s_boundDepositCallsAsync = new();

....

var del =method.CreateDelegate<Func<...>>();
s_boundDepositCallsAsync .Add(key, del);

...

var del = s_boundDepositCallsAsync[key];
del(.....);

```

tests/Paramore.Brighter.Core.Tests/CommandProcessors/TestDoubles/SpyOutbox.cs
dhickie: The issue with using the `InMemoryOutbox` is that it's the default implementation, and the scenario being tested here is when the user uses something _other_ than the default, since that was what led to this issue being found in the first place (I also wouldn't have found that we should be clearing the bound `MethodInfo`'s when we clear the outbox mediator from the `CommandProcessor`). 

To do that without a test-only `Spy` implementation we would need to either have an additional type of in memory outbox in the core package (which feels redundant?), or pull in one of the other outbox implementations (and then require all the associated infrastructure in order to be able to run the core tests). This feels like the cleanest solution to me.

src/Paramore.Brighter/CommandProcessor.cs
dhickie: There are a few reasons not to use Delegates here. If you look at how the code would change, you'd be going from this:

```
deposit = depositMethod?.MakeGenericMethod(actualRequestType, transactionType)!;
return (deposit?.Invoke(this, [actualRequest, amABoxTransactionProvider, requestContext, dictionary, batchId]) as string)!;
```

To this:
```
deposit = depositMethod?.MakeGenericMethod(actualRequestType, transactionType)!;
var delegate = deposit.CreateDelegate(
    typeof(Func<IRequest, IAmABoxTransactionProvider, RequestContext, Dictionary<string, object>, string, Type, string>),
    this);
return (delegate.DynamicInvoke(actualRequest, amABoxTransactionProvider, requestContext, dictionary, batchId) as string)!;
```

By storing the methods as `MethodInfo`'s, we can also store multiple methods with different signatures in the same collection, as we do with the the bound mediator methods in `s_boundMediatorMethods`. If they were delegates, each mediator method would need a separate member variable. So overall, using delegates makes the code more verbose without any real benefit in terms of clarity.

Lastly, using `Delegate.DynamicInvoke` is actually much slower than `MethodInfo.Invoke`, and given we're at the heart of things here we want it to be as performant as possible.

dhickie: I should also say that `MethodInfo.CreateDelegate<T>` doesn't exist in `netstandard2.0`, which we still support at the moment. Hence the need for `Delegate.DynamicInvoke`.

tests/Paramore.Brighter.Core.Tests/CommandProcessors/TestDoubles/SpyOutbox.cs
iancooper: Ah yes, that makes sense in this context. Closed.

iancooper:
I think we could merge this in if you want @dhickie?

PR #3530, opened 2025-02-26
Throw an exception if the user has configured an outbox with no transaction provider of a matching type

dhickie:
Currently, when the user registers an Outbox implementation in the call to `UseExternalBus`, if they don't also provide an implementation of `IAmABoxTransactionProvider` for a transaction type that matches the type used by the Outbox implementation, then no outbox will actually be registered in the IoC container at all. However, the method call will still succeed, and it isn't until the user tries to post a message that they will receive an error message stating that no outbox has been registered.

This PR changes `UseExternalBus` so that if the user hasn't provided an implementation of `IAmABoxTransactionProvider` for the same transaction type as the provided Outbox, the method will throw a `ConfigurationException`, letting the user know as early as possible that they need to fix their configuration.

### DIFF
--- a/src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
+++ b/src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
@@ -1,4 +1,4 @@
-#region Licence
+﻿#region Licence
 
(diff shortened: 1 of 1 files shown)


iancooper (approved):

iancooper:
Thanks Dom

PR #3549, opened 2025-03-03
Reduce dictionary lookups

Henr1k80:


iancooper (approved):
Thanks for this. I suspect there are other places where we have older code that does not pick up on later BCL improvements.

iancooper:
Merged. We would welcome other peformance improvement PRs

PR #3550, opened 2025-03-04
Avoid some string & List allocations

Henr1k80:
- Avoid allocating a lowercased string when comparing ignoring case
- Avoid some unneeded List allocations


iancooper:
Thanks @Henr1k80 I appreciate this

iancooper (approved):

iancooper:
At some point we created this scenario, where we hire for experience with the buzzword, so everyone wants experience of the buzzword, even if its not applicable to their situation. The number of times I have stumbled into CV-ware....

PR #3552, opened 2025-03-04
Making internal & private classes sealed to avoid virtual dispatch, plus some Dictionary capacity tweaks

Henr1k80:
As the good Stephen Toub [explains, there is good reasons to make classes sealed.](https://github.com/dotnet/runtime/issues/49944)

I can also make it a warning, if all new code does not obey this, by enabling [CA1852](https://learn.microsoft.com/en-gb/dotnet/fundamentals/code-analysis/quality-rules/ca1852)  
Then I would also have to mark all tests & examples sealed, no problem with me, but a LOT of diff lines.  
Alternatively make a config for tests that explicitly ignore the rule.

### DIFF
--- a/src/Paramore.Brighter.DynamoDb/DynamoDbTableBuilder.cs
+++ b/src/Paramore.Brighter.DynamoDb/DynamoDbTableBuilder.cs
@@ -114,7 +114,7 @@ namespace Paramore.Brighter.DynamoDb
 
             return tableCheck.Any(kv => kv.Value) ? 
                 (true, tableCheck.Where(tbl => tbl.Value).Select(tbl => tbl.Key)) : 
-                (false, Enumerable.Empty<string>());
+                (false, []);
 
         }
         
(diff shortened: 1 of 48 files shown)


iancooper (approved):

iancooper:
Hi @Henr1k80. It is a good point. I guess there is no real muscle memory for us around making something sealed. Probably in case we feel the need to extend it later.

iancooper:
>A MSSQL unit test is failing, can it be a timing issue?
As we push towards V10, a few fragile tests will occasionally fail. We will probably pick them up if they keep failing and see if we can rewrite, or just mark them as fragile so they don't run in CI.

But expect a little bit of instability as we push toward our next RC.

iancooper:
>I can also make it a warning, if all new code does not obey this, by enabling [CA1852](https://learn.microsoft.com/en-gb/dotnet/fundamentals/code-analysis/quality-rules/ca1852)

We should probably do this, as I think we will unlikely remember without a prompt for a while. Do you want to do that?

PR #3553, opened 2025-03-05
Add OTel tracing to the Dynamo DB outbox

dhickie:
This PR adds OTel tracing to the Dynamo DB outbox, following the same conventions used in the `InMemoryOutbox`.

In addition, it makes use of the `db.operation.parameter.message.id` and `db.response.returned_rows` tags to track the IDs of messages being operated on by the outbox and the number of rows being returned by queries for outstanding and delivered messages.


iancooper (approved):
This looks right to me.

iancooper:
@dhickie It looks good, but the dynamo tests are failing. Can you check for me

PR #3554, opened 2025-03-05
Set CA1852 as warning & fix these warnings

Henr1k80:


iancooper (approved):
Thanks. This is really helpful

PR #3557, opened 2025-03-07
Add OTel to remaining outbox implementations

dhickie:
This PR adds OTel tracing to all Outbox implementations. For relational database implementations, it leverages the `RelationDatabaseOutbox` base class, which all implementations automatically inherit.

It also expands on the existing tracing implementation for MongoDb, to ensure consistency in the span attributes used by the other outbox implementations.


src/Paramore.Brighter.Outbox.DynamoDB/DynamoDbOutbox.cs
```diff
@@ -400,13 +471,10 @@ public async Task MarkDispatchedAsync(
...
         {
-            var message = _context.LoadAsync<MessageItem>(id, _dynamoOverwriteTableConfig).Result;
-            MarkMessageDispatched(dispatchedAt ?? _timeProvider.GetUtcNow(), message);
-
-            _context.SaveAsync(
-                message, 
-                _dynamoOverwriteTableConfig)
-                .Wait(_configuration.Timeout);
+            MarkDispatchedAsync(id, requestContext, dispatchedAt, args)
+                .ConfigureAwait(ContinueOnCapturedContext)
```
iancooper, line 475: I wonder if we should use Brighter's AsyncContext and not GetResult()

src/Paramore.Brighter/RelationDatabaseOutbox.cs
```diff
@@ -46,13 +49,29 @@ public void Add(
...
+            var dbAttributes = new Dictionary<string, string>()
             {
-                logger.LogWarning(
-                    "MsSqlOutbox: A duplicate Message with the MessageId {Id} was inserted into the Outbox, ignoring and continuing",
-                    message.Id);
-            });
+                {"db.operation.parameter.message.id", message.Id}
+            };
+            var span = Tracer?.CreateDbSpan(
+                new OutboxSpanInfo(dbSystem, databaseName, OutboxDbOperation.Add, outboxTableName),
```
iancooper, line 57: The dbAttributes need to be set on the OutboxSpanInfo. It doesn't look like they are.

iancooper (commented):
@dhickie Thanks for doing this. I think there are some more value on the OutboxSpanInfo that we can set, particularly for an RDBMS, but we even have them for a DynamoDb call, such as https://opentelemetry.io/docs/specs/semconv/database/dynamodb/

You might need to extend OutboxSpanInfo to pick those up, as it probably does not have all the possible properties.

I think this is useful. Understanding the query issues can be really helpful for debugging.

When we do the work to actually make choices in the BrighterTracer based on InstrumentationOptions we can give folks an opt-out from noise reporting.

src/Paramore.Brighter.Outbox.DynamoDB/DynamoDbOutbox.cs
dhickie: We use `GetAwaiter().GetResult()` in a lot of places throughout the codebase - if that's not the way we should be doing things then it should probably be in a separate PR that does it everywhere.

dhickie:
> @dhickie Thanks for doing this. I think there are some more value on the OutboxSpanInfo that we can set, particularly for an RDBMS, but we even have them for a DynamoDb call, such as https://opentelemetry.io/docs/specs/semconv/database/dynamodb/

I did have a look at the attributes for Dynamo DB calls, but aren't these being set on the spans created by the AWS SDK?

I'll have a look at adding the operation name and query text to the span attributes.

src/Paramore.Brighter.Outbox.DynamoDB/DynamoDbOutbox.cs
iancooper: I did use a PR to try and fix many of them, as part of Proactor/Reactor work. I may well not have found all of them. Let's drop for your issue, and I will pick this up along with any others that have remained or crept back in

iancooper (approved):

iancooper:
Thanks @dhickie going to merge this. I will pick up the ConfigureAwait issue separately.

PR #3562, opened 2025-03-13
Add OTel to the InMemoryInbox

dhickie:
This PR adds Open Telemetry tracing to the `InMemoryInbox`, by reusing the `OutboxSpanInfo` and `OutboxDbOperation` class and enum, with appropriate renaming and updates to support the `Exists` operation. It also tidies up XML comments across all Inbox implementations.


iancooper (approved):
There is a lot here, but this looks good.

iancooper:
We just need to check the Sqlite test is not failing because of this change

dhickie:
> We just need to check the Sqlite test is not failing because of this change

@iancooper This was actually broken because of the previous Outbox OTel PR - the constructor signature on the relational database outbox had changed so it wasn't setting the name of the outbox table correctly in the test. I've fixed it here.

iancooper:
Thanks. Will merge

PR #3579, opened 2025-04-04
feature: Use source generated logging

jtsalva:
Addresses #3069

## Goal
We want to use source generated logging instead of string templated logging because "it offers superior performance and stronger typing" as documented [by Microsoft](https://learn.microsoft.com/en-us/dotnet/core/extensions/logging-library-authors).
## Challenge
Although the change is simple there are a large number of logs in the codebase, it's a tedious job to do manually.
#### Approach
- Limit the scope of this PR to perform a direct 1 to 1 mapping from string templated logging to source generated logging:
- Leverage LLMs to do the heavy lifting
- Manually touch up any gaps left by the LLM
#### Pattern
- Source generated logging is always used in the `src` folder
- Any class needing to write logs will have a private static class `Log` as a direct child to colocate logging methods attributed with `[LoggerMessage]` with where they're used
- EventId is omitted - at least for now, as it's not specified in existing code
#### Conversion Rules
- Log levels MUST remain the same
- String templates SHOULD remain the same unless change is required or super easy:
	- Replace all indexed template vars with named vars [SYSLIB1014](https://learn.microsoft.com/en-gb/dotnet/fundamentals/syslib-diagnostics/syslib1014)
	- Remove exception template vars [SYSLIB1013](https://learn.microsoft.com/en-gb/dotnet/fundamentals/syslib-diagnostics/syslib1013)
	- Convert camelCase template vars to PascalCase
#### Example Conversion
**String Templated Logging** (before)
```c#
namespace Brighter;

class SomeClass
{
    static readonly ILogger s_logger = ...;
  
    void Hello(name)
    {
        s_logger.LogInformation("Hello {Name}", name);
    }
}
```

**Source Generated Logging** (after)
```c#
namespace Brighter;

partial class SomeClass
{
    static readonly ILogger s_logger = ...;
  
    void Hello(name)
    {
        Log.Hello(s_logger, name);
    }
  
    static partial class Log
    {
        [LoggerMessage(LogLevel.Information, "Hello {Name}")]
        public static partial void Hello(ILogger logger, string name)
    }
}
```
### Implementation
With not much thought I chose Google's `gemini-2.5-pro-exp-03-25` model with a 1M context window, this can take as input the whole of Brighter's src folder. Unfortunately, Google's servers just kept cutting out when trying to return the transformed output even though the limit is 65536 tokens, the problem was especially noticeable with large files like `CommandProcessor.cs`.  The best we can do is instruct Gemini to make changes one file at a time.

With the help of some tooling I passed all files in the src folder that contain `s_logger` to Gemini one by one with the following prompt:
```
Convert all string templated logging to source generated logging. The new source gen class should be private and appended as a child to the class (at the bottom) that uses it. The new source gen class should be in the form "private static partial class Log" and the logging methods attributed with [LoggerMessage]. Existing logging calls should invoke the new source gen class like Log.TheLoggingMethod(s_logger, args). Make sure not to make any other changes beyond this directive. Definitely do not delete any existing methods. Do not populate the EventId field. Prefer the attribute to be on a single line. The message template should contain variables (in PascalCase) corresponding to the method parameters (in camelCase). Prefer not explicitly naming the parameters in the attribute method invocation params - so instead of `[LoggerMessage(Level = LogLevel.Error, Message = "Message {MessageId} mismatch. Message type is '{MessageType}' yet mapper produced message of type IEvent")]` use `[LoggerMessage(LogLevel.Error, "Message {MessageId} mismatch. Message type is '{MessageType}' yet mapper produced message of type IEvent")]`. Remember to add the `partial` keyword to the parent class.
```
#### LLM Supported Review
On top of manual reviews, I found LLMs useful at review stage too. Feeding file scoped diffs gives the LLM the smallest most concise text (so that it can make the least mistakes) alongside the conversion instructions. The LLM is instructed to pair each deleted string templated log to the new source generated log and find any issues. Here is an extract from a review where I asked Gemini to highlight any non-exact string template mappings:
```
=== Review for src/Paramore.Brighter/CommandProcessor.cs ===

Issues found:
- Mismatch in template for finding async pipelines. The original template used positional arguments and referred to 'event', while the new template uses named arguments and refers to 'command'.
    - Original: `"Found {0} async pipelines for event: {EventType} {Id}"`
    - New: `"Found {HandlerCount} async pipelines for command: {Type} {Id}"`
- Mismatch in whitespace for sending request with routingkey. The original template had an extra space after "request".
    - Original: `"Sending request  with routingkey {ChannelName}"`
    - New: `"Sending request with routingkey {ChannelName}"`

=== Review for src/Paramore.Brighter.MessagingGateway.RMQ.Async/RmqMessageProducer.cs ===

Issues found:
- "Extra space removed in preparing to send message template"
    - Original: `"RmqMessageProducer: Preparing  to send message via exchange {ExchangeName}"`
    - New: `"RmqMessageProducer: Preparing to send message via exchange {ExchangeName}"`
```
#### Local Testing
Source gen logs are visible when ran locally:
Screenshot 2025-04-04 at 23 39 49


iancooper (approved):

iancooper:
@jtsalva It looks good, I just need to get a clean build, and then I will merge

PR #3584, opened 2025-04-11
Migrate SLN to SLNX

lillo42:
Migrate the `sln` to `slnx`, it'll reduce the `sln` file conflict (I hope so) but I'll required developers to use .NET 9.0.200 and last version of Rider & Visual Studio (I guess)


iancooper:
So, let's pause this one. While the slnx file is a better format, I would want us to take stock of its impact on the IDEs folks use, and we can't move to a net9-alone dependency. We have customers on other formats.

iancooper:
Going to close. Let's discuss

Issue #3585, opened 2025-04-11
Defer Exception causes ASB to Attempt a Reject then Complete with the same Lock Token


CCBPMDev:
### Describe the bug

I am receiving a standard "The lock supplied is invalid" exception from your underlying ASB connection. From looking at your code and checking the logs (below), it seems to Reject the message with the lock token to handle the DeferMessageAction() case in the MessagePump.cs class. Then due to the return false, it then attempts the Acknowledge with the same message and same key. Logs show the Lock Token being used 2 times, the 2nd appears to generate the (predicable) exception. 

Azure.Messaging.ServiceBus.ServiceBusException: The lock supplied is invalid. Either the lock expired, or the message has already been removed from the queue. For more information please see https://aka.ms/ServiceBusExceptions .

I just want any failures in the handler to try x times then get Rejected to DLQ. I understand I need to throw this type of exception but experiencing this Lock issue.  Being able to Reject the message in the handler but the MessagePump doesnt support this.

### To Reproduce

Use 9.9.8 Library
External Bus using ASB with default Subscribers / Topic
    //Consumers
    
    var subscriptions = new Subscription[]
    {
      new AzureServiceBusSubscription<RegisterUserCommand>(
        new SubscriptionName(options.ServiceBusRegisterUserTopicName), // Brighter Internal Ref
        new ChannelName(options.ServiceBusRegisterUserSubscriptionName), //Service Bus - Subscription Name
        new RoutingKey(options.ServiceBusRegisterUserTopicName), //Service Bus - Topic Name
        makeChannels: OnMissingChannel.Create,
        requeueCount: 2,
        timeoutInMilliseconds: 60000,
        isAsync: true
      )
    };

    var clientProvider = new ServiceBusConnectionStringClientProvider
    (
      options.ServiceBusConnectionString
    );

    services
      .AddServiceActivator(cfg =>
      {
        cfg.Subscriptions = subscriptions;
        cfg.ChannelFactory = new AzureServiceBusChannelFactory
        (
          new AzureServiceBusConsumerFactory
          (
            clientProvider
          )
        );
        cfg.UseScoped = true;
        cfg.HandlerLifetime = ServiceLifetime.Transient;
        cfg.MapperLifetime = ServiceLifetime.Transient;
        cfg.CommandProcessorLifetime = ServiceLifetime.Singleton;
      })
      .AutoFromAssemblies();

    services
      .AddHostedService<ServiceActivatorHostedService>(); //CONSUMER

[...]

PR #3596, opened 2025-04-16
Ensure that we dont attempt to ACK if we successfully Reject

preardon:
This PR gives the ability for the Consumers to signal if they have successfully Reject a message so that we do not attempt to ACK if the message is no longer on the Channel

Issue #3585

### DIFF
--- a/src/Paramore.Brighter.MessagingGateway.AWSSQS/SqsMessageConsumer.cs
+++ b/src/Paramore.Brighter.MessagingGateway.AWSSQS/SqsMessageConsumer.cs
@@ -167,10 +167,10 @@ namespace Paramore.Brighter.MessagingGateway.AWSSQS
         /// Rejects the specified message.
         /// </summary>
         /// <param name="message">The message.</param>
-        public void Reject(Message message)
+        public bool Reject(Message message)
         {
             if (!message.Header.Bag.ContainsKey("ReceiptHandle"))
-                return;
+                return false;
 
             var receiptHandle = message.Header.Bag["ReceiptHandle"].ToString();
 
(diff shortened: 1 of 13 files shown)


iancooper:
@preardon, It looks like the tests need some fixes. It seems like an obvious break for the changed behavior.

preardon:
@iancooper  after a re-run most went away and the only remaining test failure appears to be unrelated, I'll dig into why the unrelated one is happening

iancooper:
> @iancooper after a re-run most went away and the only remaining test failure appears to be unrelated, I'll dig into why the unrelated one is happening

The RMQ one seems related: 

[xUnit.net 00:01:04.88]     Paramore.Brighter.RMQ.Tests.MessagingGateway.RmqMessageProducerDelayedMessageTests.When_reading_a_delayed_message_via_the_messaging_gateway [FAIL]
[xUnit.net 00:01:04.88]       Expected delayedResult.Header.DelayedMilliseconds to be 3000, but found 0 (difference of -3000).

AWS credentials works on V10, not on V9, which is super-annoying.
The MQTT one might be an accidental upgrade of the package?

preardon:
this failure `[xUnit.net 00:01:04.88] Expected delayedResult.Header.DelayedMilliseconds to be 3000, but found 0 (difference of -3000).` is an Odd one

The problem is that when the value is placed on the queue it is (int32)3000 but when its removed off the queue its (Int64)-3000

its being return as zero as there is a case statement that looks for type int and defaults to 0 if not found.

@iancooper  how would you like to proceed

iancooper:
> The problem is that when the value is placed on the queue it is (int32)3000 but when its removed off the queue its (Int64)-3000

I guess we put it on and off as int64?

Issue #3597, opened 2025-04-20
CorrelationId in Reply class is not settable


SteveBush:
### Describe the bug
I'm struggling with the purpose of the CorrelationId in the Reply class. It's not assigned in the constructor, and it's a Guid type rather than a string like the CorrelationId used in the ReplyAddress class. Two proposals for fixing:

1. Change the type of CorrelationId on Reply to string, add a setter, and set the value to CorrelationId from the passed-in sender address.   

2.  Change the type of CorrelationId on Reply to string, and the CorrelationId could refer to this.SendersAddress.CorrelationId.  

I'm happy to make whatever change is preferred.

### To Reproduce
```
var myReply = new Reply(new ReplyAddress("mytopic", Guid.NewGuid().ToString()); 

// my.Reply.CorrelationId is always an empty Guid.
```

SteveBush:
See PR #3602.  After looking at the usages of CorrelationId on Reply, I decided to keep the type as a Guid to match the Request interface.  I added code to parse the SenderAddress string correlationId into a Guid.  I added tests to validate the CorrelationId on the Reply matches the senderAddress.

SteveBush:
PR #3602 merged into master branch

PR #3602, opened 2025-04-20
Update CorrelationId  from SenderAddress.

SteveBush:
See issue #3597 

- Added a setter to Reply CorrelationId
- Set CorrelationId in Reply from SenderAddress
- Add tests


iancooper (approved):
Thanks @SteveBush

iancooper:
Thanks @SteveBush I think this is fine. We appear to have the AWS tests misfiring again (thanks GH) and RMQ is just misfiring

PR #3604, opened 2025-04-23
Add support to Cloud Events in JSON format

lillo42:
- Add a new enum for Cloud Event attributes, so now we can select which format use with the attribute
 - Add a Cloud Event message mapper (it should be fast than use it via attribute)
 - Allow use the Cloud Event message mapper as default mapper
 - Add a new property to `Publication` called `CloudEventsAdditionalProperties` that allow set extra property to Cloud Event in Json format


src/Paramore.Brighter/Transforms/Transformers/CloudEventsTransformer.cs
```diff
@@ -84,9 +97,160 @@ public Message Wrap(Message message, Publication publication)
...
+        {
+            Id = message.Id,
+            SpecVersion = message.Header.SpecVersion,
+            Source = message.Header.Source,
+            Type = message.Header.Type,
+            DataContentType = message.Header.ContentType,
+            DataSchema = message.Header.DataSchema,
+            Subject = message.Header.Subject,
+            Time = message.Header.TimeStamp,
+            Data = JsonSerializer.Deserialize<JsonElement>(message.Body.Value)
```
iancooper, line 116: The deserialize here could throw. We shoud catch the exception, log, and return the untransformed original message

src/Paramore.Brighter/Transforms/Transformers/CloudEventsTransformer.cs
```diff
@@ -84,9 +97,160 @@ public Message Wrap(Message message, Publication publication)
...
+        }
+
+        var cloudEvents = JsonSerializer.Deserialize<CloudEventMessage>(message.Body.Value, JsonSerialisationOptions.Options);
+
+        if (cloudEvents == null)
+        {
+            return message;
+        }
+
+        var bag = cloudEvents.AdditionalProperties ?? [];
```
iancooper, line 140: What do we do if names collide? i.e., can you overwrite an existing header? Document if we think you should - this is an override - that is the behaviour, otherwise check not contains

lillo42: `System.Text.Json` will override the value, I'll document it

iancooper:
@lillo42 Thanks for the properties fix. Don't forget the potential serialization exceptions and the fix for that

PR #3605, opened 2025-04-24
OTel Transports

iancooper:
We need to propogate our context when sending, and retrieve our context when recieving. In addition, we need to add a span to the producer. 

We don't need a span on the receiver, because the message pump will create one, but we do need to attach the relevant context from the message to the span.


src/Paramore.Brighter/Id.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// </summary>
+        public string Value { get; }
+
+        /// <summary>
+        /// Gets an empty Id instance.
+        /// </summary>
+        /// <remarks>
+        ///  Should be used to indicate a lack of value or an uninitialized state.
+        /// </remarks>
+        public static Id? Empty { get; } = new(string.Empty);
```
lillo42, line 26:
```suggestion
        public static Id Empty { get; } = new(string.Empty);
```

PR #3609, opened 2025-04-30
Add string conversion from ChannelName & RoutingKey

lillo42:
It'll improve migration to Brighter V10


iancooper (approved):
Nice addition

PR #3618, opened 2025-06-02
feat: Update Kafka publication default values

lillo42:
I'm setting the same value as here: https://github.com/confluentinc/librdkafka/blob/master/CONFIGURATION.md?plain=1#L142C75-L142C82


taharafique (approved):

iancooper:
@taharafique We should probably also make these changes in the Release/9x branch

iancooper (approved):

PR #3620, opened 2025-06-05
Enforcing the InstrumentationOption on all Tracer Operations

preardon:
enforcing the InstrumentationOption on all Tracer Operations

### DIFF
--- a/samples/CommandProcessor/HelloWorldInternalBus/Program.cs
+++ b/samples/CommandProcessor/HelloWorldInternalBus/Program.cs
@@ -3,6 +3,7 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Hosting;
 using Paramore.Brighter;
 using Paramore.Brighter.Extensions.DependencyInjection;
+using Paramore.Brighter.Observability;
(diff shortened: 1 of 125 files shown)


preardon:
@iancooper can you please give me your initial thoughts on this approach, the biggest questions I have are around what we should be filtering with each Flag, should we be creating more flags?

iancooper:
> @iancooper can you please give me your initial thoughts on this approach, the biggest questions I have are around what we should be filtering with each Flag, should we be creating more flags?

I think the approach here is sensible. I would suspect we need "just enough" flags. So I think that database information is sensible as it drifts into another domain etc. I guess, we don't want to make configuring this a lot of work, but we do want to let you (a) reduce the noise/cost (b) remove things that might be insecure (i.e. message contents) from telemetry.

preardon:
This latest push should have the outlines of everything that needs to be done,

Some observations would be
  - Do we need Messaging & Requests?
  - Request Body is exclusively used for Bodies (MessageBody & RequestBody)
  - I've Added a Specific Brighter Instrumentation Option for Internal Operations, Thoughts?

iancooper:
A lot of work @preardon. Going to merge in, so that we can get ready to cut RC1

PR #3625, opened 2025-06-12
Utilize that SqlDataReaders currently return MemoryStream to avoid manual stream copy

Henr1k80:
- Utilize that SqlDataReaders currently return MemoryStream to avoid manual stream copy.
The internal stream `ToArray` has black magic array allocation for large arrays & `Copy` method uses `ArrayPool` for buffers instead of allocation.

- Dispose streams in case they change implementation or type to require disposing.
MemoryStream does not currently leak anything and does not require disposing.

I considered to change the implementation to only consider MemoryStream and throw exceptions if it is not.
Something like `throw new NotImplementedExpection("GetStream no longer returns MemoryStream")`
Then the unit tests would fail, if the type of returned stream ever changes.
Then the code will be much simpler, as we do not need the fallback code.

Should I change the pull request to only consider MemoryStream and simplify the code?

### DIFF
--- a/src/Paramore.Brighter.MessagingGateway.Postgres/PostgresMessageConsumer.cs
+++ b/src/Paramore.Brighter.MessagingGateway.Postgres/PostgresMessageConsumer.cs
@@ -375,7 +375,7 @@ public partial class PostgresMessageConsumer(
     {
         var id = reader.GetInt64(0);
 
-        using var content = reader.GetStream(3);
+        var content = reader.GetStream(3);
         var message = JsonSerializer.Deserialize<Message>(content, JsonSerialisationOptions.Options)!;
         
         message.Header.Bag["ReceiptHandle"] = id;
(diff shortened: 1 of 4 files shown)


iancooper:
>Should I change the pull request to only consider MemoryStream and simplify the code?
Good suggestion, let's do that. Thanks for these contributions btw. We are close to RC1 on V10, when they will come out.

Henr1k80:
My assumption that is was MemoryStream always is false...

Fallbacks are probably needed

iancooper:
@Henr1k80 Do you want me to review "as is" then?

Henr1k80:
I have provided fallbacks now.

Not for MySql though, it always returns a MemoryStream

iancooper (approved):
Thanks @Henr1k80. I always appreciate the performance fixes.

PR #3627, opened 2025-06-16
Otel and Cloud Events Transports Fixes

iancooper:
We need to add OTel to our transports, which means support for TraceParent, TraceState and Baggage. We also want to add Cloud Events support to our transports.

This PR fixes a range of transport issues with OTel


iancooper:
Doing this transport by. transport to get them green

Issue #3616, opened 2025-05-28
Provide Support for Migration from Mass Transit to Brighter


iancooper:
**Is your feature request related to a problem? Please describe.**
With Mass Transit going commercial, some engineers may want to switch to Brighter. Based on our conversations, it is likely that this would be an app-by-app approach, meaning we would coexist with Mass Transit in the ecosystem for some time.

**Describe the solution you'd like**
The main issue for interop with Mass Transit is mapping. We should provide:
* A default message mapper that maps to and from the Mass Transit format
* Transform pipeline middleware that allows an existing message mapper to be attributed so that we convert the output of that mapping to and from the Mass Transit format
* A how-to guide for moving from Mass Transit to Brighter. We should optimize the instructions for usage with an LLM to help folks using an LLM to convert. 

**Additional context**
We may consider the same pattern for other libraries that are out there: JustSaying, Wolverine, Rebus, and Dapr in the dotnet space, and even Spring and Quarkus in the Java space. (Some of the Dapr work will need to be behind the support for heterogeneous channels that use the Cloud Events type for the channel type).

PR #3634, opened 2025-06-20
[Feature] Create Adaptors for other messaging integration

lillo42:
Add support for integration for other messaging framework/lib (see https://github.com/BrighterCommand/Brighter/issues/3616)
- [x] JustSaying
- [x] Masstransit with envelop payload (basic scenario)

I also following the same approach as on AWS SNS/SQS in case a message doesn't have Brighter headers, I'm going to set a default value 
- MessageId => `Id.Empty`
- MessageType => `MT_EVENT`
- TimeStamp => `DateTimeOffset.UtcNow`


src/Paramore.Brighter.MessagingGateway.Redis/RedisMessageCreator.cs
```diff
@@ -306,14 +304,14 @@ private static HeaderResult<MessageType> ReadMessageType(Dictionary<string, stri
             return new HeaderResult<MessageType>(MessageType.MT_EVENT, true);
         }
 
-        private static HeaderResult<string> ReadMessageId(IDictionary<string, string> headers)
+        private static HeaderResult<Id> ReadMessageId(IDictionary<string, string> headers)
```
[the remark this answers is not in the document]
lillo42: Done

src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageCreator.cs
```diff
@@ -297,7 +297,7 @@ private static HeaderResult<RoutingKey> ReadTopic(string topic)
                 {
                     if (string.IsNullOrEmpty(messageId))
                     {
-                        return new HeaderResult<Id?>(Id.Empty, false);
+                        return new HeaderResult<Id?>(Id.Empty, true);
```
iancooper: Should we return Id. Random in the event that this is missing, and thendebugg log that we have created a random id for a message that does not have one?

src/Paramore.Brighter.Transformers.JustSaying/JustSayingEvent.cs
```diff
@@ -0,0 +1,68 @@
...
+    public string? RaisingComponent { get; set; }
+    
+    /// <inheritdoc />
+    public string? Version { get; set; }
+    
+    /// <inheritdoc />
+    public string? SourceIp { get; set; }
+    
+    /// <inheritdoc />
+    public string? Tenant { get; set; }
```
iancooper: I wonder about creating types for some of these; `Tenant` leaps out as a domain concept rather than just using primitives,  but `Version` and `RaisingComponent` possibly. In favor might be that we would avoid primitives for folks working with JS, negative might be introducing JS concepts into our domain, but as you need to pull in the specified module, that ought to be fine. In addition, IPAddress is a System type, so we should probably use that over a string.

src/Paramore.Brighter.Transformers.JustSaying/JustSayingMessageMapper.cs
```diff
@@ -0,0 +1,291 @@
...
+            }
+            
+            return node.GetValue<string>();
+        }
+        
+        string GetCorrelationId()
+        {
+            var node = doc[nameof(IJustSayingRequest.Conversation)];
+            if (node != null 
+                && node.GetValueKind() == JsonValueKind.String
```
iancooper: It is possible we could extract some duplication here to a method

src/Paramore.Brighter.Transformers.JustSaying/JustSayingTransform.cs
```diff
@@ -0,0 +1,315 @@
...
+        {
+            id = Id.Random;
+        }
+        
+        node[nameof(IJustSayingRequest.Id)] = id.Value;
+    }
+    
+    private static void SetTimeStamp(JsonNode node, Message message)
+    {
+        var jsonNode = node[nameof(IJustSayingRequest.TimeStamp)];
```
iancooper: Again, there is a possible commonality in how we read nodes. You could extract this and the message mapper implementation of reading notes into a helper class.

src/Paramore.Brighter.Transformers.MassTransit/MassTransitMessageAttribute.cs
```diff
@@ -0,0 +1,62 @@
...
+public abstract class MassTransitMessageAttribute : Attribute
+{
+    /// <summary>
+    /// Gets or sets the destination address where messages of this type should be sent.
+    /// </summary>
+    /// <remarks>
+    /// This corresponds to MassTransit's endpoint configuration for message routing. 
+    /// Format should follow transport-specific conventions (e.g., "rabbitmq://host/queue").
+    /// </remarks>
+    public string? DestinationAddress { get; set; }
```
iancooper, line 25: Again, shoud we turn some of these primitives into a type like `DestinationAddress`

src/Paramore.Brighter.Transformers.MassTransit/MassTransitMessageEnvelop.cs
```diff
@@ -0,0 +1,111 @@
...
+    public Id? CorrelationId { get; set; }
+
+    /// <summary>Identifier for the conversation context</summary>
+    public Id? ConversationId { get; set; }
+
+    /// <summary>Identifier of the message initiator</summary>
+    public Id? InitiatorId { get; set; }
+
+    /// <summary>Source address where the message originated</summary>
+    public string? SourceAddress { get; set; }
```
iancooper: These would then map to the types defined for Mass Transit, and not primitives

iancooper (commented):
This is great, thanks. I would like pick up on some of the primitive types, as we always regret those, before we merge.

src/Paramore.Brighter.Transformers.MassTransit/MassTransitMessageAttribute.cs
lillo42: I've changed to `URI`

src/Paramore.Brighter.Transformers.MassTransit/MassTransitMessageEnvelop.cs
lillo42: I've changed to `URI`

src/Paramore.Brighter.Transformers.MassTransit/MassTransitMessageAttribute.cs
lillo42: I'm going let this one as string, because it's an attribute and on attribute it's hard to set non-primitive type

PR #3645, opened 2025-07-02
[Feature] Add support to GCP

lillo42:
I'm adding support to Google Cloud Platform

- [x] Pub/Sub
- [x] Google Cloud Store (Luggage Store)
- [ ] Firestore for Inbox/Outbox/Distributed Lock
- [ ]  Spanner for Inbox/Outbox/Distributed Lock


lillo42:
Waiting for https://github.com/BrighterCommand/Brighter/pull/3659

iancooper (approved):
I have skimmed, and this looks great, follows our approach, but I don't know GCP well enough for a detailed review beyond that. So our usual policy applies: check this follows our conventions, approve in and then fix it up as we find bugs

CharlieDigital:
@iancooper I'm currently reviewing to see if there is any standout issue here with the impl.

src/Paramore.Brighter.MessagingGateway.GcpPubSub/Extensions/ChukExtensions.cs
CharlieDigital, line 1: This file is mis-named `ChuckExtension` -> `ChunkExtension`

iancooper:
> @iancooper I'm currently reviewing to see if there is any standout issue here with the impl.

Thanks, I am less familiar with GCP, so some extra eyes for @lillo42 would be really helpful

src/Paramore.Brighter.MessagingGateway.GcpPubSub/GcpMessagingGatewayConnection.cs
```diff
@@ -0,0 +1,73 @@
...
+        var builder = new PublisherServiceApiClientBuilder { Credential = Credential };
+        PublishConfiguration?.Invoke(builder);
+        return await builder.BuildAsync();
+    }
+    
+    public SubscriberServiceApiClient CreateSubscriberServiceApiClient()
+    {
+        var builder = new SubscriberServiceApiClientBuilder { Credential = Credential };
+        SubscribeConfiguration?.Invoke(builder);
+        return builder.Build();
```
CharlieDigital: Leaving a note here (still reading through code): this is the lower level client and has some inherent instability intended to be managed by the higher level `SubscriberClient`.  Wolverine's Pub/Sub implementation used this on read as well as subscription management which caused its instability when used with cloud GCP Pub/Sub (transient errors do not surface when testing with the emulator).

Some more details here: https://github.com/JasperFx/wolverine/issues/1648

A minimum repro: https://github.com/CharlieDigital/dn-pubsub-error-repro (run against an actual cloud instance for 60-90 seconds and the subscriber will disconnect).

I read through the entire PR and would note that my repro and the Wolverine implementation is using the `StreamingPull` API so the implementation here may not have the same issue.

CharlieDigital (commented):
Would using the `SubscriberClient` API not work for Brighter's design?

https://cloud.google.com/dotnet/docs/reference/Google.Cloud.PubSub.V1/latest/Google.Cloud.PubSub.V1.SubscriberClient

The reason I ask is that the lower level `SubscriberServiceApiClient` has some inherent instability on pulls (which will not surface when testing against the emulator) and will cause the connection to appear "flaky" in actual GCP Pub/Sub.

src/Paramore.Brighter.MessagingGateway.GcpPubSub/GcpMessagingGatewayConnection.cs
iancooper: We should definitely take stock of whether this is not working when it's not running against the emulator. @lillo42, we should also get some feedback from the JET Connect team on this at JET, as they use GCP Pub/Sub as their daily driver.

CharlieDigital: It is overall a very good implementation and especially like the addition of Firestore for inbox/outbox and clever distributed locking.  I will pull this branch and try it out and see how it behaves.  

I wonder if some of the tradeoff here is for the design/architecture of Brighter as the high level client -- as I understand it -- effectively manages a pool of low-level subscriber clients and is optimized for throughput.  It also manages:

- Lease renewal
- Ack
- Nack

I wonder if the usage of the high level client would not fit into the design goals of Brighter (thus my line of questioning on the discussion with respect to Pub/Sub support to try to understand some of the key architecture goals).

iancooper: I will still try to respond to that thread. Brighter has its own message pump, so we tend not to use higher-level libraries that have a message pump, as they often struggle to work with our architecture. We have issues with behaviors, such as acking before the handler has run, that would erode our reliability. But I would need to do some digging unless @lillo42 has some insights to share

iancooper: (As another example, we want to move our MQTT implementation to a lower level library for the same reason)

iancooper: Typically, we deal with instability in our framework (such as RabbitMQ's ensure connection which manages our pool). We might need to apply similar logic here

lillo42: > is the lower-level client and has some inherent instability intended to be managed by the higher level SubscriberClient
Really? I'm going to check, but I use the real GCP Pub/Sub for development and testing (I tried to use the emulator first, but I got some error)

I read the Pub/Sub documentation, and I saw a lot of samples, and that seems to make sense to me to use the low-level API with pull mode (because the Brighter message pump uses the pull mode)

lillo42: Since we are using the pull mode, I think we won't have any issue because in case we have some issue getting a message from Pub/Sub, we are going to retry after some X time (probably after a few milliseconds)

iancooper: A quick review indicates that we would avoid the high-level client because it is its own message pump, which we (and other messaging frameworks) would struggle to adapt to, as we provide our own pump.  We provide a pump because we follow the pattern:

```pseudo
while (! quit message received)
 message = read message from queue/stream
 request = translate message to request (with middleware)
 dispatch request (with middleware)
 ack message
end
```

Because of this, if we fail before the message has been processed, it will not be acknowledged, and thus it will be picked up by a restart (or by another instance acting as a competing consumer). This ensures durable execution, guaranteeing that we will process the message **at least once**.

However, we can certainly examine the implementation of the high-level API to identify Pub-Sub issues that need to be addressed to make the transport robust. If the client is stateful (such as RMQ), we strive to ensure reliability by pooling and maintaining a valid connection for each operation. We can examine that in this case, if appropriate.

If, as you say, you are seeing production issues with the lower-level APIs that is a goods head up we should look at how the high-level client tries to resolve these and emulate that

iancooper: Just in case @CharlieDigital, do you have any insights into what the high-level client needs to do, beyond running a message pump, that we should be aware of?

lillo42: After reading more Pub/Sub docs, I'll add support to the Stream API as well. Using the high-level API seems like the low-level; we will need to do more work. I hope to finish it today.

cc @CharlieDigital

lillo42: Hey @CharlieDigital, I think I've understood the issue that you were talking about.

Basically, for pull-mode, we should use the low-level API; the high-level one doesn't offer support for it ✅ 

For stream-mode, it's highly recommended to use the high-level API  due to some issues, like disconnection. If we use the low-level API, we (Brighter) need to handle this kind of issue. I've just added support to the Stream API using the high-level API

So on the `GcpSubscription`, we have a parameter called `subscriptionMode` where we can switch between `Pull` and `Stream` mode, by default, it'll be `Stream`

CharlieDigital: @iancooper My understanding is that the high level API `SubscriberClient` is effectively a message pump that pools a set of `SubscriberServiceApiClient`.  You can see the code here: https://github.com/googleapis/google-cloud-dotnet/blob/main/apis/Google.Cloud.PubSub.V1/Google.Cloud.PubSub.V1/SubscriberClientImpl.cs#L117-L143 where the `SingleChannel` is a wrapper around a `SubscriberServiceApiClient`.

I'm sorry, I'm not familiar enough with the Brighter architecture to make a distinction of whether this is "good" or "bad", but I suspect that based on your comments and @lillo42, that this is not the right fit for Brighter's architecture.

With the `StreamingPull` API , it *will* disconnect: https://cloud.google.com/pubsub/docs/pull#streamingpull-api

> The StreamingPull API keeps an open connection. The Pub/Sub servers recurrently close the connection after a time period to avoid a long-running sticky connection. The client library automatically reopens a StreamingPull connection.

So this is the "gotcha" that foils many implementations!

src/Paramore.Brighter.MessagingGateway.GcpPubSub/Extensions/ChukExtensions.cs
CharlieDigital: It is now misnamed as `ChukExtension`, missing the `n` :D

PR #3652, opened 2025-07-07
Command Processor Dispatching Strategy

iancooper:
This work enables us to dispatch to a handler not only based on type, but also using other strategies, such as content-based routing of a request that utilizes the specification pattern.


src/Paramore.Brighter.Extensions.DependencyInjection/IBrighterBuilder.cs
```diff
@@ -81,20 +85,20 @@ public interface IBrighterBuilder
         /// </summary>
         /// <param name="assemblies">The assemblies to scan</param>
         /// <returns>This builder, allows chaining calls</returns>
-        IBrighterBuilder MapperRegistryFromAssemblies(params Assembly[] assemblies);
```
lillo42, line 84: Why did you replace the params with IEnumerable? I think during setup, use the params is very helpful

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionSubscriberRegistry.cs
```diff
@@ -78,6 +108,22 @@ public void Register<TRequest, TImplementation>()
...
             _registry.Register<TRequest, TImplementation>();
         }
+        
+        /// <summary>
+        /// Registers this instance.
+        /// </summary>
+        /// <typeparam name="TRequest">The type of the t request.</typeparam>        
+        /// <param name="router">The routing function that takes a request and a request context and returns one or more handlers.</param>
+        /// <param name="handlerTypes">We need a <see cref="IEnumerable{T}"/> of <see cref="Type"/> that router can return to register with the factory</param>
+
```
lillo42: blank line

lillo42 (approved):
LGTM

src/Paramore.Brighter.Extensions.DependencyInjection/IBrighterBuilder.cs
iancooper: Because I needed additional parameters for registering handler types that use dynamic routing, and I wanted that list to be optional. As params needs to be last, that would mean the exclusion list preceded the assembly list, which feels like a poor API design. So I removed the params because we need more than just the params

PR #3669, opened 2025-07-16
Dynamic Request Type

iancooper:
Previously we have used a generic parameter on the MessagePump types (Reactor and Proactor) to determine the type that a message maps to. This limits our ability to determine that from the message itself, i.e. the Cloud Events type, and prevents us from supporting channels with multiple message types. This change supports those scenarios. It supports both consuming - we need to dynamically determine the how to map the consumed message to a derived type from `IRequest` and producing - we need to allow multiple publications on the same topic - and determine which publication to use based on `IRequest`, but disambiguate by cloud events type if there is more than one match.

### DIFF
--- a/tests/Paramore.Brighter.Core.Tests/CommandProcessors/TestDoubles/MyEvent.cs
+++ b/tests/Paramore.Brighter.Core.Tests/CommandProcessors/TestDoubles/MyEvent.cs
@@ -23,7 +23,6 @@ THE SOFTWARE. */
 #endregion
 
 using System;
-using Microsoft.VisualStudio.TestPlatform.ObjectModel;
 
 namespace Paramore.Brighter.Core.Tests.CommandProcessors.TestDoubles
 {
--- a/tests/Paramore.Brighter.Core.Tests/CommandProcessors/When_A_Request_Context_Is_Provided.cs
+++ b/tests/Paramore.Brighter.Core.Tests/CommandProcessors/When_A_Request_Context_Is_Provided.cs
@@ -266,10 +266,7 @@ public class RequestContextPresentTests : IDisposable
         var producerRegistry =
             new ProducerRegistry(new Dictionary<RoutingKey, IAmAMessageProducer>
             {
-                { routingKey, new InMemoryMessageProducer(new InternalBus(), timeProvider, new Publication{RequestType = typeof(MyCommand), Topic = routingKey})
-                {
-                    Publication = new Publication{RequestType = typeof(MyCommand), Topic = routingKey}
-                } },
+                { routingKey, new InMemoryMessageProducer(new InternalBus(), timeProvider, new Publication{RequestType = typeof(MyCommand), Topic = routingKey})} 
             });
 
         var tracer = new BrighterTracer(timeProvider);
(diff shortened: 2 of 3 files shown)


src/Paramore.Brighter/CloudEventsType.cs
```diff
@@ -0,0 +1,95 @@
...
+
+namespace Paramore.Brighter;
+
+/// <summary>
+/// Represents the CloudEvents "type" attribute, describing the event's semantic meaning.
+/// </summary>
+/// <remarks>
+/// See: https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#type
+/// </remarks>
+public sealed class CloudEventsType : IEquatable<CloudEventsType>
```
lillo42, line 36: We can convert it to record, so we can remove everything from ToString()

src/Paramore.Brighter/ProducerKey.cs
```diff
@@ -0,0 +1,67 @@
...
+#endregion
+
+using System;
+
+namespace Paramore.Brighter;
+
+/// <summary>
+/// A producer key is used to identify a producer in the <see cref="ProducerRegistry"/>.
+/// </summary>
+public class ProducerKey
```
lillo42: We can convert it to record and sealed it

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionMessageMapperRegistryBuilder.cs
```diff
@@ -105,37 +120,5 @@ public void SetDefaultMessageMapperAsync(Type defaultMapper)
...
-        public void EnsureDefaultMessageMapperIsRegistered()
-        {
-            if(DefaultMessageMapper != null)
-                serviceCollection.TryAdd(new ServiceDescriptor(DefaultMessageMapper, DefaultMessageMapper, lifetime));
-            if (DefaultMessageMapperAsync != null)
-                serviceCollection.TryAdd(new ServiceDescriptor(DefaultMessageMapperAsync, DefaultMessageMapperAsync,
-                    lifetime));
-        }
-    }
+        }    }
```
lillo42: Add a new line between }

src/Paramore.Brighter.MessagingGateway.RMQ.Async/RmqMessageProducer.cs
```diff
@@ -64,7 +64,11 @@ public partial class RmqMessageProducer : RmqMessageGateway, IAmAMessageProducer
     /// <summary>
     /// The publication configuration for this producer
     /// </summary>
-    public Publication Publication { get { return _publication; } }
+    public Publication Publication
+    {
+        get { return _publication; }
+        set { _publication = (RmqPublication)value ?? throw new ArgumentNullException(nameof(value), "RmqMessageProducer: Publication cannot be null"); }
+}
```
lillo42: Add space

lillo42 (approved):
LGTM

lillo42 (approved):

src/Paramore.Brighter/CloudEventsType.cs
iancooper: Yeah, I worry about not turning all our primitives to records, and what happens if you do. I know I looked at Command and Event, and decided there were problems with using a record type in the past. Perhaps V11?

PR #3672, opened 2025-07-16
Fix TimedOutboxArchiver calling only Archive async method

### LINKED ISSUE #3670: [Bug] TimedOutboxArchiver is always calling async version of archiver
### Describe the bug
`TimedOutboxArchived` is not checking whether the IAmOutbox is sync or async, it is always calling `OutboxArchiver.ArchiveAsync`. This causes background exception when outbox have sync implementation. Messages are not archived.

### To Reproduce

Register synchronous outbox provider and outbox archiver. You will see exceptions logged in the background

### Exceptions (if any)

```
fail: Paramore.Brighter.OutboxArchiver[0]
      Error while archiving from the outbox
      System.NullReferenceException: Object reference not set to an instance of an object.
         at Paramore.Brighter.OutboxArchiver.ArchiveAsync(Int32 minimumAge, CancellationToken cancellationToken, Boolean parallelArchiving) in /_/src/Paramore.Brighter/OutboxArchiver.cs:line 81
```

### Further technical details
- Brighter version 9.9.10
- Include the output of `dotnet --info`
- The OS (Window, Linux, MacOs, etc)

jakoss:
Fixes #3670

TimedOutboxArchiver should check whether the registered outbox provider is sync or async and call proper Archive method instead of using always the async one

### DIFF
--- a/src/Paramore.Brighter.Extensions.Hosting/TimedOutboxArchiver.cs
+++ b/src/Paramore.Brighter.Extensions.Hosting/TimedOutboxArchiver.cs
@@ -12,7 +12,8 @@ namespace Paramore.Brighter.Extensions.Hosting
     {
         private readonly TimedOutboxArchiverOptions _options;
         private static readonly ILogger s_logger = ApplicationLogging.CreateLogger<TimedOutboxSweeper>();
-        private IAmAnOutbox<Message> _outbox;
+        private IAmAnOutboxSync<Message> _outboxSync;
+        private IAmAnOutboxAsync<Message> _outboxAsync;
         private IAmAnArchiveProvider _archiveProvider;
         private readonly IDistributedLock _distributedLock;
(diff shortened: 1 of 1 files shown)


lillo42 (approved):

iancooper (approved):
Thanks @jakoss

iancooper:
There are a few CI errors that I need to look at before I can do a release @jakoss. I know that MQTT is erratic, and AWS and RMQ can sometimes be, but I probably need to get to the bottom of the latter two before we can ship this one

jakoss:
No worries, we are months before any production release, the archiver does not block anything

iancooper:
BTW, do we know if this fix needs rolling forward into master for V10

jakoss:
TBH i haven't looked at the V10 side of it

EDIT: seems like we need same adjustement there

PR #3675, opened 2025-07-19
Change names of Brighter builder methods

iancooper:
Change names of Brighter builder methods to AddProducers and AddConsumers as existing names are a leaky abstraction


holytshirt:
Think it's much clearer, general people don't know what a service activator is. Consumer and Producer are much more common language.

Issue #2880, opened 2023-10-22
[Feature] Support for Polly v8 policies with backwards compatibility


iancooper:
**Is your feature request related to a problem? Please describe.**
Polly, which we use for blocking retry or circuit breakers (blocking here in the sense that a lock on a queue is not released or an offset committed) now has released v8 API. If we want to support the v8API we will need to consider how to support this, as well as V7 (see https://github.com/App-vNext/Polly)

Instead of policies, we now have resilience pipelines as a target.

**Describe the solution you'd like**
The new resilience builder pipeline is a welcome addition, as we ourselves have had to add support for a single attribute that has multiple policies over multiple policy attributes. 

So we need a version of our attribute that instead of taking a policy name, takes a resilience pipeline name. In order to retain backwards compatibility with folks that still want the v7 API we either need:

* A UseResiliencePipeline attribute that uses the V8 approach, keeping UsePolicy for v7
* Additional parameters for the UsePolicy attribute to allow setting a resilience pipeline over an array of policies

There are trade-offs to these, but the latter might be more obvious to folks.

**Describe alternatives you've considered**
The main alternative is to stick on v7, but v8 is more compatible with our needs, not less.

martincostello:
Happy to help out with this migration.

FYI, you can upgrade to v8 in a non-breaking way as it retains the previous APIs (i.e. #2918 should be safe to take) - you just won't get the performance benefits of the new API in Polly.Core.

iancooper:
@martincostello Help would be appreciated. I think we have two options. Move in v10 and make v8 style Polly a dependency, or support both syntaxes in our middleware for a bit.

iancooper:
See also: https://github.com/BrighterCommand/Brighter/issues/1197 for Polly issues for V10

Issue #1197, opened 2020-07-16
[Feature] Support Polly Context


iancooper:
**Is your feature request related to a problem? Please describe.**
We cannot access the Polly Context in a policy, to allow logging for example, because we do not use the version of ExecuteAsync which supplies a context

**Describe the solution you'd like**
Pass a Polly Context to the ExecuteAsync call. We may need to use our own internal context to transport a Polly Context through to the handler site. Consumers can then set the Polly Context in the context bag and thus pass values into the call handling site.

**Describe alternatives you've considered**
There are few options around the pipeline that allow us to access the Polly context when logging. We might be able to log the exception that bubbles out of Polly etc. but we don't get to see the parameters of the retry or fallback

**Additional context**
Polly Contexts: https://github.com/App-vNext/Polly/wiki/Keys-and-Context-Data

Issue #1065, opened 2020-03-03
Use Polly Timeout for Timeout


iancooper:
**Is your feature request related to a problem? Please describe.**
We have a Timeout feature that is based on MS code (based on the old idea of start timer with callback, if callback finishes before work, time out). Polly not supports a similar Timeout policy, and our code is erratic, so we should deprecate our version for Polly

**Describe the solution you'd like**
We should mark the attribute as deprecated and recommend moving towards Polly policies for this. We can re-implement our handler in terms of Polly for now, by adding a default timeout Policy key, creating a default timeout policy, and using that to implement.

Even if we deprecate, there is not necessarily a race to remove, if we implement in terms of Polly

**Describe alternatives you've considered**
We could have kept our policy, it's not broken, but it is unreliable in testing

**Additional context**
https://github.com/App-vNext/Polly/wiki/Timeout

PR #3677, opened 2025-07-21
feat:  Add  support to Polly Resilience Pipeline

lillo42:
Add support to Polly Resilience Pipeline via a new attribute 

- New attributes (`UseResiliencePipelineAttribute` and `UseResiliencePipelineAsyncAttribute`)
- Changes on `IRequestContext` to have support to `ResiliencePipeline` and `ResilienceContext`
-  Mark `TimeoutPolicyAttribute` as deprecated

This PR should cover this GH Issues:

https://github.com/BrighterCommand/Brighter/issues/2880
https://github.com/BrighterCommand/Brighter/issues/1197
https://github.com/BrighterCommand/Brighter/issues/1065


src/Paramore.Brighter/ControlBusSenderFactory.cs
```diff
@@ -69,6 +69,7 @@ public IAmAControlBusSender Create<T, TTransaction>(IAmAnOutbox outbox,
                 CommandProcessorBuilder.StartNew()
                 .Handlers(new HandlerConfiguration())
                 .DefaultPolicy()
+                .DefaultResilencePipeline()
```
iancooper: I wonder if we have DefaultResilencePipeline() and internally call DefaultPolicy i.e. don't need to have both on the API but need the results for backwards compatibility?

src/Paramore.Brighter.ServiceActivator/ControlBus/ControlBusReceiverBuilder.cs
```diff
@@ -183,6 +185,7 @@ public Dispatcher Build(string hostName)
             commandProcessor = CommandProcessorBuilder.StartNew()
                 .Handlers(new HandlerConfiguration(subscriberRegistry, new ControlBusHandlerFactorySync(_dispatcher, () => commandProcessor)))
                 .Policies(policyRegistry)
```
iancooper: I wonder if we need a single configuration endpoint that we can pass both into, making them optional, and then determine which one to use. I think the separate APIs may be more confusing than an API that documents you can add old school, new school, or both

iancooper (commented):
Thanks for this, it will be great to fix this in V10.

1. I wonder about having both DefaultPolicy and DefaultResiliencePolicy on the HostBuilder extensions, maybe we just need the new one, and call the old one inside>
2. I think we should replace our internal uses of Polly with the Resilience Builder. I think that is mainly in OutboxProducerMediator, though it may have crept into some other places. Our internal code should move on, and no one externally depends on that. 
3. We may want to mark UsePolicy as obsolete in the next version; it would be good to clean up and drop it by then.

src/Paramore.Brighter/Policies/Attributes/UsePolicyAsyncAttribute.cs
```diff
@@ -34,19 +34,20 @@ namespace Paramore.Brighter.Policies.Attributes
     /// Class UsePolicyAttribute.
     /// This attribute supports the use of <a href="https://github.com/michael-wolfenden/Polly">Polly</a> to provide quality of service around exceptions
```
slang25:
```suggestion
    /// This attribute supports the use of <a href="https://github.com/App-vNext/Polly">Polly</a> to provide quality of service around exceptions
```

src/Paramore.Brighter/Policies/Attributes/UsePolicyAtttribute.cs
```diff
@@ -34,19 +34,20 @@ namespace Paramore.Brighter.Policies.Attributes
     /// Class UsePolicyAttribute.
     /// This attribute supports the use of <a href="https://github.com/michael-wolfenden/Polly">Polly</a> to provide quality of service around exceptions
```
slang25:
```suggestion
    /// This attribute supports the use of <a href="https://github.com/App-vNext/Polly">Polly</a> to provide quality of service around exceptions
```

src/Paramore.Brighter/Policies/Attributes/UseResiliencePipelineAsyncAttribute.cs
```diff
@@ -0,0 +1,73 @@
...
+
+using System;
+using Paramore.Brighter.Policies.Handlers;
+using Polly;
+using Polly.Registry;
+
+namespace Paramore.Brighter.Policies.Attributes;
+
+/// <summary>
+/// This attribute supports the use of <a href="https://github.com/michael-wolfenden/Polly">Polly</a> to provide quality of service around exceptions
```
slang25:
```suggestion
/// This attribute supports the use of <a href="https://github.com/App-vNext/Polly">Polly</a> to provide quality of service around exceptions
```

src/Paramore.Brighter/Policies/Handlers/ExceptionPolicyHandlerAsync.cs
```diff
@@ -36,7 +36,7 @@ namespace Paramore.Brighter.Policies.Handlers
     /// Class ExceptionPolicyHandlerAsync.
     /// The <see cref="UsePolicyAttribute" /> supports the use of <a href="https://github.com/michael-wolfenden/Polly">Polly</a> to provide quality of service around exceptions
```
slang25:
```suggestion
    /// The <see cref="UsePolicyAttribute" /> supports the use of <a href="https://github.com/App-vNext/Polly">Polly</a> to provide quality of service around exceptions
```

src/Paramore.Brighter/Policies/Handlers/ResilienceExceptionPolicyHandler.cs
```diff
@@ -0,0 +1,99 @@
...
+
+using Paramore.Brighter.Policies.Attributes;
+using Polly;
+using Polly.Registry;
+
+namespace Paramore.Brighter.Policies.Handlers;
+
+/// <summary>
+/// Class ExceptionPolicyHandler.
+/// The <see cref="UsePolicyAttribute"/> supports the use of <a href="https://github.com/michael-wolfenden/Polly">Polly</a> to provide quality of service around exceptions
```
slang25:
```suggestion
/// The <see cref="UsePolicyAttribute"/> supports the use of <a href="https://github.com/App-vNext/Polly">Polly</a> to provide quality of service around exceptions
```

src/Paramore.Brighter/Policies/Handlers/ResilienceExceptionPolicyHandlerAsync.cs
```diff
@@ -0,0 +1,88 @@
...
+using Paramore.Brighter.Extensions;
+using Paramore.Brighter.Policies.Attributes;
+using Polly;
+using Polly.Registry;
+
+namespace Paramore.Brighter.Policies.Handlers;
+
+/// <summary>
+/// Class ExceptionPolicyHandlerAsync.
+/// The <see cref="UsePolicyAttribute" /> supports the use of <a href="https://github.com/michael-wolfenden/Polly">Polly</a> to provide quality of service around exceptions
```
slang25:
```suggestion
/// The <see cref="UsePolicyAttribute" /> supports the use of <a href="https://github.com/App-vNext/Polly">Polly</a> to provide quality of service around exceptions
```

slang25 (commented):
By complete coincidence, I made a start on a similar change this morning, and then found this PR 😅 

I would expect that this change also removes the "old" non-core approach, so the end result would be only a reference to `Polly.Core`, it would be more of an upgrade burden for folks, but I think it keeps things more simple.

src/Paramore.Brighter/Policies/Handlers/ResilienceExceptionPolicyHandlerAsync.cs
iancooper: Shows how long ago we adopted Polly :-D

iancooper:
> By complete coincidence, I made a start on a similar change this morning, and then found this PR 😅
> 
> I would expect that this change also removes the "old" non-core approach, so the end result would be only a reference to `Polly.Core`, it would be more of an upgrade burden for folks, but I think it keeps things more simple.

As we use an attribute to indicate what Middleware you want, it's easier for us to declare that Obsolete and indicate that we will drop in V11, and then add an attribute for the new Middleware that adopts upgraded Polly policies. That gives folks time to be aware change is coming, and then shift.

I agree that internally to Brighter, though, so not in a middleware pipeline but where the framework uses it for resilience (such as producers etc.) we should move to the new approach. We should only keep the old approach for end users

src/Paramore.Brighter/CommandProcessorBuilder.cs
```diff
@@ -83,12 +83,11 @@ namespace Paramore.Brighter
     /// </list> 
     /// </summary>
     public class CommandProcessorBuilder : INeedAHandlers,
-        INeedPolicy,
+        INeedPolly,
```
iancooper: I might have preferrred INeedResilience

src/Paramore.Brighter/CommandProcessorBuilder.cs
```diff
@@ -145,39 +144,35 @@ public INeedAHandlers ConfigureFeatureSwitches(IAmAFeatureSwitchRegistry feature
...
         }
 
-        /// <summary>
-        /// Use this if you do not require a policy and only want to retry once(i.e. No Tasks Queues or QoS needs).
-        /// </summary>
-        /// <returns>INeedLogging.</returns>
-        public INeedResiliencePipeline DefaultPolicy()
+        
+        /// <inheritdoc />
+        public INeedMessaging DefaultPolly()
```
iancooper: And then DefaultResilience

tests/Paramore.Brighter.Core.Tests/FeatureSwitch/When_A_Handler_Is_Feature_Switch_Missing_Config_Exception.cs
```diff
@@ -46,8 +46,7 @@ public FeatureSwitchByConfigMissingConfigStrategyExceptionTests()
                 .StartNew()
                 .ConfigureFeatureSwitches(featureSwitchRegistry)
                 .Handlers(new HandlerConfiguration(registry, handlerFactory))
-                .DefaultPolicy()
-                .DefaultResilencePipeline()
+                .DefaultPolly()
```
iancooper: DefaultResilience?

src/Paramore.Brighter/CommandProcessorBuilder.cs
```diff
@@ -377,22 +377,22 @@ public interface INeedAHandlers
     /// <summary>
     /// Defines an interface for configuring Polly resilience pipelines within Brighter.
     /// </summary>
-    public interface INeedPolly
+    public interface INeedResilience
```
iancooper, line 382: I think this reads nicely, thanks for making that change

iancooper (commented):
This is looking good. The one thing we should change is the internal usage. For example, in `OutboxProducerMediatory` the RetryAsync method supports our retries, but is policy not resilience based.

```       
 private async Task<bool> RetryAsync(
            Func<CancellationToken, Task> send,
            RequestContext? requestContext,
            bool continueOnCapturedContext = true,
            CancellationToken cancellationToken = default)
        {
            var policy = _policyRegistry.Get<AsyncPolicy>(CommandProcessor.RETRYPOLICYASYNC);
            
            var result = await policy
                .ExecuteAndCaptureAsync(send, cancellationToken, continueOnCapturedContext)
                .ConfigureAwait(continueOnCapturedContext);

            if (result.Outcome != OutcomeType.Successful)
            {
                if (result.FinalException != null)
                {
                    Log.ExceptionWhilstTryingToPublishMessage(s_logger, result.FinalException);
                    CheckOutstandingMessages(requestContext);
                }

                return false;
            }

            return true;
        }
```

There may be some others. They should use default resilience pipelines that replace the default policies used here.

iancooper:
@lillo42 This error usually pops up when we have changed a constructor that is called via reflection and thus does not get picked up by refactoring tools:

```
 Error Message:
   System.MissingMethodException : Constructor on type 'Paramore.Brighter.OutboxProducerMediator`2[[Paramore.Brighter.Message, Paramore.Brighter, Version=10.0.0.0, Culture=neutral, PublicKeyToken=b90f5623ae237e2c],[System.Data.Common.DbTransaction, System.Data.Common, Version=9.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a]]' not found.
  Stack Trace:
     at System.RuntimeType.CreateInstanceImpl(BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture)
   at Paramore.Brighter.Extensions.DependencyInjection.ServiceCollectionExtensions.BuildOutBoxProducerMediator(IServiceProvider serviceProvider, Type transactionType, ProducersConfiguration busConfiguration, IPolicyRegistry`1 policyRegistry, IAmAnOutbox outbox) in /_/src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs:line 480
   at Paramore.Brighter.Extensions.DependencyInjection.ServiceCollectionExtensions.<>c__DisplayClass2_0.<AddProducers>b__2(IServiceProvider serviceProvider) in /_/src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs:line 237
```

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -70,32 +71,58 @@ public partial class CommandProcessor : IAmACommandProcessor
...
+        /// </list>
+        /// </remarks>
+        public const string RequestReply = "Paramore.Brighter.CommandProcessor.RequestReply";
+        
         /// <summary>
         /// Use this as an identifier for your <see cref="Policy"/> that determines for how long to break the circuit when communication with the Work Queue fails.
         /// Register that policy with your <see cref="IPolicyRegistry{TKey}"/> such as <see cref="PolicyRegistry"/>
         /// You can use this an identifier for you own policies, if your generic policy is the same as your Work Queue policy.
         /// </summary>
+        [Obsolete("Migrate to Resilience Pipeline with OutboxProducer", error: false)]
```
iancooper, line 101: Not  a problem but I expect we tell users to migrate to the resilience pipeline, and keep our strings for internal usage

iancooper (approved):

src/Paramore.Brighter/OutboxProducerMediator.cs
```diff
@@ -1041,48 +1041,61 @@ private void OutstandingMessagesCheck(RequestContext? requestContext)
...
-                    Log.ExceptionWhilstTryingToPublishMessage(s_logger, result.FinalException);
-                    CheckOutstandingMessages(requestContext);
+                    await resiliencePipeline
+                        .ExecuteAsync(async context => await send(context.CancellationToken), requestContext.ResilienceContext)
+                        .ConfigureAwait(continueOnCapturedContext);
                 }
-
+                else
+                {
+                    await resiliencePipeline.ExecuteAsync(async _ => await send(cancellationToken), cancellationToken)
```
martincostello, line 1087: You should pass through the cancellation token from the delegate arguments, rather than re-use `cancellationToken` in the closure, as the token might be different (e.g. combined with a token internal to Polly for handling timeouts).

For example:

```csharp
await resiliencePipeline.ExecuteAsync(async token => await send(token), cancellationToken)
```

iancooper: Thanks @martincostello. @lillo42 one to pick up.

Issue #3655, opened 2025-07-10
[Feature] Add support to Apache Pulsar


lillo42:
**Describe the solution you'd like**
Would be nice if Brighter have support to [Apache Pulsar](https://pulsar.apache.org/).

Apache Pulsar is an open-source, distributed messaging and streaming platform designed for high performance, scalability, and durability. It operates on a publish-subscribe model, where producers send messages to "topics" and consumers subscribe to those topics to receive the data.

iancooper:
@lillo42 I am guessing that we stalled behind V10?

lillo42:
I'm waiting for V10 to be released to chase you for merging it

iancooper:
@lillo42 Where is this?

PR #3682, opened 2025-07-23
feat: Add support Apache Pulsar

lillo42:
Add support to Apache Pulsar https://github.com/BrighterCommand/Brighter/issues/3655

- Apache Pulsar will create the Topic/Subscription if it doesn't exists
- The current .NET implementation of Apache Pulsar doesn't have support to dead letter queue or retry queue


src/Paramore.Brighter.MessagingGateway.Pulsar/PulsarMessageConsumerFactory.cs
```diff
@@ -0,0 +1,62 @@
...
+        => CreatePulsarConsumer(subscription);
+
+    private PulsarMessageConsumer CreatePulsarConsumer(Subscription subscription)
+    {
+        if (subscription is not PulsarSubscription pulsarSubscription)
+        {
+            throw new ConfigurationException("We expect PulsarSubscription or PulsarSubscription<T> as a parameter");
+        }
+        
+        var background = s_backgroundConsumers.GetOrAdd(pulsarSubscription, CreateConsumerBackground);
```
iancooper, line 38: We might want to chat about what this is, as I don't know Pulsar? But I might need to understand it to see how this works with our pump

src/Paramore.Brighter.MessagingGateway.Pulsar/PulsarMessageConsumerFactory.cs
```diff
@@ -0,0 +1,62 @@
...
+            .ReadCompacted(pulsarSubscription.ReadCompacted)
+            .SubscriptionName(pulsarSubscription.ChannelName.Value)
+            .SubscriptionType(pulsarSubscription.SubscriptionType)
+            .Topic(pulsarSubscription.RoutingKey);
+        
+        pulsarSubscription.Configuration?.Invoke(builder);
+
+        var consumer = builder.Create();
+
+        return new PulsarBackgroundMessageConsumer(pulsarSubscription.NoOfPerformers, consumer);
```
iancooper, line 60: We tend to be single threaded and rely on our performers to start multiple pumps. This would seen to counteract that, but, again I don't know Pulsar well enough to understand that.

src/Paramore.Brighter.MessagingGateway.Pulsar/PulsarBackgroundMessageConsumer.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// Starts the background message consumption loop
+    /// </summary>
+    /// <remarks>
+    /// Implements reference counting:
+    /// <list type="bullet">
+    ///   <item><description>First call starts the background loop</description></item>
+    ///   <item><description>Subsequent calls increment the reference count but don't start additional loops</description></item>
+    /// </list>
+    /// </remarks>
+    public void Start()
```
iancooper, line 53: Why do we have a background loop here? Why not just consume as we do elsewhere and rely on the single-threaded pump to scale performers?

src/Paramore.Brighter.MessagingGateway.Pulsar/PulsarBackgroundMessageConsumer.cs
```diff
@@ -0,0 +1,119 @@
...
+    /// 
+    /// <para>Errors during message reception are silently ignored to maintain loop continuity.</para>
+    /// </remarks>
+    private async Task ExecuteAsync(CancellationToken  cancellationToken)
+    {
+        while (!cancellationToken.IsCancellationRequested)
+        {
+            try
+            {
+                var message = await consumer.Receive(cancellationToken);
```
iancooper, line 83: This feels like a pump within our pump?

iancooper (commented):
I am probably missing something here, but this feels like we are re-implementing our message pump inside the consumer, which doesn't feel like the "right thing" to be doing. Maybe we need a catch up conversation

Issue #3680, opened 2025-07-22
[Bug] Outbox + Sweeper System.NullReferenceException:


lillo42:
### Describe the bug
When I run the outbox sweeper without tracer, I'm getting `NullReferenceException`

### To Reproduce
Setup outbox sweeper,

### Exceptions (if any)
```csharp
System.NullReferenceException: Object reference not set to an instance of an object.
   at Paramore.Brighter.OutboxProducerMediator`2.BackgroundDispatchUsingAsync(Int32 amountToClear, TimeSpan timeSinceSent, Boolean useBulk, RequestContext requestContext, Dictionary`2 args, CancellationToken cancellationToken) in /_/src/Paramore.Brighter/OutboxProducerMediator.cs:line 606
   at Paramore.Brighter.OutboxProducerMediator`2.ClearOutstandingFromOutboxAsync(Int32 amountToClear, TimeSpan minimumAge, Boolean useBulk, RequestContext requestContext, Dictionary`2 args, CancellationToken cancellationToken) in /_/src/Paramore.Brighter/OutboxProducerMediator.cs:line 418
   at Paramore.Brighter.OutboxSweeper.SweepAsync() in /_/src/Paramore.Brighter/OutboxSweeper.cs:line 79
   at Paramore.Brighter.Outbox.Hosting.TimedOutboxSweeper.Sweep(Object state) in /_/src/Paramore.Brighter.Outbox.Hosting/TimedOutboxSweeper.cs:line 124
   at Paramore.Brighter.Outbox.Hosting.TimedOutboxSweeper.Sweep(Object state) in /_/src/Paramore.Brighter.Outbox.Hosting/TimedOutboxSweeper.cs:line 131
   at System.Threading.Tasks.Task.<>c.<ThrowAsync>b__128_1(Object state)
   at System.Threading.ThreadPoolWorkQueue.Dispatch()
   at System.Threading.PortableThreadPool.WorkerThread.WorkerThreadStart()
```

### Further technical details
- Brighter version: 10.0.0-preview.6

PR #3683, opened 2025-07-23
fix: NullReference with outbox sweeper

lillo42:
fixes: https://github.com/BrighterCommand/Brighter/issues/3680


iancooper (approved):
Thanks @lillo42

Issue #3661, opened 2025-07-12
[Bug] CloudEvents only works for wrap pipeline


lillo42:
### Describe the bug
The current implementation of `CloudEventsAttribute` only works for wrap pipeline, because it extend the `WrapWithAttribute` 

### To Reproduce
Add the `CloudEventsAttribute` in a custom mapper, send and consume a message

### Further technical details
- Brighter version: 10.0.0-preview-6

PR #3686, opened 2025-07-26
fix: Unwrap Cloudevent

lillo42:
Fix https://github.com/BrighterCommand/Brighter/issues/3661

Fixes unwrap Cloudevent when it's using  JSON format


iancooper (approved):
Thanks @lillo42

Issue #3663, opened 2025-07-12
[Feature] Use UUID/GUID v7 for ID generation


lillo42:
**Is your feature request related to a problem? Please describe.**
In many part of the Brighter code we are generation an Guid V4, I think it would be nice if Brighter could use V7, specially for message ID, it could improve performance for outbox/inbox pattern with RDS 

**Describe the solution you'd like**
Replace `Guid.NewGuid()` with `Guid.CreateVersion7()`
 
**Describe alternatives you've considered**
N\A

**Additional context**
https://learn.microsoft.com/en-us/dotnet/api/system.guid.createversion7?view=net-9.0

jakoss:
Would that imply that v10 requires .NET 9? Because that is a blocker for a lot of people unfortunately

lillo42:
No, we can use C# preprocessor for .NET9+ to use the `Guid.CreateVersion7` and other version we  will use the `Guid.NewGuid()`

PR #3687, opened 2025-07-28
feat: Use GUID v7 for .NET 9+

lillo42:
Implmenting this https://github.com/BrighterCommand/Brighter/issues/3663

I'm replacing `Guid.NewGuid()` to  `Id.Random` or `Uuid.New()` 

We should use `Uuid` instance of `Guid` because the `Uuid` has this logic to call the correct method depending on .NET runtime

# Why should we use GUID V7?
GUID  v7 strikes an excellent balance between uniqueness, time-sortability, and database performance, making it a powerful and efficient choice for generating unique identifiers in modern applications, especially when dealing with databases and distributed systems

1. Improved Database Performance (Reduced Index Fragmentation)
2. Time-Sortability


iancooper (approved):
Ouch, that's going to hurt my merge as I need to change them all into function calls :-D

PR #3688, opened 2025-07-31
feat: bulk dispatch to support circuit breaking

mikechrisfreeman:
Currently only `AzureServiceBusMessageProducer` and `InMemoryProduce` implement `IAmABulkMessageProducerAsync`. 

The current behaviour batches messages per topic and batch size then publishes the batch. If one batch fails an exception is raised and successful batches are not recorded. This therefore renders the `MarkDispatchedAsync` on successful messages useless.

Proposal is to bring the Bulk dispatch inline with the single dispatch implementation so that it can be wrapped in a retry and make use of `IAmAnOutboxCircuitBreaker`. A further benefit is individual batches can be retried as opposed to the entire list of messages and better control over implementation specific batching, asb can be quite funny.


mikechrisfreeman:
@iancooper Added a proposal for circuit breaking to support bulk dispatching. I'll continue on the premise that this implementation is ok and make changes if we have differing opinions.

iancooper (commented):
Looks good, let's give it a shot. I've added to V10 as its a breaking change

src/Paramore.Brighter/IAmAMessageBatch.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System.Collections.Generic;
+
+namespace Paramore.Brighter;
+
+/// <summary>
+/// Class MessageBatch
+/// A collection of T <see cref="Message"/>
+/// </summary>
+public interface IAmAMessageBatch<out T> : IAmAMessageBatch
```
mikechrisfreeman, line 9: This abstraction allows for vendor specific batching to be implemented.. Currently our batching implementation is quite primitive - it fetches a batch of messages from the outbox and then attempt to batch them up, not considering whether it exceeds the vendor specific batching limitations.

We intend to add a further PR  to fix this, for now its setting it up for that use case.

src/Paramore.Brighter.MessagingGateway.AzureServiceBus/AzureServiceBusMessageProducer.cs
```diff
@@ -117,47 +118,59 @@ public async Task SendAsync(Message message, CancellationToken cancellationToken
...
+
+    /// <summary>
+    /// Sends a batch of messages.
+    /// </summary>
+    /// <param name="batch">A batch of messages to send</param>
+    /// <param name="cancellationToken">The Cancellation Token.</param>
+    /// <exception cref="NotImplementedException"></exception>
+    public async Task SendAsync(IAmAMessageBatch batch, CancellationToken cancellationToken)
+    {
+        if (batch is not MessageBatch messageBatch)
```
mikechrisfreeman, line 161: An azure specific message batch will be implemented later so that we can fine tune vendor specific constraints

preardon (approved):
Looks good to me, thanks for getting this in

iancooper:
@mikechrisfreeman Let us know when this is done, and I'll review

mikechrisfreeman:
> @mikechrisfreeman Let us know when this is done, and I'll review

Thanks @iancooper  - Updated from main, resolved merge conflicts and ready for review.

I did notice that RocketMessageProducerFactory hasn't been updated to use the new IAmAMessageProducerFactory interface. Though i couldn't see any tests failing as a result. :/ Is that intentional?

iancooper (approved):
Looks good. Thanks for this @mikechrisfreeman

PR #3711, opened 2025-08-11
fix: correctly wrapping/unwrapping servicebus message user properties

mikechrisfreeman:
Fixes 3 bugs

# issue 1 service bus User Properties only support value types

We're incorrectly setting user properties with complex types. Servicebus message user properties only supports value types.

image

ref : https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.servicebus.message.userproperties?view=azure-dotnet-legacy&viewFallbackFrom=azure-dotnet

# issue 2 Partition key not set

# issue 2 duplicate keys when re-enqueing service bus messages


lillo42 (approved):

mikechrisfreeman:
@iancooper @preardon  - be good to get some reviews here when you get a moment. I've got a pending pr depending on this.

🙇‍♂️

PR #3712, opened 2025-08-11
fix: RabbitMQ Async pulling

lillo42:
The RabbitMQ Async was using the wrong variable for body payload


iancooper:
Thanks @lillo42 I wonder if this explains the odd test failure with the mangled body content.

PR #3717, opened 2025-08-14
[Feature] Use ServiceBusMessageBatch with fallback

mikechrisfreeman:
### Problem
Currently the IServiceBusSenderWrapper implements SendAsync(ServiceBusMessage message..) and SendAsync(ServiceBusMessage[] messages, ..)

Whilst premium tiers of Azure Service Bus support message sizes of 100MB a batch can only be a max size of 1MB.

We're using Microsofts SendMessage**s**() implementation, it does not make use of ServiceBusMessageBatch which gives us a view on when we're exceeding the threshold.

###  Solution
Proposal to use ServiceBusMessageBatch.TryAddMessage which will allow us to batch on size, not on number. A fallback solution if a single message exceeds the allowed batch size (1MB) to use SendAsync(ServiceBusMessage message, ..) 

This will allow use to batch on size and but also produce messages up to Azure Service Bus allowed quotas.

(an inflight pr will allow us to make these changes with little impact as generic batches need to implement an IAmAMessageBatch interface)


Brighter.sln
```diff
@@ -2046,6 +2046,18 @@ Global
 		{5D57A811-A3C2-42BA-A086-A5BFE8B590C7}.Release|Mixed Platforms.Build.0 = Release|Any CPU
```
mikechrisfreeman: these updates aren't required

src/Paramore.Brighter.ServiceActivator.Control.Api/Properties/launchSettings.json
```diff
@@ -0,0 +1,12 @@
+{
```
mikechrisfreeman: can revert this

mikechrisfreeman (commented):
Initial review with a few comments

preardon (approved):
Looks good to me

iancooper (approved):
Not an ASB expert, but from what I understand, this looks sensible

PR #3727, opened 2025-08-15
docs: reproduce invalid operation on iamcommandprocessor

mikechrisfreeman:
This draft pr is used for documentation purposes and should not be merged. it serves to illustrate an issue we have with using a scoped dbcontext whilst resolving the command processor.


iancooper:
Fixed under #3754

PR #3752, opened 2025-08-28
Do not use a Semaphore to sequence calls to the explicit ClearOutbox

iancooper:
Before the global lock, we had a semaphore on the explicit clear that made calls to ClearOutbox for a list of message IDs in sequence. This causes a scaling issue because your process can only clear one set of messages at a time. Under load, for example, when responding to an HTTP API, this will likely lead to backpressure, queuing, and eventually 502 Bad Gateway errors.

This is no longer needed as we have a global lock


iancooper:
For some reason there are too many files here.

Accidentally compared to master

PR #3754, opened 2025-08-29
Transaction Provider Scope

iancooper:
The Transaction Provider is currently a singleton for the CommandProcessor, and not set per instance of the CommandProcessor. This creates an issue that when the CommandProcessor depends on an ITransactionProvider that has state i.e. DbContext, if the hosting application chooses to scope the DbContext per request, it will not have the same scope as the CommandProcessor, resulting in the code using different DbContexts.

### DIFF
--- a/src/Paramore.Brighter/CommandProcessor.cs
+++ b/src/Paramore.Brighter/CommandProcessor.cs
@@ -946,7 +946,7 @@ namespace Paramore.Brighter
         // type before we call it.
         private Id CallDepositPost<TRequest>(
             TRequest actualRequest, 
-            IAmABoxTransactionProvider? iamABoxTransactionProvider,
+            IAmABoxTransactionProvider? transactionProvider,
             RequestContext? requestContext, 
             Dictionary<string, object>? dictionary, 
             string? batchId,
@@ -971,7 +971,7 @@ namespace Paramore.Brighter
             }
 
             return CallMethodAndPreserveException(() =>
-                (deposit?.Invoke(this, [actualRequest, iamABoxTransactionProvider, requestContext, dictionary, batchId]) as Id)!
+                (deposit?.Invoke(this, [actualRequest, transactionProvider, requestContext, dictionary, batchId]) as Id)!
             );
         }
 
--- a/src/Paramore.Brighter/CommandProcessor.cs
+++ b/src/Paramore.Brighter/CommandProcessor.cs
@@ -946,7 +946,7 @@ namespace Paramore.Brighter
         // type before we call it.
         private Id CallDepositPost<TRequest>(
             TRequest actualRequest, 
-            IAmABoxTransactionProvider? amABoxTransactionProvider,
+            IAmABoxTransactionProvider? iamABoxTransactionProvider,
             RequestContext? requestContext, 
             Dictionary<string, object>? dictionary, 
             string? batchId,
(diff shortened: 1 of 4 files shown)


iancooper:
Will fix https://github.com/BrighterCommand/Brighter/issues/3755

src/Paramore.Brighter.Sqlite.EntityFrameworkCore/SqliteEntityFrameworkTransactionProvider.cs
```diff
@@ -11,15 +11,15 @@ namespace Paramore.Brighter.Sqlite.EntityFrameworkCore
     /// A connection provider that uses the same connection as EF Core
     /// </summary>
     /// <typeparam name="T">The Db Context to take the connection from</typeparam>
-    public class SqliteEntityFrameworkConnectionProvider<T> : RelationalDbTransactionProvider where T: DbContext
+    public class SqliteEntityFrameworkTransactionProvider<T> : RelationalDbTransactionProvider where T: DbContext
```
mikechrisfreeman, line 14: I notice you've renamed this to ...TransactionProvider. Is it worth renaming all the other TransactionProviders?

i.e.

MySqlEntityFramework**Connection**Provider
MsSqlEntityFrameworkCore**Connection**Provider
PostgreSqlEntityFramework**Connection**Provider

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -771,7 +789,7 @@ public Id DepositPost<TRequest>(
             Dictionary<string, object>? args = null
         ) where TRequest : class, IRequest
         {
-            return CallDepositPost(request, s_defaultTransactionProvider, requestContext, args, null, s_transactionType);
```
mikechrisfreeman, line 774: Do we have a change in behaviour here? A transactionProvider was passed in by default without the need to provide it when using DepositPost. Does that mean we'll no longer use a transaction when invoking this method?

mikechrisfreeman (commented):
LGTM with a couple questions.

src/Paramore.Brighter.Sqlite.EntityFrameworkCore/SqliteEntityFrameworkTransactionProvider.cs
iancooper: I'll check. Where they derive from RelationalDbTransactionProvider I will rename

src/Paramore.Brighter/CommandProcessor.cs
iancooper: The default is not the right approach; you should go ahead and pass it in. The usage of the default was a change, and it had unintended consequences. This worked for providers who were stateless, which was the primary motivation for the change; however, it has unintended consequences for stateful providers. This rolls it back to how it used to work.

iancooper: What should be happening is that whatever calls DepositPost, say a handler, should have the transaction provider injected into it by the IoC container, which is how we get the right scope

iancooper: Not passing a provider means: I don't want a transaction

iancooper: I think the samples are clear on how this should be done

Issue #3606, opened 2025-04-25
DynamoDB outbox sharding prevents guaranteed message ordering


dhickie:
**Is your feature request related to a problem? Please describe.**
When messages are added to the Dynamo DB outbox, they are assigned a random "shard" for their partition key, in order to avoid hot partitions for particular topics. When the sweeper queries the outbox for outstanding messages, it iterates through all the available shards. This makes it impossible to guarantee message ordering for any given partition key, as messages for that key will be randomly distributed between the different shards.

**Describe the solution you'd like**
Instead of assigning messages to shards randomly, use a deterministic hashing algorithm that distributes messages evenly between shards while guaranteeing that all messages for a particular partition key are assigned to a particular shard:

```
var keyBytes = Encoding.UTF8.GetBytes(message.Header.PartitionKey);
var sha256 = SHA256.Create();
var keyHash = sha256.ComputeHash(keyBytes);
var shardNumber = BitConverter.ToUInt32(keyHash, 0) % _configuration.NumberOfShards;
```

**Describe alternatives you've considered**
No other alternatives were considered. This would solve the problem at hand, at the cost of some CPU in order to calculate the hash.

**Additional context**
This is important for users using a Dynamo DB outbox with Kafka - if log compaction is enabled for Kafka then it's critical that messages are added to partitions in the order the deposits were originally made to the outbox.

iancooper:
Hi @dhickie do we think we can close this due to #3761

iancooper:
Hi @dhickie can we close?

iancooper:
Going to close, as I believe that the new methodology does a lot for this

PR #3761, opened 2025-09-03
Optimise Dynamo DB outbox usage

dhickie:


docs/adr/0033-optimise-reads-and-writes-from-dynamo-outbox.md.md
```diff
@@ -0,0 +1,91 @@
...
+
+Every time one or more messages are cleared from the outbox, the `OutboxProducerMediator` checks whether it needs to update it's internal metric for how many outstanding messages currently sit in the outbox, based on when that last check was last run. If it determines a refresh is neccesary, it fetches all outstanding messages from the outbox into memory (but does so asynchronously). This comes with all the queries described above, and if there are a large number of messages outstanding, can eat up a problematic amount of memory. The count of outstanding messages only appears to be used for monitoring purposes.
+
+## Decision
+
+All of the inefficiencies above can be improved with non-breaking changes.
+
+### `OutstandingMessages` operation
+
+Given that the `Outstanding` index is a sparse index, and we wish to pull out the entirety of that index when we perform the operation, this can be a `Scan` operation on the index instead of a `Query`. This removes the need to iterate over topics and shards, and can instead be done as a single HTTP call if the number of outstanding messages allows it (with paging if it doesn't).
```
iancooper: What would happen if the partition key (hash key) was the status instead of the topic (as today) i.e. Outstanding or Delivered? This would be sparse as the record would not exist the value was null. As ticks are numeric then hashing them ought to ensure that we had high cardinality to avoid hotspots on writes. Then the only items in the GSI ought to be the Outstanding or Delivered rows. You could then Scan and filter by Topic as, as you allude to here, you want all the results back for a topic anyway. Assuming that the Scan is no more painful than the Query that needs to get all rows anyway?

docs/adr/0033-optimise-reads-and-writes-from-dynamo-outbox.md.md
```diff
@@ -0,0 +1,91 @@
...
+
+The one downside of this is that we cannot specify the ordering of results from a `Scan` operation. If the results are paged, we will not be able to specify that the oldest messages should be retrieved first. Given the performance issues using `Query` operations, and the limitations of Dynamo DB as a storage platform, this feels like a reasonable comprompise to make.
```
iancooper: If we are retrieving all the outstanding items (without a limit), we can order them ourselves in memory and then dispatch them. That would preserve ordering. This also set us up to use a parallel scan to accelerate execution. It's relatively safe to read them all into a buffer, because if we fail, we don't dispatch them.

docs/adr/0033-optimise-reads-and-writes-from-dynamo-outbox.md.md
```diff
@@ -0,0 +1,91 @@
...
+
+Add overloads of both `Get` and `GetAsync` to the outbox interfaces that take a collection of message IDs instead of just one. For the Dynamo DB implementation, use a `BatchGet` operation to fetch all of them at once. For the other implementations, they can just iterate over their other `Get` methods for now.
+
+If the `BatchGet` operation only returns a subset of the requested messages, throw an exception.
+
+Update `OutboxProducerMediator` to use the new `Get` methods.
+
+### Marking messages as dispatched
+
+When marking a collection of messages as dispatched, us a `BatchWrite` operation to update all of them at once. If any of the updates fail, throw an exception.
```
iancooper: If we fail to mark dispatched in a batch, do we want to throw an exception, or log? In principle, an undispatched message would just be picked up again and resent. The main issue would be increased latency. Perhaps a boolean to indicate if you throw on an partial write, or just log, and the from a Sweeper we would just let it try again next time, but from an immediate write, you would choose by setting an arg

docs/adr/0033-optimise-reads-and-writes-from-dynamo-outbox.md.md
dhickie: Scans are only more painful than a Query if:
* You need to be able to sort results (since you need to fetch everything to sort in memory)
* You want to apply a filter that filters out a substantial number of records.

Filter expressions are applied _after_ the data has been read from the partition, but before the data is returned to the client. So while they reduce the data being sent over the wire, they don't reduce the number of read units being consumed or the amount of data that has to be read out of the table/index.

Using `OutstandingCreatedTime` and `DeliveredTime` as the hash key for the `Outstanding` and `Delivered` indices would certainly solve the cardinality problem, but could make things problematic for users who have a use case requiring retrieval of records for a specific topic. If, for example, an outbox was being used for one high throughput topic, and one low throughput topic, then attempts to get the outstanding messages on the low throughput topic would require reading everything from the high throughput topic as well.

Another thing to note is that the `limit` applied to both Query and Scan operations applies to the number of items read from the table, _not_ the number of items returned to the client. So if one was applying a filter expression that filters out the majority of a table, the operation may well return no items at all even if there _are_ items in the table that would be returned in subsequent pages.

That's why I left the partition structure in place here - it isn't as clean when it comes to cardinality, but is much cleaner for the use case where a user may only want outstanding/delivered messages for a particular topic.

docs/adr/0033-optimise-reads-and-writes-from-dynamo-outbox.md.md
dhickie: I think this is the way to go. Preserving ordering at the database level can only be done with a Query, and since we have requirements to both increase the cardinality of partition keys (via sharding) and reduce the number of requests sent over the network, Queries just don't work here.

The only potential issue is when there is a large number of messages in the Outstanding index, necessitating the use of paging to avoid issues with memory consumption or latency. Since the order of records returned by a Scan is essentially random, we just need to ensure it's clear somewhere in documentation or in comments that ordering of published messages is only guaranteed if `Number of outstanding messages < Page size`.

dhickie: This does in theory negate the need for #3606, but if we keep the existing index structure for the Outstanding and Delivered indices then I still think it's worth making shard assignment deterministic - it would at least mean users could preserve ordering even if there are a large number of messages in the Outstanding index by querying topics individually.

docs/adr/0033-optimise-reads-and-writes-from-dynamo-outbox.md.md
dhickie: Yeah I thought about this one for a while. I ended up with the suggestion to throw an exception because that would effectively reproduce the existing behaviour, but logging makes a lot of sense. I'm inclined to say:

* Default to throwing
* Have an optional bool in the `args` bag to allow logging instead
* Use the logging option when invoking the outbox from the Sweeper

docs/adr/0033-optimise-reads-and-writes-from-dynamo-outbox.md.md
iancooper: We could have both GSIs that let you query by topic and those that don't. However, if we end up paging in the records, as shown below, we should not expose a different way to handle high-throughput and low-throughput topics. If we do (and I missed it) or intend to (which, if we don't, might be a good idea), then we could explicitly use the GSI that uses the topic for the hash key if you expressly ask us to clear a given topic. We could note that the topic based version is preferred for clearing the low throughput topic if it's being drowned by a 'noisy neighbor'

docs/adr/0033-optimise-reads-and-writes-from-dynamo-outbox.md.md
iancooper: For convenience, I note above, that we could possibly have both, if we choose to have an optimal structure for a parallel scan, and then let you give us a specific index if noisy neighbor is a problem.

docs/adr/0033-optimise-reads-and-writes-from-dynamo-outbox.md.md
dhickie: Ah yes I hadn't considered effective parallel scanning. Having both makes sense. To avoid requiring people to add a new GSI if they're using v9, I'll update the ADR to reflect:

In v9:
- Scan the existing outstanding index and only do a single scan for now

In v10:
- Add a new GSI for outstanding messages indexed by `OutstandingCreatedTime` (with message ID as the range key for completeness)
- When scanning for all outstanding messages, use the new index with a config option for the number of parallel scans to perform at once
- When querying for all outstanding messages for a particular topic, use the old index

lillo42 (approved):

src/Paramore.Brighter.Outbox.DynamoDB/DynamoDbOutbox.cs
```diff
@@ -560,151 +613,251 @@ public async Task<IEnumerable<Message>> OutstandingMessagesAsync(
...
+                var response = await _client.ScanAsync(request, cancellationToken);
+
+                lastEvaluatedKey = response.LastEvaluatedKey;
+                segmentCount += response.Count;
+            } while (lastEvaluatedKey != null && lastEvaluatedKey.Keys.Any() && segmentCount < maxCount);
+
+            return segmentCount;
+        }
+
         private async Task<IEnumerable<Message>> OutstandingMessagesForAllTopicsAsync(TimeSpan dispatchedSince, int pageSize, int pageNumber, CancellationToken cancellationToken)
```
dhickie, line 701: Comment that this can only be done once at a time

docs/adr/0033-optimise-reads-and-writes-from-dynamo-outbox.md
```diff
@@ -0,0 +1,115 @@
...
+## Consequences
+
+* When performing the `OutstandingMessages` or `DispatchedMessages` operations for all topics, we will only be able to guarantee the order of the returned messages if the number of outstanding messages is less than the page size for the operation.
+* Shards will be assigned to messages deterministically based on their partition key
+* The possibility of future improvements to other outbox implementations, to take advantage of the new bulk operation methods
+* Users of the Dynamo DB outbox implementation in Brighter v9 will need to update their table as part of their migration to v10:
+    * Add a new GSI called `OutstandingAllTopics`, that uses `OutstandingCreatedTime` as its HASH key and `MessageId` as its RANGE key
+    * Add a new GSI called `DeliveredAllTopics`, that uses `DeliveryTime` as its HASH key and `MessageId` as its RANGE key
+    * Change the HASH key used by the `Delivered` index. This can be achieved by:
+        * Adding a _new_ GSI, which for the sake of example we'll call `DeliveredV10`, which uses `TopicShard` as its HASH key and `DeliveryTime` as its RANGE key
```
dhickie: DeliveredSharded?

iancooper (approved):
Approved as of meeting on Wed 1 Oct

PR #3772, opened 2025-09-10
fix: quartz sample

lillo42:
The issue was with the receiver, where we hadn't set up the SNS Topic as FIFO. I enjoy this moment to improve the sample, by:
- delaying the message by 10s (before it was 1 min)
- Removing the Program/main class
- Allowing to stop the sample by pressing CONTROL+C


iancooper (approved):

iancooper:
Thanks!

PR #3774, opened 2025-09-11
[Feature] Specification Driven Development

iancooper:
To test specification-driven development, we are making a few changes:

- Sharing instruction files for agents between Copilot and Claude.
- Adding a specs folder in which to put specs, for use by agents

We are then preparing to test using agents, driven by specifications, to implement new messaging gateways. 

- Writing up specs for a messaging gateway that encapsulates our interfaces and usual approach
- Writing up a spec for a Redis Streams gateway that references this, so we can prompt to create it

The principle here is to be able to reuse the specs for other gateways, thus amortizing the cost of writing the specs across multiple implementations.

If it works we can push the idea to inboxes and outboxes.


iancooper:
Moving to a fresh PR

Issue #3776, opened 2025-09-12
[Bug] Reactor mode isn't respect the empty dely


lillo42:
### Describe the bug
When the empty delay is set to reactor message pump, it isn't respected

### To Reproduce
Configure a subscription with a custom empty delay and reactor mode

### Further technical details
- Brighter version: 10.0.0-preview.9

PR #3777, opened 2025-09-12
[Bug] Fixes Reactor mode not respecting empty channel delay

lillo42:
fix: https://github.com/BrighterCommand/Brighter/issues/3776


iancooper (approved):
Thanks @lillo42

PR #3789, opened 2025-09-16
patch(Azure): Update Azure Identity and remove no longer supported providers

preardon:
Azure Identity have depricated some of Providers


iancooper (approved):
Thanks @preardon

PR #3827, opened 2025-10-02
feat: Add support to binary payload for inbox pattern

lillo42:
Add support to binary payload and `create table if not exists`


iancooper (approved):
Looks good. Thanks for adding all the tests too

PR #3837, opened 2025-10-06
feat: Register the brighter default transformer/mapper

lillo42:


iancooper (approved):
Makes sense as an approach. Assume we check that double registration won't happen, if so all fine

PR #3844, opened 2025-10-10
Migrate tests to xUnit V3

lillo42:


iancooper:
@lillo42 What is the status on this one?

lillo42:
@iancooper I'm not looking at this right now, I'm focusing more on creating shared tests

iancooper:
I picked it up under a separate piece of work to give us some tests around memory leaks, we can liaise when I merge, but I agree with your priority. We will get a lot of value out of the test generation plus agents

iancooper:
This is now in #3971 as I needed it there. It has everything we need.

iancooper:
Suggest we close this one @lillo42

PR #3846, opened 2025-10-11
fix: MongoDBUnitOfWork implement IAmABoxTransactionProvider<IClientSessionHandle>

lillo42:
During the MongoDB implementation, I have implemented the wrong interface for  `IAmAMongoDbTransactionProvider`, to avoid any breaking changes, I have implemented the correct one in `MongoDbUnitOfWork`


iancooper:
I may have to release 10.0.1 then :-D

PR #3858, opened 2025-10-14
docs: Corrects comment on default HandlerLifetime

Jonny-Freemarket:
Looking at upgrading our v9 -> v10 and noticed this comment doesn't align with the code. I've updated the comment to match the code, but can switch it if that's incorrect.


iancooper:
Thanks @Jonny-Freemarket

Issue #3864, opened 2025-10-17
[Bug] Brighter is ignoring custom `ResiliencePipelineRegistry`


lillo42:
### Discussed in https://github.com/BrighterCommand/Brighter/discussions/3862

<div type='discussions-op-text'>

<sup>Originally posted by **horsdal** October 16, 2025</sup>
How can I register and use resiliency pipelines?
I have the code below where I try to first register a couple of resiliency pipelines and then use one of them in a command handler via an attribute, but the it throws because the pipeline is not found while creating the command processor:
```
info: Paramore.Brighter.CommandProcessor[148105941]
      Building send pipeline for command: GreetCommand 0199ee3a-aa99-71f5-aed6-85e7e10b87b8
Unhandled exception. Paramore.Brighter.ConfigurationException: Error when building pipeline, see inner Exception for details
 ---> System.Collections.Generic.KeyNotFoundException: Unable to find a resilience pipeline associated with the key 'ImmediateRetry'. Please ensure that either the resilience pipeline or the builder is registered.
   at Polly.Registry.ResiliencePipelineProvider`1.GetPipeline(TKey key)
   at Paramore.Brighter.Policies.Handlers.ResilienceExceptionPolicyHandler`1.InitializeFromAttributeParams(Object[] initializerList) in /_/src/Paramore.Brighter/Policies/Handlers/ResilienceExceptionPolicyHandler.cs:line 74
   at Paramore.Brighter.HandlerFactory`1.CreateRequestHandler(IAmALifetime lifetime) in /_/src/Paramore.Brighter/HandlerFactory.cs:line 54
   at Paramore.Brighter.PipelineBuilder`1.<>c__DisplayClass21_0.<PushOntoPipeline>b__0(RequestHandlerAttribute attribute) in /_/src/Paramore.Brighter/PipelineBuilder.cs:line 395
   at Paramore.Brighter.Extensions.EnumerationExtensions.Each[T](IEnumerable`1 collection, Action`1 doThis) in /_/src/Paramore.Brighter/Extensions/Each.cs:line 43
   at Paramore.Brighter.PipelineBuilder`1.PushOntoPipeline(IEnumerable`1 attributes, IHandleRequests`1 lastInPipeline, IRequestContext requestContext, IAmALifetime instanceScope) in /_/src/Paramore.Brighter/PipelineBuilder.cs:line 390
[...]

PR #3865, opened 2025-10-17
fix: Resilience Pipeline was not propertly set

lillo42:
fix: https://github.com/BrighterCommand/Brighter/issues/3864


iancooper (approved):
Looks right to me

PR #3907, opened 2025-11-14
Add ServiceProvider-aware extension methods to ServiceCollectionExtensions

thomhurst:
This should allow for much more flexibility. 

Consumers can use their built services or options, or can utilise things like `.PostConfigure<TOptions>` and retrieve the `IOptions<TOptions>` from inside the lambda via the ServiceProvider.

The latter is actually what I'm wanting to do because I have a TestServer where I override some configuration dynamically, but without access to the service provider this is proving difficult.

### DIFF
--- a/src/Paramore.Brighter.Extensions.DependencyInjection/ProducersServiceCollectionExtensions.cs
+++ b/src/Paramore.Brighter.Extensions.DependencyInjection/ProducersServiceCollectionExtensions.cs
@@ -28,8 +28,6 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
-using Paramore.Brighter.CircuitBreaker;
-using Paramore.Brighter.Observability;
 using Polly.Registry;
 
 namespace Paramore.Brighter.Extensions.DependencyInjection
@@ -102,11 +100,7 @@ namespace Paramore.Brighter.Extensions.DependencyInjection
             brighterBuilder.Services.TryAddSingleton<IAmProducersConfiguration>(busConfiguration);
             brighterBuilder.ResiliencePolicyRegistry ??= new ResiliencePipelineRegistry<string>().AddBrighterDefault();
 
-            brighterBuilder.Services.TryAdd(new ServiceDescriptor(
-                typeof(IAmAnOutboxProducerMediator),
-                sp => BuildOutBoxProducerMediator(sp, transactionType, busConfiguration, brighterBuilder.ResiliencePolicyRegistry, outbox)
-                    ?? throw new ConfigurationException("Unable to create an outbox producer mediator; are you missing a registration?"),
-                ServiceLifetime.Singleton));
+            RegisterOutboxProducerMediator(brighterBuilder, transactionType, busConfiguration, outbox);
 
             return brighterBuilder;
         }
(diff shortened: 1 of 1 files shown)


src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
```diff
@@ -873,7 +873,7 @@ private static void RegisterLuggageStore<TStoreProvider>(this IServiceCollection
                 {
                     IAmAStorageProviderAsync store = provider.GetRequiredService<TStoreProvider>();
                     store.Tracer = provider.GetRequiredService<IAmABrighterTracer>();
-                    store.EnsureStoreExistsAsync().GetAwaiter().GetResult();
+                    store.EnsureStoreExistsAsync().ConfigureAwait(false).GetAwaiter().GetResult();
```
iancooper: Do you need ConfigureAwait here? You are overriding any context.

thomhurst: Reverted back to the original 😄

iancooper:
Hi @thomhurst. 

First, thanks for the contribution. Always appreciated. I have a few questions.

Can you explain the design here? An issue or ADR always helps. Is your goal here to add support for `IConfigure`? 

- It is worth noting that our use of BrighterOptions predates the adoption of `IOptions<TOptions>` by .NET
- Also worth noting is our use of a DI-Friendly Framework approach, which means that Brighter itself does not depend on a DI framework but on factories. We provide implementations of those factories for ServiceCollection as a convenience
- I note that you added a set of UseXXX methods. We moved away from these in favor of asking folks to set their values via BrighterOptions, to simplify our API. Before we add these back in, what do you want to achieve by adding them?

I am sure it will help, but want to help an old man understand what you are trying to achieve?

thomhurst:
> Hi @thomhurst.
> 
> First, thanks for the contribution. Always appreciated. I have a few questions.
> 
> Can you explain the design here? An issue or ADR always helps. Is your goal here to add support for `IConfigure`?
> 
> * It is worth noting that our use of BrighterOptions predates the adoption of `IOptions<TOptions>` by .NET
> * Also worth noting is our use of a DI-Friendly Framework approach, which means that Brighter itself does not depend on a DI framework but on factories. We provide implementations of those factories for ServiceCollection as a convenience
> * I note that you added a set of UseXXX methods. We moved away from these in favor of asking folks to set their values via BrighterOptions, to simplify our API. Before we add these back in, what do you want to achieve by adding them?
> 
> I am sure it will help, but want to help an old man understand what you are trying to achieve?

Chucked you a message on slack 😄

thomhurst:
Had to split some stuff out because quality gates were failing as too complex. Let me know any more changes you'd like to see

iancooper:
@thomhurst I have reopened this one, as I have time to look at it properly. Feel free to close it, and note that you don't want me to re-open and look at it.

- New Extension Methods. These should live in Paramore.Brighter.Extensions.DependencyInjection. I can see why you would want to add them to Paramore.Brighter in its Extensions namespace. However, we don't consider support for ServiceCollection to be part of our core library. You could use another DI framework. So these extensions should live in the above assembly. (I accept that it could be better named, but that would be a breaking change now).
- Options not being passed on. The main issue seems to be the options not beind passed down, and some missing registrations.

Let me know what you would like to do. My apologies for not having time for this earlier.

thomhurst:
> @thomhurst I have reopened this one, as I have time to look at it properly. Feel free to close it, and note that you don't want me to re-open and look at it.
> 
> * New Extension Methods. These should live in Paramore.Brighter.Extensions.DependencyInjection. I can see why you would want to add them to Paramore.Brighter in its Extensions namespace. However, we don't consider support for ServiceCollection to be part of our core library. You could use another DI framework. So these extensions should live in the above assembly. (I accept that it could be better named, but that would be a breaking change now).
> * Options not being passed on. The main issue seems to be the options not beind passed down, and some missing registrations.
> 
> Let me know what you would like to do. My apologies for not having time for this earlier.

That's okay. Maybe we can chat in the new year because I'm not sure this is even the right approach. I was diving into the codebase a bit more and I think things need restructuring a fair bit (properly defer things until service resolution) to achieve what I want - And this is just trying to workaround that currently. 

I've raised an issue with some more details just so nothing gets lost, and hopefully to give a bit more context around this:

#3951

iancooper:
Thanks @thomhurst I would love to chat. There is also the team meeting and you would be welcome to join that, so that you can get more than just my feedback

PR #3909, opened 2025-11-18
Create a generated base test suite

lillo42:
This PR is about creating a tool to generate a base test


iancooper:
@lillo42 This looks great. I didn't spot it, but it might be useful in the ADR to show a structure for a test project that separated generated and custom (user defined tests) like

Paramore.Brighter.MessagingGateway.X.
|
- Generated
- MessagingGatewaySpecific

To show that we would reserve the write to overwrite the Generated folder on an update, but keep the MessagingGatewaySpecific folder intact i.e. your tests

I guess once we generate an Outbox it would become OutboxSpecific etc.

PR #3910, opened 2025-11-18
Add entity framework transaction provider for MongoDB

### LINKED ISSUE #3879: [Feature] Add Transaction provider for MongoDB EntityFramework
Since we already have official entity framework provider for MongoDB (https://github.com/mongodb/mongo-efcore-provider) - we should add new package to Brighter so we can use it as a transaction provider.

I can create a PR for that if it's fine for you

jakoss:
Fixes #3879


src/Paramore.Brighter.MongoDb.EntityFramework/MongoDbEntityFrameworkTransactionProvider.cs
```diff
@@ -0,0 +1,190 @@
...
+        protected void Dispose(bool disposing)
+        {
+            if (_disposed)
+            {
+                return;
+            }
+    
+            if (disposing)
+            {
+                _connection?.Dispose();
```
jakoss: @iancooper I'm not sure if we should dispose the connection and transaction in here, those should be managed by the DbContext itself i guess? This is a copy of postgres ef transaction provider

src/Paramore.Brighter.MongoDb.EntityFramework/MongoDbEntityFrameworkTransactionProvider.cs
```diff
@@ -0,0 +1,190 @@
...
+        {
+            var tcs = new TaskCompletionSource<IClientSessionHandle>();
+    
+            if (cancellationToken.IsCancellationRequested)
+            {
+                tcs.SetCanceled(CancellationToken.None);
+            }
+    
+            tcs.SetResult(GetTransaction());
+            return tcs.Task;
```
lillo42: Why not?
```suggestion
        {
            return Task.FromResult<IClientSessionHandle>(GetTransaction());
```

src/Paramore.Brighter.MongoDb.EntityFramework/MongoDbEntityFrameworkTransactionProvider.cs
```diff
@@ -0,0 +1,190 @@
...
+        /// <summary>
+        /// Get the current connection of the database context
+        /// </summary>
+        /// <param name="cancellationToken">A cancellation token</param>
+        /// <returns></returns>
+        public Task<DbConnection> GetConnectionAsync(CancellationToken cancellationToken = default)
+        {
+            var tcs = new TaskCompletionSource<DbConnection>();
+            tcs.SetResult(_context.Database.GetDbConnection());
+            return tcs.Task;
```
lillo42: Why not?
```suggestion
            return Task.FromResult(_context.Database.GetDbConnection());
```

src/Paramore.Brighter.MongoDb.EntityFramework/MongoDbEntityFrameworkTransactionProvider.cs
jakoss: This is just an early draft to show the issue with internal api and mismatch between translation type between the Mongo outbox provider and the Ef provider. I'll refractor that when we agree on the solution

src/Paramore.Brighter.MongoDb.EntityFramework/MongoDbEntityFrameworkTransactionProvider.cs
iancooper: It depends on whether we create it. As we don't, then I assume we would not. The question that arises is whether we might actually be creating the transaction, but I would assume not

iancooper:
I would say that this looks right. Let's have a stab at it

jakoss:
> I would say that this looks right. Let's have a stab at it

Ok, so i'll clean this up a little bit and i'll create a ticket to MongoDB EF Provider team so they maybe consider making the Session property public

jakoss:
@iancooper @lillo42 ready for review

jakoss:
I think our claude friend hallucinated a little in the last review, it started talking about some mysql bug. Overall the reliance on the internal property is bad, i'll create a ticket to mongo ef team to maybe find a solution for that

iancooper:
> I think our claude friend hallucinated a little in the last review, it started talking about some mysql bug. Overall the reliance on the internal property is bad, i'll create a ticket to mongo ef team to maybe find a solution for that

Yeah, as we iterate it starts to be less useful.

iancooper:
Happy to merge @jakoss if you are ready?

iancooper:
Then I can do a build and get a few fixes/features out

jakoss:
@iancooper Sure, i'm done with the implementation :)

tests/Paramore.Brighter.MongoDb.Tests/EntityFramework/MongoDbEntityFrameworkTransactionProviderTest.cs
```diff
@@ -0,0 +1,284 @@
...
+using Xunit;
+
+namespace Paramore.Brighter.MongoDb.Tests.EntityFramework;
+
+[Trait("Category", "MongoDb")]
+[Trait("Feature", "EntityFramework.TransactionProvider")]
+public class MongoDbEntityFrameworkTransactionProviderTest
+{
+    [Fact]
+    public void When_Creating_Provider_With_DbContext_Should_Initialize_Correctly()
```
iancooper, line 17: Not a blocker, but our style is generally to avoid mocks, unless there are speed or shared fixture concerns. Instead, we prefer to work through the SUT's interface or use a fully featured in-memory drop-in. This is to avoid coupling tests to details over behaviors.

But my usual strategy is to get this in, get it working once we release it, and then address issues of style/approach

jakoss: Ideally i would do integrations tests, since those unit tests doesn't really test much, not in the context of brighter. But so far i had no time to setup such test

iancooper: No worries. Let's take, I can build, you can see if it works, and then we can get those tests in later.

lillo42 (approved):

PR #3912, opened 2025-11-20
Use ConcurrentDictionary In InMemoryLock to avoid race conditions

jakoss:
I had random errors on service startup with Outbox setup that multiple code paths tried to get the lock from InMemoryLock. Standard dictionary does not like this, so i migrated to Concurrent one


iancooper:
I'll see about a new release at the weekend, just one to pick up some other stragglers.

PR #3913, opened 2025-11-21
[Fix] Allow use of Trace for databases that are not in DbSystem

lillo42:
Allow use of Trace for databases that are not in `DbSystem`, like `Tablestore`, `GaussDB` and others


iancooper:
@lillo42 I think we should action the Claude feedback on this one

PR #3916, opened 2025-11-24
Add support to configuration brighter from IConfiguration

lillo42:
Add support to read Brighter configuration from IConfiguration and aspire configuration


iancooper:
At this point, I am just reviewing the ADR, and it looks good. It strikes the right balance between code vs. configuration. Aspire is the tipping point here, as it would be good to grab the relevant configuration from Aspire, which would shift the balance toward the "implementation left to the reader" approach we had before.

Hopefully we can generate useful docs from this too.

iancooper:
Going to review because we want to pull from IConfiguration to populate our code based config, not create custom config sections

PR #3918, opened 2025-11-28
Add Universal Support for Dead Letter Channels

iancooper:
Not all of our transports support the idea of a Dead Letter Channel or an Invalid Message Channel. This forces folks into workarounds with a Fallback Handler to post to a "DLQ". This is particularly egregious for folks using streams (as queues).

This PR adds support for moving a message to an Invalid Message Channel if there are serialization failures, and to a Dead Letter Channel if we fail to process it.

We still want to strongly advise against an "error queue" as opposed to use for moving ill-configured or blocking messages out of the way, but this will simplify the situation for folks

### DIFF
--- a/tests/Paramore.Brighter.Core.Tests/MessageDispatch/Reactor/When_an_event_handler_throw_a_reject_message_exception.cs
+++ b/tests/Paramore.Brighter.Core.Tests/MessageDispatch/Reactor/When_an_event_handler_throw_a_reject_message_exception.cs
@@ -1,6 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
-using System.Diagnostics.CodeAnalysis;
+using System.Threading.Tasks;
 using Microsoft.Extensions.Time.Testing;
 using Paramore.Brighter.Core.Tests.MessageDispatch.TestDoubles;
 using Paramore.Brighter.Extensions;
@@ -20,10 +20,12 @@ public class MessageDispatchRejectMessageExceptionTests
 
     public MessageDispatchRejectMessageExceptionTests()
     {
+        MyRejectedEventHandler.Reset();
+
         var subscriberRegistry = new SubscriberRegistry();
         subscriberRegistry.Register<MyRejectedEvent, MyRejectedEventHandler>();
-        
-        var handlerFactory = new SimpleHandlerFactory( 
+
+        var handlerFactory = new SimpleHandlerFactory(
             (type) => new MyRejectedEventHandler(),
             (type) => throw new ConfigurationException()
         );
@@ -31,11 +33,11 @@ public class MessageDispatchRejectMessageExceptionTests
         var mapperFactory = new SimpleMessageMapperFactory((r) => new MyRejectedEventHandlerMessageMapper());
         var messageMapperRegistry = new MessageMapperRegistry(mapperFactory, null, null, null);
         messageMapperRegistry.Register<MyRejectedEvent, MyRejectedEventHandlerMessageMapper>();
-            
+
 
         var resiliencePipelineRegistry = new ResiliencePipelineRegistry<string>();
         resiliencePipelineRegistry.AddBrighterDefault();
-        
+
         var commandProcessor = new CommandProcessor(
             subscriberRegistry,
             handlerFactory,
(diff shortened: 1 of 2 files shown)


docs/adr/0034-provide-dlq-where-missing.md
```diff
@@ -0,0 +1,60 @@
...
+
+When load-shedding, because a stream does not delete an undeliverable message, but instead marks it's offset as read, it is possible to go back and look at the record (until its TTL). Thus, theoretically, records can be reviewed in their original stream so a dead-letter channel would be redundant.
+
+Why then would we need a Dead Letter Channel for a stream?
+
+Normally the answer is process, with monitoring of Dead Letter Channels a process that a team adheres too. Reviewing the channel, because it has more than zero records is easier than looking for the offset of records that failed in the logs, and examining the instead. Replaying from the channel is easier than reprocessing the record in its original position in the stream.
```
jakoss: I'd add that the reviewing manually is one thing, for us the best thing is that we have automatic alerts set to any DLQ in our system. If there is a DLQ and it has entries - we have red lights in our dashboard and inboxes

iancooper: Thanks, that's helpful

src/Paramore.Brighter.ServiceActivator/IAmAMessagePump.cs
```diff
@@ -58,5 +58,10 @@ public interface IAmAMessagePump
         /// This method will block and run continuously until a quit message is received or an unrecoverable error occurs.
         /// </remarks>
         void Run();
+        
+        /// <summary>
+        /// The <see cref="MessagePumpStatus"/> of the pump
+        /// </summary>
+        public MessagePumpStatus Status { get; set; }
```
lillo42:
```suggestion
       MessagePumpStatus Status { get; set; }
```

src/Paramore.Brighter.ServiceActivator/MessagePump.cs
```diff
@@ -56,6 +76,7 @@ public abstract partial class MessagePump
         protected int UnacceptableMessageCount;
         protected readonly Dictionary<Type, MethodInfo> UnWrapPipelineFactoryCache = new();
         protected readonly Dictionary<Type, MethodInfo> DispatchMethodCache = new();
+        protected DateTimeOffset? UnacceptableMessageWindowOpenedA = null;
```
lillo42, line 79: Seems it's missing a `t`

```csharp
        protected DateTimeOffset? UnacceptableMessageWindowOpenedAt = null;
```

src/Paramore.Brighter/Extensions/CharacterEncodingExtensions.cs
```diff
@@ -37,6 +49,20 @@ public static class CharacterEncodingExtensions
             _ => null
         };
 
+    /// <summary>
+    /// Turns a string into a character encoding
+    /// We support:
+    /// 
+    ///  "us-ascii" => CharacterEncoding.ASCII,
```
lillo42:
```suggestion
    /// "us-ascii" => CharacterEncoding.ASCII,
```

src/Paramore.Brighter/Extensions/ResiliencePipelineRegistryExtensions.cs
```diff
@@ -1,11 +1,59 @@
...
 
 public static class ResiliencePipelineRegistryExtensions
 {
+    /// <summary>
+    /// Adds default strategies to the <see cref="ResiliencePipelineRegistry{TKey}"/> needed by Brighter internally to
+    /// ensure the resilience of calls, such as to a <see cref="IAmAMessageProducer"/>
+    ///
+    /// We add:
+    ///
+    ///  CommandProcessor.OutboxProducer => new RetryStrategyOptions
```
lillo42:
```suggestion
    /// CommandProcessor.OutboxProducer => new RetryStrategyOptions
```

src/Paramore.Brighter/InMemoryMessageConsumer.cs
```diff
@@ -99,7 +106,15 @@ public async Task AcknowledgeAsync(Message message, CancellationToken cancellati
     /// </summary>
     /// <param name="message">The message.</param>
     public bool Reject(Message message)
-        => _lockedMessages.TryRemove(message.Id, out _);
+    {
+        var removed =_lockedMessages.TryRemove(message.Id, out _);
```
lillo42:
```suggestion
        var removed = _lockedMessages.TryRemove(message.Id, out _);
```

src/Paramore.Brighter/InMemorySubscription.cs
```diff
@@ -0,0 +1,96 @@
+﻿using System;
```
lillo42, line 1: Missing MIT headers

src/Paramore.Brighter.ServiceActivator/MessagePump.cs
iancooper: After all the difficulty getting "acceptable" right, imagine struggling on "at" :-D

iancooper:
@jakoss Soon. I need to fix a test that only fails on the server (weird as its a developer test). Then I can merge. I will follow up with middleware that will allow you to catch any exception at the outer layer and reject the message in response.

I have other middleware to support, but this should get you Kafka.

I may fix the lack of Kafka integration to our new scheduler for deferred requests,  before I turn to the other universal DLQ middleware. Kafka is the one most folks want

iancooper:
## Addressed PR Review Feedback

Fixed the issues identified in the review:

### ✅ CRITICAL: Disposal Logic Bug Fixed
The nested `if` statements in `Dispose()` that caused resource leaks have been fixed. All four producer disposal checks are now independent:
- `_deadLetterProducer`
- `_invalidMessageProducer`
- `_deadLetterProducerAsync`
- `_invalidMessageProducerAsync`

Each producer is now disposed regardless of whether other producers were created.

### ✅ Async Producer Comment Added
Added clarifying comment explaining that `KafkaMessageProducer` implements both `IAmAMessageProducerSync` and `IAmAMessageProducerAsync` interfaces.

### Already Implemented (Verified)
- **UnacceptableMessageLimit**: The `<= 0` check was already implemented in both `Reactor.cs:511` and `Proactor.cs:518`
- **Header constants**: All constants (`REJECTION_REASON`, `ORIGINAL_TOPIC`, `REJECTION_TIMESTAMP`, etc.) are defined in `HeaderNames.cs`
- **RejectMessageAction documentation**: The remarks section already contains meaningful content

Commit: f08b35b7b

PR #3920, opened 2025-12-04
Update tickerq to version 8

AboubakrNasef:
Some breaking changes for the update from TickerQ Side  
but to note:
- the minimum time is above 1 second if schedule with 1 second this is as start immediate so i've changed the tests to be at least 2 sec 
- the source generator has a [ModuleInitializer] attribute so it only called once and this does clear some caches form the inside 
so the first test class will run but the second will throw null exception so i've made a Singleton like class to build TickerQ functions  once 

- the update function only works if the ticker is idle or queue you also they removed the update by id from the manager so i've added the persistence provider to get the entity

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
             8.0.x
             9.0.x
       - run: dotnet --info
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 0
       - uses: actions/cache@v4
@@ -91,7 +91,7 @@ jobs:
     needs: [build]
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - name: Setup dotnet
         uses: actions/setup-dotnet@v5
         with:
@@ -109,7 +109,7 @@ jobs:
     needs: [ build ]
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - name: Setup dotnet
         uses: actions/setup-dotnet@v5
         with:
@@ -127,7 +127,7 @@ jobs:
     needs: [ build ]
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - name: Setup dotnet
         uses: actions/setup-dotnet@v5
         with:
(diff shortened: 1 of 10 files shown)


AboubakrNasef:
@iancooper  
i've managed to make the pr sorry if you have started on this 😅

iancooper:
>i've managed to make the pr sorry if you have started on this 😅
No worries, I was struggling to get the tests to pass. It was no harm for me to dig into this, as it upskilled me a little, which makes it easier to review your PR

src/Paramore.Brighter.MessageScheduler.TickerQ/TickerQScheduler.cs
```diff
@@ -40,7 +42,8 @@ namespace Paramore.Brighter.MessageScheduler.TickerQ
     /// <param name="getOrCreateSchedulerId">Function to generate or retrieve a scheduler identifier </param>
     /// <param name="parseSchedulerId">Function to parse a string scheduler ID into a Guid</param>
     public class TickerQScheduler(
-        ITimeTickerManager<TimeTicker> timeTickerManager,
+        ITimeTickerManager<TimeTickerEntity> timeTickerManager,
+        ITickerPersistenceProvider<TimeTickerEntity, CronTickerEntity> tickerPersistenceProvider,
```
iancooper, line 46: Yes, I noted that this was one big change, as any reschedule now needed to manually get the job

tests/Paramore.Brighter.TickerQ.Tests/TestDoubles/Fixtures/BaseTickerQFixture.cs
```diff
@@ -95,11 +93,13 @@ protected BaseTickerQFixture()
             });
 
             CommandProcessor.ClearServiceBus();
+
             ServiceProvider = _serviceCollection.BuildServiceProvider();
             Processor = ServiceProvider.GetRequiredService<IAmACommandProcessor>();
             SchedulerFactory = ServiceProvider.GetRequiredService<TickerQSchedulerFactory>();
-            var myHost = new MyHost() { Services = ServiceProvider };
-            myHost.UseTickerQ();
+            TickerQModuleInitializer.EnsureOneTimeSetupTickerQ();
```
iancooper, line 100: Ah, thanks. This is what I was missing, which was causing me to fail

iancooper (approved):

iancooper:
Going to merge, and then I can add this into the build

PR #3928, opened 2025-12-08
Fixes Kafka Sweeper

lillo42:
Currently, the Kafka Sweeper runs only after the Ack method. However, in certain situations, this fails to trigger the Sweeper, causing messages to remain uncommitted for hours.

This fix ensures that the Sweeper strictly respects the _sweepUncommittedInterval configuration, preventing these delays


iancooper:
Good spot. The lack of a new Ack would prevent this from triggering.

iancooper:
@lillo42 I am guessing we need this for V10 too. I am just about to do a V10 build. It feels like we should get that in tor this too

PR #3943, opened 2025-12-15
Migrate outbox test to  generated

lillo42:
In PR I'll migrate all existing outbox test, to the generated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,7 @@ jobs:
           dotnet-version: |
             8.0.x
             9.0.x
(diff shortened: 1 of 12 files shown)


iancooper:
This looks really nice @lillo42.  I assume we have a README.MD that tells folks how to make this work. We don't want to end up in the situation that you are the only one who can generate the test harness for a new Db or update it. I think that document - what process do you use is critical to us being to make this generated approach maintainable.

But great work in getting us this far.

iancooper:
We will probably need to tell Claude that test files don't need MIT headers :-D

Issue #3902, opened 2025-11-07
[Feature] RabbitMQ Mutual TLS


dssldavidmorgan:
Our Rabbit server is set up to require a client certificate but there appears not to be an interface in Brighter for providing one.

Some interface to provide it in the RmqMessagingGatewayConnection object would be nice.

I'll be working on an implementation for our case next week, I'm hoping someone on your side gets there before I do :)

Thank you.

lillo42:
Thank you @dssldavidmorgan if you need any help  shout to us :)

dssldavidmorgan:
That was to be my plan but instead we decided simply not to use mTLS for this case.

PR #3946, opened 2025-12-20
ADR: Optional RabbitMQ mutual TLS support

darrenschwarz:
Proposes an ADR for adding optional, opt-in mutual TLS (mTLS) support to RabbitMQ messaging gateways.

This PR is intentionally ADR-only to align on scope, constraints, and non-goals before implementation.

Related issue: #3902.

### DIFF
--- a/docs/adr/00xx-rabbitmq-mutual-tls.md
+++ b/docs/adr/00xx-rabbitmq-mutual-tls.md
@@ -107,7 +107,7 @@ These may be revisited in future ADRs if requirements evolve.
 
 - Brighter Contributing Guide (`Contributing.md`)
 - Existing Architecture Decision Records under `docs/adr`
-- Issue #3902: RabbitMQ Mutual TLS support
+- Issue #3902: RabbitMQ Mutual TLS support (https://github.com/BrighterCommand/Brighter/issues/3902)
 - PATH-210X extracted Critical Review Guidelines for Brighter
 
 ---


docs/adr/00xx-rabbitmq-mutual-tls.md
```diff
@@ -0,0 +1,116 @@
...
+
+mTLS is increasingly required in regulated, zero-trust, and defence-in-depth environments, where network-level trust is insufficient and explicit client authentication is mandated. Users integrating Brighter into such environments currently need to rely on external tooling, custom forks, or unsupported configuration paths to achieve this.
+
+Introducing mTLS support affects:
+- connection semantics for RabbitMQ gateways
+- user-facing configuration surfaces
+- security posture and operational expectations
+- testing strategy (particularly around transport-level behaviour)
+
+Under Brighter’s contribution guidelines, changes that introduce new capabilities or alter connection semantics are expected to be documented via an Architecture Decision Record (ADR) prior to implementation.
```
iancooper, line 21: Yep. Thanks for this.

docs/adr/00xx-rabbitmq-mutual-tls.md
```diff
@@ -0,0 +1,116 @@
...
+### Positive
+
+- Enables Brighter adoption in zero-trust and regulated environments
+- Provides a clear, supported path for RabbitMQ mTLS configuration
+- Preserves backwards compatibility and existing behaviour
+- Aligns with Brighter’s architectural principles and contributor guidelines
+
+### Negative
+
+- Increases configuration surface area for RabbitMQ gateways
```
iancooper, line 80: I think this is alright, as its opt in, so only you pay the cost if you take the benefit

docs/adr/00xx-rabbitmq-mutual-tls.md
```diff
@@ -0,0 +1,116 @@
...
+- Requires careful testing across sync/async variants
+- Introduces additional security-related documentation and support burden
+
+## Testing Strategy
+
+The implementation will be supported by:
+
+- **Unit tests** verifying correct propagation of mTLS configuration into the RabbitMQ connection factory
```
iancooper, line 88: Generally we tend to think about developer tests and acceptance tests, so the unit/integration distinction is a little spurious for us. For a transport, we tend to test against the actual transport, configured in Docker. (The core engine is tested against in-memory versions). So you really only need to extend the RMQ tests for this, and you can use Docker containers. Your issue might be that you need to configure Docker differently for mTLS, so other tests would fail. In that case, use an attibute for tests that use mTLS so that we can run you in GHA against a different container more easily.

docs/adr/00xx-rabbitmq-mutual-tls.md
```diff
@@ -0,0 +1,116 @@
...
+
+Brighter will introduce **optional, opt-in mutual TLS (mTLS) support** for RabbitMQ messaging gateways, with the following constraints:
+
+1. mTLS support is **fully optional** and **disabled by default**.
+2. Existing non-TLS and TLS-only configurations will continue to function unchanged.
+3. mTLS configuration will be expressed explicitly via RabbitMQ gateway configuration objects.
+4. Support will be implemented with **parity across synchronous and asynchronous RabbitMQ gateways**.
+5. No changes will be made to existing message wire formats, headers, or CloudEvents propagation.
+
+The primary mechanism for enabling mTLS will be the ability to supply an `X509Certificate2` (and associated trust configuration) to the RabbitMQ connection factory in a controlled, backwards-compatible manner.
```
iancooper, line 34: One note. We continue to support both RMQ v6 as it is sync, alongside RMQ v7, which is async. For now, we should try to support pushing this option to both versions.

iancooper:
Hi @darrenschwarz  -

Thanks for the ADR.

A few comments, mostly giving you situational awareness of things that will help you. Any questions, do shout.

darrenschwarz:
> Hi @darrenschwarz -
> 
> Thanks for the ADR.
> 
> A few comments, mostly giving you situational awareness of things that will help you. Any questions, do shout.

Hey, This is the only comment I can see?

docs/adr/00xx-rabbitmq-mutual-tls.md
lillo42, line 1: Can you add a number for this ADR?

iancooper:
No,an ADR is always good and in our guidelines. Some of you, no names, should be learning from this ;-)

ADRs are not only good to record design decisions, but we can point agents to them for docs or building a plan.

docs/adr/00xx-rabbitmq-mutual-tls.md
darrenschwarz: Will do.

PR #3952, opened 2025-12-23
feat: Instance-based mediator, DI function overloads, and deferred resolution for testability (parallelisation, overrides, etc.)

### LINKED ISSUE #3951: [Feature] Add service provider lambda overloads and Options pattern support to DI extensions for testability

  The current DI extension methods (AddBrighter(), AddProducers(), AddConsumers()) configure everything eagerly at registration time, with no way to defer decisions to resolution time or override configuration later. This creates several limitations:

  1. No service provider access during configuration - I can't do things like:
  services.AddBrighter(opts =>
  {
      // Can't access other services here!
      var someService = sp.GetRequiredService<IMyService>();
      opts.RequestContextFactory = new MyFactory(someService);
  });
  2. Can't use PostConfigure<T> for test overrides - The options classes (BrighterOptions, ConsumersOptions, ProducersConfiguration) aren't integrated with Microsoft's Options pattern. So I can't do:
  // In my test setup
  services.PostConfigure<BrighterOptions>(opts =>
  {
      opts.HandlerLifetime = ServiceLifetime.Scoped;
  });
  3. Static state in CommandProcessor prevents test parallelization - The s_mediator static field means all tests share state. This requires:
    - Using [Collection("CommandProcess")] to serialize tests instead of running them in parallel
    - Manually calling CommandProcessor.ClearServiceBus() in every test's Dispose
    - Sequential test execution even when tests are otherwise independent
  4. The intermediate provider is built and immediately disposed - Looking at BrighterHandlerBuilder():
  using var intermediateProvider = services.BuildServiceProvider();
  var options = optionsFunc(intermediateProvider);
  // Provider is disposed here - can't use it for anything else
  4. This means I can't even access services registered by other extensions.

  1. Add Func<IServiceProvider, T> overloads to all the `.AddXXX(...)` extension methods
  2. Integrate with the Options pattern - so IConfigureOptions<T> and IPostConfigureOptions<T> work
  3. Defer configuration to resolution time where possible
  4. Eliminate static state - instance-scope or service-provider-scope for test parallelization

  - Collection-based serialization with [Collection("CommandProcess")]
  - Manual ClearServiceBus() calls in test disposal
  - Isolated ServiceCollections per test (doesn't fix static field problem)

  **Additional context**

  The codebase already uses deferred instantiation in some places (BuildCommandProcessor(), BuildOutBoxProducerMediator()). The pattern exists, it's just not applied consistently to the configuration phase.

I'm currently writing a blog post on some testing techniques - Which may give you more context into what I'm doing. I'll try to remember to post it here after so you can get more of an idea.

Edit: Here's what I've been doing recently - That Brighter is blocking me from fully utilising:

https://medium.com/@thomhurst/per-test-isolation-in-asp-net-core-a-tunit-aspnetcore-guide-ce09f7d4a05f?postPublishedType=initial

thomhurst:
## Summary

I created a new ADR for traceability and a history. I've just set the status to propsal for now since this wasn't pre-agreed or anything.

This PR implements ADR 0034: Instance-based Mediator for Testability, adds `Func<IServiceProvider, T>` overloads for DI registration, and implements deferred resolution - enabling parallel test execution without `[Collection]` serialization.

For why this was necessary and the problems that occur without these changes, please read #3951 

### Key Changes

   **1. CommandProcessor Static→Instance Refactoring**
   - Convert `s_mediator` static field to `_mediator` instance field
   - Convert `s_transactionType` to `_transactionType` instance field
   - Mark `ClearServiceBus()` as obsolete (still functional for backward compatibility)
   - Preserve static reflection caches (`s_boundDepositCalls` etc.) as they're stateless and safe to share

   **2. Service Provider Function Overloads with Deferred Resolution**
   - Add `Func<IServiceProvider, BrighterOptions>` overload to `AddBrighter`
   - Add `Func<IServiceProvider, ProducersConfiguration>` overload to `AddProducers`
   - Add `Func<IServiceProvider, ConsumersOptions>` overload to `AddConsumers`
   - Enable runtime service resolution during configuration (deferred until service provider is built)

   **3. Fix Intermediate Service Provider Disposal**
   - Previously, `BrighterHandlerBuilder` built an intermediate `IServiceProvider` to resolve options during registration
   - This intermediate provider was disposed prematurely, causing issues with scoped services
   - Refactored to defer all resolution until the final service provider is built
   - All options, factories, and registries now resolve lazily via `Func<IServiceProvider, T>` delegates

   **4. Thread Safety Fixes in ServiceProviderHandlerFactory**
   - **Singleton caching bug**: Singleton mode wasn't caching instances - now uses `Lazy<T>` with `ConcurrentDictionary` to guarantee only one instance ever created
   - **Scoped race condition**: Check-then-act pattern caused duplicate instances under concurrent access - now uses `GetOrAdd` with factory delegate

   **5. Simplified Registry Architecture**
   - Removed lifetime parameters from `ServiceCollectionSubscriberRegistry`, transformer, and mapper registries
   - Registries always register handlers as Transient in DI container
   - Actual lifetime (Singleton/Scoped/Transient) managed at runtime by `ServiceProviderHandlerFactory`
   - Centralizes lifetime logic in one place instead of scattered across registries

   **6. Options Pattern Integration**
   - Integrate `Microsoft.Extensions.Options` for `PostConfigure<BrighterOptions>` support
   - Enable test-time overrides via standard .NET pattern

   **7. Factory-Managed Lifetimes**
   - Implement proper Singleton/Scoped/Transient lifetime management in `ServiceProviderHandlerFactory`
   - Thread-safe singleton caching with `Lazy<T>` and `ConcurrentDictionary`
   - Thread-safe scoped caching with `GetOrAdd` pattern
   - Proper scope disposal and cleanup

   **8. Comprehensive Test Coverage**
   - `FactoryLifetimeTests.cs` - 11 tests for all lifetime modes (sync + async)
   - `FactoryThreadSafetyTests.cs` - 4 tests for concurrent resolution
   - `CommandProcessorIsolationTests.cs` - 2 tests for instance isolation
   - `FactoryErrorHandlingTests.cs` - 4 tests for error scenarios
   - `ServiceProviderLambdaTests.cs` - Tests for new lambda overloads

   ### Benefits

   - **Parallel Test Execution**: Tests can run in parallel without `[Collection("CommandProcessor")]` serialization
   - **Test Isolation**: Each test can have isolated `CommandProcessor` with its own mediator
   - **Deferred Resolution**: Configuration can access services registered later in the DI container
   - **No Premature Disposal**: Intermediate service provider no longer built/disposed during registration
   - **Thread Safety**: Race conditions fixed - concurrent handler resolution safe for all lifetime modes
   - **Simplified Architecture**: Lifetime management centralized in factory, not scattered across registries
   - **Standard Patterns**: Uses standard .NET Options pattern for test overrides
   - **Backward Compatibility**: All existing APIs unchanged, `ClearServiceBus()` still works

   ### Migration

   **Existing code continues to work unchanged:**
   ```csharp
   // Still works
   services.AddBrighter(options => options.HandlerLifetime = ServiceLifetime.Scoped);
   ```

   **New patterns enabled:**
   ```csharp
   // Service Provider Function Overload with Deferred Resolution (NEW)
   services.AddBrighter(sp => new BrighterOptions
   {
       RequestContextFactory = sp.GetRequiredService<IAmARequestContextFactory>()
   });

   // Test Override via Options Pattern (NEW)
   services.AddBrighter(o => o.HandlerLifetime = ServiceLifetime.Scoped);
   services.PostConfigure<BrighterOptions>(o => o.RequestContextFactory = testFactory);
   ```

   ## Test Plan

   - [x] All 55 Paramore.Brighter.Extensions.Tests pass
   - [x] All 495 Paramore.Brighter.Core.Tests pass
   - [x] All 90 Paramore.Brighter.InMemory.Tests pass
   - [x] Thread safety tests with barrier synchronization verify concurrent access
   - [x] Singleton tests verify only one instance ever created under contention
   - [x] Backward compatibility tests verify existing API patterns work

The static singleton would actually cause a bug if two command processors were ever instantiated in the same app, which was the root cause of the test isolation issues:

  Original Code (Bug):
```csharp
  private static IAmAnOutboxProducerMediator? s_mediator;
  private static readonly object s_padlock = new();

  private static void InitExtServiceBus(IAmAnOutboxProducerMediator bus, Type? transactionType)
  {
      if (s_mediator == null)  // If already set, silently ignore new bus!
      {
          lock (s_padlock)
          {
              if (s_mediator == null)
              {
                  s_mediator = bus;  // Only first caller wins
                  s_transactionType = transactionType ?? typeof(CommittableTransaction);
              }
          }
      }
      // BUG: If s_mediator was already set, the 'bus' parameter is silently ignored!
  }
```

  The Problem:
```csharp
  // Create first CommandProcessor with outbox1
  var outbox1 = new InMemoryOutbox();
  var bus1 = new OutboxProducerMediator(..., outbox1);
  var cp1 = new CommandProcessor(..., bus1, ...);  // Sets static s_mediator = bus1

  // Create second CommandProcessor with outbox2
  var outbox2 = new InMemoryOutbox();
  var bus2 = new OutboxProducerMediator(..., outbox2);
  var cp2 = new CommandProcessor(..., bus2, ...);  // bus2 SILENTLY IGNORED!

  // User expects cp2 to use outbox2, but...
  cp2.DepositPost(myCommand);  // Actually writes to outbox1, not outbox2!
```

  New Code (Fixed):
```csharp
  private readonly IAmAnOutboxProducerMediator? _mediator;
  private readonly Type _transactionType;

  // In constructor - direct assignment, no silent ignoring
  public CommandProcessor(..., IAmAnOutboxProducerMediator bus, Type? transactionType, ...)
  {
      _mediator = bus;  // Each instance uses what you give it
      _transactionType = transactionType ?? typeof(CommittableTransaction);
  }

  Now works correctly:
  var cp1 = new CommandProcessor(..., bus1, ...);  // cp1._mediator = bus1
  var cp2 = new CommandProcessor(..., bus2, ...);  // cp2._mediator = bus2

  cp1.DepositPost(cmd);  // Uses outbox1 ✓
  cp2.DepositPost(cmd);  // Uses outbox2 ✓ (no longer silently redirected to outbox1)
```

## Related

Fixes #3951


tests/Paramore.Brighter.Extensions.Tests/FactoryThreadSafetyTests.cs
```diff
@@ -0,0 +1,211 @@
...
+        public ThreadSafetyTestCommand() : base(Guid.NewGuid()) { }
+    }
+
+    private class CountingHandler : RequestHandler<CountingCommand>
+    {
+        private static int _instanceCount;
+        public static int InstanceCount => _instanceCount;
+        public static void ResetCount() => _instanceCount = 0;
+
+        public CountingHandler() => Interlocked.Increment(ref _instanceCount);
```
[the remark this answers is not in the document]
thomhurst: @copilot This is fine and there is locking

iancooper:
Cool. It will take us a few days to look through, as weirdly we switched to singleton in V10 because it simplified configuration, which is a pain point.

Not to say we should not do this, but we'll need to puzzle through how this trades off.

That said, there is a lot here that looks attractive in terms of bug fixes and this area has become "overgrown" over time and could do with some serious weeding.

So I think we should shoot for this, but it probably needs a full design review

iancooper:
First, I made some notes around the goals of this in the original issue, [here](https://github.com/BrighterCommand/Brighter/issues/3951). I did that so that they don't get lost in the PR.

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -184,6 +182,9 @@ public CommandProcessor(
             _instrumentationOptions = instrumentationOptions;
             _schedulerFactory = requestSchedulerFactory;
             _resiliencePipelineRegistry = resilienceResiliencePipelineRegistry;
+
+            _mediator = null;
```
iancooper, line 186: The transaction type and the mediator are implicitly linked. I would rather keep a method to set them, to show this relationship, rather than treating them separately.

src/Paramore.Brighter/CommandProcessor.cs
```diff
@@ -1481,30 +1484,16 @@ public async Task ClearOutboxAsync(
...
         }
 
-            /// <summary>
-        /// The external service bus is a singleton as it has app lifetime to manage an Outbox.
-        /// This method clears the external service bus, so that the next attempt to use it will create a fresh one
-        /// It is mainly intended for testing, to allow the external service bus to be reset between tests
+        /// <summary>
+        /// Clears reflection caches. No longer needed to clear mediator state as it is now instance-based.
         /// </summary>
+        [Obsolete("No longer needed - each CommandProcessor instance has its own mediator. Will be removed in next major version.")]
```
iancooper, line 1496: This isn't true if we continue to use the memento pattern to build our dispatch pipelines or around the bound calls. These are shared fixture states for tests, and we need to clear them between tests.

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionMessageMapperRegistryBuilder.cs
```diff
@@ -77,7 +77,7 @@ public void RegisterAsync<TRequest, TMessageMapper>() where TRequest : class, IR
         /// <param name="mapper">The type of the mapper</param>
         public void Add(Type message, Type mapper)
         {
-            serviceCollection.TryAdd(new ServiceDescriptor(mapper, mapper, lifetime));
+            serviceCollection.TryAdd(new ServiceDescriptor(mapper, mapper, ServiceLifetime.Transient));
```
iancooper, line 80: The majority of message mappers don't hold state, so we default to a singleton for this reason. 

I see point 5. But unlike the handler factory, where we explicitly control the lifetime of the handler, we do not do so in `ServiceProviderMessageMapperFactory` or `ServiceProviderMessageMapperFactoryAsync`. 

That code in those factories just grabs the instance out of the DI container.

Now, we could fix that and make this behave much more like the handler factory, so that we track the lifetime of the mapper, and then explicitly control disposal, but I think we need to make a choice here.

Transient mappers and transforms only really exist to support folks who want state in their mapper. Mostly that is for things like monotonically increasing counts etc.

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionMessageMapperRegistryBuilder.cs
```diff
@@ -96,7 +96,7 @@ public void Add(Type message, Type mapper)
         /// <param name="mapper">The type of the mapper</param>
         public void AddAsync(Type message, Type mapper)
         {
-            serviceCollection.TryAdd(new ServiceDescriptor(mapper, mapper, lifetime));
+            serviceCollection.TryAdd(new ServiceDescriptor(mapper, mapper, ServiceLifetime.Transient));
```
iancooper, line 99: See earlier comment

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionMessageMapperRegistryBuilder.cs
```diff
@@ -113,8 +113,8 @@ public void AddAsync(Type message, Type mapper)
         /// </summary>
         public void EnsureDefaultMessageMapperIsRegistered()
         {
-            serviceCollection.TryAdd(new ServiceDescriptor(DefaultMessageMapper, DefaultMessageMapper, lifetime));
-            serviceCollection.TryAdd(new ServiceDescriptor(DefaultMessageMapperAsync, DefaultMessageMapperAsync, lifetime));
+            serviceCollection.TryAdd(new ServiceDescriptor(DefaultMessageMapper, DefaultMessageMapper, ServiceLifetime.Transient));
+            serviceCollection.TryAdd(new ServiceDescriptor(DefaultMessageMapperAsync, DefaultMessageMapperAsync, ServiceLifetime.Transient));
```
iancooper, line 117: See earlier comment

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionMessageMapperRegistryBuilder.cs
```diff
@@ -124,7 +124,7 @@ public void EnsureDefaultMessageMapperIsRegistered()
         /// <param name="defaultMapper">Type of default Message Mapper</param>
         public void SetDefaultMessageMapper(Type defaultMapper)
         {
-            serviceCollection.TryAdd(new ServiceDescriptor(defaultMapper, defaultMapper, lifetime));
+            serviceCollection.TryAdd(new ServiceDescriptor(defaultMapper, defaultMapper, ServiceLifetime.Transient));
```
iancooper, line 127: See earlier comment

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionMessageMapperRegistryBuilder.cs
```diff
@@ -134,7 +134,7 @@ public void SetDefaultMessageMapper(Type defaultMapper)
         /// <param name="defaultMapper">Type of default async message mapper</param>
         public void SetDefaultMessageMapperAsync(Type defaultMapper)
         {
-            serviceCollection.TryAdd(new ServiceDescriptor(defaultMapper, defaultMapper, lifetime));
+            serviceCollection.TryAdd(new ServiceDescriptor(defaultMapper, defaultMapper, ServiceLifetime.Transient));
```
iancooper, line 137: See earlier comment

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionSubscriberRegistry.cs
```diff
@@ -64,7 +60,7 @@ public ServiceCollectionSubscriberRegistry(IServiceCollection services, ServiceL
         /// <param name="handlerType">The type of the handler for this request type</param>
         public void Add(Type requestType, Type handlerType)
         {
-            _services.TryAdd(new ServiceDescriptor(handlerType, handlerType, _lifetime));
+            _services.TryAdd(new ServiceDescriptor(handlerType, handlerType, ServiceLifetime.Transient));
```
iancooper: This is NOT CORRECT. We must allow you to control the lifetime of the handler. This will enable you to set the type scoped, transient, or even singleton as we won't function correctly unless you can, for example, choose scoped if you want to work with EF Core.

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionSubscriberRegistry.cs
```diff
@@ -135,7 +131,7 @@ public void RegisterAsync<TRequest, TImplementation>()
             where TRequest : class, IRequest
             where TImplementation : class, IHandleRequestsAsync<TRequest>
         {
-            _services.TryAdd(new ServiceDescriptor(typeof(TImplementation), typeof(TImplementation), _lifetime));
+            _services.TryAdd(new ServiceDescriptor(typeof(TImplementation), typeof(TImplementation), ServiceLifetime.Transient));
```
iancooper: See earlier comment

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionSubscriberRegistry.cs
```diff
@@ -149,9 +145,9 @@ public void RegisterAsync<TRequest>(Func<IRequest?, IRequestContext?, List<Type>
         {
             handlerTypes.Each(handlerType =>
             {
-                _services.TryAdd(new ServiceDescriptor(handlerType, handlerType, _lifetime));
+                _services.TryAdd(new ServiceDescriptor(handlerType, handlerType, ServiceLifetime.Transient));
```
iancooper: See earlier comment

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionTransformerRegistry.cs
```diff
@@ -37,26 +37,23 @@ namespace Paramore.Brighter.Extensions.DependencyInjection
...
         private readonly IServiceCollection _services;
-        private readonly ServiceLifetime _serviceLifetime;
 
         /// <summary>
         /// Constructs a new instance
         /// </summary>
         /// <param name="services">The Service Collection to register the transforms with</param>
-        /// <param name="serviceLifetime">The lifetime to use for registration. Defaults to Singleton</param>
-        public ServiceCollectionTransformerRegistry(IServiceCollection services, ServiceLifetime serviceLifetime = ServiceLifetime.Singleton)
+        public ServiceCollectionTransformerRegistry(IServiceCollection services)
```
iancooper, line 45: Again, most of our transformer pipelines are singletons as they do not hold state and are probably singletons. As a result, `ServiceProviderFactory` and `ServiceProviderFactory` don't explicitly manage lifetime. So if you want a per-transform state, you have to configure it as transient

And, again, I guess an open question if we just shift this to be more like the handler approach

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionTransformerRegistry.cs
```diff
@@ -37,26 +37,23 @@ namespace Paramore.Brighter.Extensions.DependencyInjection
...
         {
             _services = services;
-            _serviceLifetime = serviceLifetime;
         }
 
         /// <summary>
-        /// Register a transform with the IServiceCollection using the ServiceLifetime
+        /// Register a transform with the IServiceCollection using Transient lifetime
```
iancooper, line 51: See above comment

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionTransformerRegistry.cs
```diff
@@ -37,26 +37,23 @@ namespace Paramore.Brighter.Extensions.DependencyInjection
...
         /// </summary>
         /// <param name="transform">The type of the transform to register</param>
         public void Add(Type transform)
         {
-            _services.TryAdd(new ServiceDescriptor(transform, transform, _serviceLifetime));
+            _services.TryAdd(new ServiceDescriptor(transform, transform, ServiceLifetime.Transient));
```
iancooper, line 56: See above comment

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceProviderHandlerFactory.cs
```diff
@@ -91,28 +94,64 @@ public ServiceProviderHandlerFactory(IServiceProvider serviceProvider)
...
-            if (_isSingleton) return;
-            
+            if (_handlerLifetime == ServiceLifetime.Singleton) return;
+
             var disposal = handler as IDisposable;
             disposal?.Dispose();
             ReleaseScope(lifetime);
         }
 
+        private T? GetOrCreateScoped<T>(Type handlerType, IAmALifetime lifetime) where T : class
```
iancooper: This doesn't get scoped and or transient correctly when matching the names.

iancooper: I understand the change here, which provides a better way to manage the lifetime and is likely to prevent leaks. It took me a little bit to get my head around it, but overall I think this approach is better

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionTransformerRegistry.cs
thomhurst: I'd vote for shifting it to the handler approach. This would give control to the users as to what lifetime they want and also keeps things more consistent

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionMessageMapperRegistryBuilder.cs
thomhurst: Copying other comment as it applies here too: I'd vote for shifting it to the handler approach. This would give control to the users as to what lifetime they want and also keeps things more consistent

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionTransformerRegistry.cs
iancooper: Alright, in that case, would you be alright updating how this works for consistency as well?

iancooper (approved):
@thomhurst This looks great. We need to get a working build, and merge, but I am happy that we can merge this once we get that.

iancooper:
It looks as though `SpyTransaction` which is used to check our Transactional Messaging with the Outbox is failing. We often use `CommitableTransaction` as a default for not a transaction type belonging to a specific Db or our in memory representation. Give me a shout if you can't figure out why it is broken and I can take a peek

thomhurst:
Looks like a caching issue - Updated the cache key to consider the type as well. Ready for a re-run

iancooper:
@thomhurst Mostly looks like timing issues. I suspect these tend to turn up when GHA moves us on to instances with noisy neighbors. We are OSS I guess. If you update for changes to master, then I am happy to merge this in.

thomhurst:
Thanks @iancooper ! Update with the latest changes from main

iancooper:
@thomhurst builds apart from timing failures. going to merge. will probably update packages, etc run locally to check on timing issues, before doing a release. Hopefully next few days.

iancooper:
@thomhurst and thank you so much for the contribution and for sticking with the iterations through it.

PR #3971, opened 2026-01-12
[Feature] Peformance Testing

iancooper:
We have reports of memory leaks, issues with DbContext, etc. So we need some tests that we can run, either when work completes on a PR, or on an intermittent basis (such as pre-release).

This work uses our samples as the application to run and check for leaks.

It uses in-memory Sqlite and RMQ for test reliability.


iancooper:
Review using the sample to do this, as it has high complexity against WebAPI and needs a simpler approach

PR #3990, opened 2026-01-27
use collection expressions where it improves performance

SimonCropp:


iancooper:
Thanks @SimonCropp on a bit of a roll...

PR #3991, opened 2026-01-27
[Fix] ReflectionExtensions Performance Improvments

SimonCropp:


tests/Paramore.Brighter.Core.Tests/Extensions/ReflectionExtensionsTests.cs
lillo42, line 1: I think we don't need this test, I think it'll be catch by other tests

src/Paramore.Brighter/Paramore.Brighter.csproj
```diff
@@ -8,6 +8,10 @@
     <Nullable>enable</Nullable>
   </PropertyGroup>
 
+  <ItemGroup>
```
iancooper: We DON"T allow this.

You may never use InternalsVisibleTo in Brighter as we only test public exports from modules, not implementation details, which have an Internal access modifer.

We discover inner classes that have an Internal modfier through TDD, in particular the refactoring step.

iancooper: This MUST be removed.

tests/Paramore.Brighter.Core.Tests/Extensions/ReflectionExtensionsTests.cs
```diff
@@ -0,0 +1,468 @@
...
+        var method = typeof(NoAttributesHandler)
+            .GetMethod(nameof(NoAttributesHandler.Handle))!;
+
+        var result = method.GetOtherHandlersInPipeline();
+
+        Assert.Empty(result);
+    }
+
+    [Fact]
+    public void When_method_has_single_request_handler_attribute_should_return_it()
```
iancooper: We DONT Test Internals

tests/Directory.Build.props
```diff
@@ -6,5 +6,7 @@
     <IsPackable>false</IsPackable>
     <BrighterTestTargetFrameworks>net9.0;net10.0</BrighterTestTargetFrameworks>
     <BrighterTestNineOnlyTargetFrameworks>net9.0</BrighterTestNineOnlyTargetFrameworks>
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>..\..\Brighter.snk</AssemblyOriginatorKeyFile>
```
iancooper: This is failing the build, and not needed if we remove the reflection tests and InternalsVisibleTo attribute

iancooper:
Thanks @SimonCropp

PR #3995, opened 2026-01-28
feat: Universal scheduler delay support for all transports

iancooper:
## Summary

Implements universal scheduler delay support across all messaging transports as defined in ADR-0037. Previously, each transport handled delayed requeue differently — some blocked the message pump (`Task.Delay`), some ignored delay entirely, and some returned false. This PR unifies the approach:

- **Transports without native delay** (RMQ, Kafka, MQTT, MsSql, Redis) now delegate delayed requeue to a lazily-created producer with a configured scheduler, avoiding message pump blocking
- **InMemory transport** uses the scheduler via its producer, with a timer fallback for backward compatibility
- **Postgres** already had native delay via SQL `visible_timeout` — verified with a new test, no changes needed

### Key changes

- **InMemoryScheduler**: Fixed race condition using `AddOrUpdate` for atomic timer management
- **InMemoryMessageProducer**: `SendWithDelay`/`SendWithDelayAsync` delegate to scheduler when configured, fall back to direct timer otherwise
- **InMemoryMessageConsumer**: Delegates delayed requeue to a lazily-created producer
- **All transport consumers** (RMQ sync/async, Kafka, MQTT, MsSql, Redis): Added lazy producer pattern with scheduler injection for non-blocking delayed requeue
- **Postgres consumer**: Verified existing native SQL delay works correctly (no code changes)

### Implementation pattern (all transports)

1. Add `_producer` field (nullable, lazily created) and `_scheduler` field
2. Add optional `IAmAMessageScheduler?` parameter to consumer constructor
3. Add `EnsureProducer()` using `Interlocked.CompareExchange` for thread safety
4. In `Requeue`/`RequeueAsync`: when delay > 0, use `_producer.SendWithDelay` instead of blocking
5. Dispose `_producer` in `Dispose()`/`DisposeAsync()`

## Test plan

- [x] Phase 1: InMemoryScheduler atomic timer replacement test
- [x] Phase 2: InMemoryProducer scheduler integration tests (4 tests)
- [x] Phase 3: InMemoryConsumer producer delegation tests (4 tests)
- [x] Phase 4: Consumer scheduler injection test
- [x] Phase 5: End-to-end InMemory integration tests (2 tests)
- [x] Phase 6: RabbitMQ consumer tests — sync and async (5 tests)
- [x] Phase 7: Kafka consumer tests + scheduler wiring bug fix (4 tests)
- [x] Phase 8: MQTT consumer tests (3 tests)
- [x] Phase 9: MsSql consumer tests (4 tests)
- [x] Phase 10: Redis consumer tests (4 tests)
- [x] Phase 11: Postgres native SQL delay verification test (1 test)
- [x] All 36/36 tasks complete across 11 phases

### DIFF
--- a/src/Paramore.Brighter/OnSchedulerConflict.cs
+++ b/src/Paramore.Brighter/OnSchedulerConflict.cs
@@ -1,10 +1,13 @@
 ﻿namespace Paramore.Brighter;
 
 /// <summary>
-/// Action that should be executed when there is a conflict during create a scheduler
+/// Action to take when scheduling with a duplicate ID
 /// </summary>
 public enum OnSchedulerConflict
 {
+    /// <summary>
+    /// Throw an <see cref="System.InvalidOperationException"/> if a scheduler with the same ID already exists
+    /// </summary>
     Throw,
 
     /// <summary>


src/Paramore.Brighter.MessagingGateway.RMQ.Sync/RmqMessageConsumer.cs
```diff
@@ -187,6 +187,72 @@ public void Purge()
...
+                HandleException(exception);
+            }
+            catch (Exception exception)
+            {
+                HandleException(exception);
+            }
+
+            return [_noopMessage]; // Default return in case of exception
+        }
+
```
lillo42, line 265: Is it possible to put it in the original place? It's difficult to understand what line was changed

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionBrighterBuilder.cs
```diff
@@ -250,7 +250,7 @@ private void RegisterHandlersFromAssembly(Type interfaceType, IEnumerable<Assemb
             assemblies = assemblies.Concat([assembly]);
             var subscribers =
                 from ti in assemblies.SelectMany(GetLoadableTypes).Distinct()
-                where ti is { IsClass: true, IsAbstract: false, IsInterface: false }
+                where ti is { IsClass: true, IsAbstract: false, IsInterface: false } && (ti.IsPublic || ti.IsNestedPublic)
```
lillo42, line 253:
```suggestion
                where ti is { IsClass: true, IsAbstract: false, IsInterface: false, IsPublic: true, IsNestedPublic: true }
```

lillo42, line 253: Why do we need this?

src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageConsumer.cs
```diff
@@ -62,10 +62,10 @@ public partial class KafkaMessageConsumer : KafkaMessagingGateway, IAmAMessageCo
         private bool _isClosed;
         private readonly RoutingKey? _deadLetterRoutingKey;
         private readonly RoutingKey? _invalidMessageRoutingKey;
-        private readonly Lazy<IAmAMessageProducerSync?>? _deadLetterProducer;
-        private readonly Lazy<IAmAMessageProducerSync?>? _invalidMessageProducer;
-        private readonly Lazy<IAmAMessageProducerAsync?>? _deadLetterProducerAsync;
-        private readonly Lazy<IAmAMessageProducerAsync?>? _invalidMessageProducerAsync;
+        private readonly Lazy<KafkaMessageProducer?>? _deadLetterProducer;
+        private readonly Lazy<KafkaMessageProducer?>? _invalidMessageProducer;
```
lillo42, line 67: Isn't it better to keep to `IAmAMessageProducerSync`? Like in case someone needs/wants to push failed messages to the queue system like SNS/SQS, RabbitMQ and etc

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionBrighterBuilder.cs
iancooper: Stops tests picking up classes that we then can't successfully create. Cause of erratic failing tests

src/Paramore.Brighter.MessagingGateway.RMQ.Sync/RmqMessageConsumer.cs
iancooper: Not sure I understand this comment. If it's ordering, no, it was too complex to navigate the file with jumbled methods

src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageConsumer.cs
iancooper: Helps with dispose, and the interface type would need to be cast everywhere, causing more complex code. I would leave unless we begin to implement read on one channel, dlq on another. I don't think that is a wise change to support btw. I think its a bs requirement

src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionBrighterBuilder.cs
iancooper: In essence, any handler that you want to pick up should be public (otherwise how do you create it)

iancooper:
@claude 

`1. Redis: HandledCount increment removed (regression)

Both Requeue and RequeueAsync in RedisMessageConsumer previously incremented message.Header.HandledCount++ before doing anything. That line is gone in this PR. HandledCount is what the message pump checks to decide whether a message has exceeded its retry limit (RequeueCount). Removing it means Redis consumers with a configured RequeueCount will retry indefinitely. This needs to be restored in both paths (immediate requeue via _inflight and delayed requeue via the producer).`

The Handled Count is incremented in the message pump (Reactor or Proactor). The HandledCount in RedisMessageConsumer was an error that would have resulted in double counting

`3. Kafka Requeue was a documented no-op -- now actively publishes

The previous implementation explicitly documented it as a no-op on Kafka because the stream is immutable and returned true silently. This PR changes Requeue to publish a new message to the same topic. Any existing code that relied on Kafka requeue being a safe no-op will now get actual infinite redelivery unless HandledCount is correctly enforced end-to-end. This is a reasonable change but it is a behavioral breaking change -- it warrants a migration note and ideally a test verifying that RequeueCount still terminates the retry loop for Kafka.`

To force a requeue a user must throw a DeferMessageAction. The previous response, which was not to requeue at all was unexpected for most users. Documentation is a separate project, as it lives in another repository.

`ServiceCollectionBrighterBuilder: internal handlers no longer auto-discovered

The PR adds a visibility check: (ti.IsPublic || ti.IsNestedPublic). Any application using internal handler classes (common in test projects or bounded-context assemblies) will silently stop having those handlers registered after upgrading. This is a breaking change that needs a migration note. The old behavior was intentional -- you control which assemblies are scanned via AddHandlersFromAssemblies. Adding a visibility filter on top of that may be surprising and could break existing users.`

This will be documented in release notes and documentation. A handler needs to be public, so that it can be created by DI. This prevents errors particularly in test projects

PR #3998, opened 2026-01-31
refactor(samples): modernise CommandProcessor samples

holytshirt:
- Use Host.CreateApplicationBuilder() instead of CreateDefaultBuilder()
- Replace GetService with GetRequiredService for safer DI resolution
- Remove Serilog dependencies in favor of built-in logging
- Use string interpolation in handlers
- Make command classes internal sealed
- HelloWorldInternalBus: switch to async handler, remove custom mapper


iancooper:
@holytshirt Do you want to update, and then we can merge?

holytshirt:
Yeah! I try later this evening or over the weekend

PR #3999, opened 2026-02-01
Backstop Message Error Handler

iancooper:
This supplements the work on a universal DLQ by adding a backstop attribute `RejectMessageOnError` that will catch any exception and raise a `RejectMessageAction` which will be captured by the message pump. The pump will, in turn, call Reject on the queue. If the queue is configured to use a dead-letter channel, Reject routes the message to the DLQ.

So far, for middleware without DLQ support, we have rolled this out for Kafka.

We provide a Kafka sample as part of this work.


iancooper:
@jakoss Once this one is in, we will be able to support DLQs for Kafka. (Work will continue in other PRs to pick up the backlog)

jakoss:
@iancooper question around usage - we have our own small abstraction over the handlers, since one of the requirements is to be independent from the brighter api in the rest of the application. Due to that - we cannot use annotations. If i understand correctly - we could just use `RejectMessageOnErrorHandlerAsync` as our base class for the handler and it should work the same?

iancooper:
> @iancooper question around usage - we have our own small abstraction over the handlers, since one of the requirements is to be independent from the brighter api in the rest of the application. Due to that - we cannot use annotations. If i understand correctly - we could just use `RejectMessageOnErrorHandlerAsync` as our base class for the handler and it should work the same?

You just need to throw RejectMessageAction from a handler to make it work. I just added the attribute as a convenience to wrap the handler in a try-catch block that throws RejectMessageAction. You could do that with an abstract base class of your own, that implements try...catch in the handler and calls a virtual method in your derived type. 

Ask me if that doesn't make sense.

jakoss:
Makes total sense :)

iancooper:
Hoping to get a 10.2 out at the end of the week, with Kafka support. Will drop support for the other transports incrementally, but Kafka is the one folks are clamoring for

PR #4012, opened 2026-02-10
FEAT: add Ralph loop for unattended autonomous TDD implementation

iancooper:
Add /spec:ralph-tasks and /spec:ralph-implement commands alongside existing spec workflow, plus scripts/ralph.sh bash loop that restarts Claude with fresh context to prevent context rot during overnight runs.

- ralph-tasks.md: approval-gate-free task format with RALPH-VERIFY and References
- ralph-implement: TDD cycle without AskUserQuestion for unattended execution
- ralph.sh: outer loop with count/file-based stop, model/budget env vars
- Updated .gitignore, spec README, and skills overview with documentation


iancooper:
This doesn't alter code, and I need to test on a genuine piece of work, so checking it straight in.

PR #4013, opened 2026-02-10
feat: tag-based cleanup for orphaned AWS test resources

iancooper:
## Summary

- **Make `SnsAttributes.Tags` configurable** — add optional `List<Tag>? tags` parameter to the constructor (both V4 and non-V4), closing the gap with `SqsAttributes` which already supported tags
- **Wire tags into `CreateTopicAsync()`** — add `CreateTopicTags()` method mirroring the existing `CreateQueueTags()` pattern, so custom tags are merged with the default `Source=Brighter` tag
- **Tag all 200+ AWS test resources** with `Environment=Test` so orphaned resources are identifiable by operators in the AWS Console
- **Replace the cleanup script** — the old `clean_failed_tests_aws_assets.sh` deleted *everything* indiscriminately; the new version uses the Resource Groups Tagging API to find and delete only resources tagged `Environment=Test`, with `--dry-run` support and proper deletion ordering (subscriptions → topics → queues)
- **Add CI cleanup step** to `aws-ci` and `aws-scheduler-ci` jobs with `if: always()` and `continue-on-error: true`

## Design

See [ADR 0037 — AWS Test Resource Cleanup](docs/adr/0037-aws-test-resource-cleanup.md) for the full design rationale.

## Test plan

- [x] Unit test: `SnsAttributes` stores and retrieves custom tags (`When_creating_sns_attributes_with_tags.cs`)
- [x] Integration test: SNS topic created with custom tags via `SnsMessageProducer` (`When_creating_a_topic_with_custom_tags_async.cs`)
- [x] Integration test: SNS pub-sub resources tagged end-to-end (`Sns/.../When_posting_a_message_resources_are_tagged_async.cs`)
- [x] Integration test: SQS point-to-point resources tagged end-to-end (`Sqs/.../When_posting_a_message_resources_are_tagged_async.cs`)
- [x] Bash integration test: cleanup script deletes tagged resources, preserves untagged, supports `--dry-run` (`test_clean_failed_tests_aws_assets.sh`)
- [ ] Verify CI workflow runs cleanup step after AWS tests (requires CI run)
- [ ] Verify `resourcegroupstaggingapi` IAM permissions are available in CI

### DIFF
--- a/clean_failed_tests_aws_assets.sh
+++ b/clean_failed_tests_aws_assets.sh
@@ -33,7 +33,7 @@ if [[ $TAG_API_EXIT -ne 0 ]]; then
     exit 1
 fi
 
-if [[ -z "$RESOURCE_ARNS" ]]; then
+if [[ -z "$RESOURCE_ARNS" || "$RESOURCE_ARNS" == "None" ]]; then
     echo "No resources found with Environment=Test tag."
     exit 0
 fi
--- a/specs/0002-sqs-cleanup/tasks.md
+++ b/specs/0002-sqs-cleanup/tasks.md
@@ -24,8 +24,8 @@ Task 4 (SQS test tags)    ────────▶ Task 5 (Cleanup script)
   - Test file: `When_creating_sns_attributes_with_tags.cs`
   - Test should verify:
     - `SnsAttributes` constructed with `tags: [new Tag { Key = "Environment", Value = "Test" }]` stores them in the `Tags` property
-    - `SnsAttributes` constructed without tags has `Tags` as `null`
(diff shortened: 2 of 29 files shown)


iancooper:
@claude You are wrong on 2. Missing pagination in cleanup script. The [documentation](https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-pagination.html) is clear.  

"When using a command, by default the AWS CLI automatically makes multiple calls to return all possible results to create pagination. One call for each page. Disabling pagination has the AWS CLI only call once for the first page of command results."

So, by default, the AWS CLI will pick up all possible results.

PR #4015, opened 2026-02-16
Universal DLQ

iancooper:
Whilst we have worked on Kafka and InMemory, this PR extends the work out to all of our providers, to ensure that we have consistent behaviour around RejectMessageAction

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -1,6 +1,5 @@
 {
   "permissions": {
-    "ask": [],
     "deny": [
       "Read(./.env*)",
       "Read(./secrets/**)",
@@ -12,7 +11,8 @@
       "Bash(curl:*)",
       "Bash(wget:*)",
       "Bash(ssh:*)"
-    ]
+    ],
+    "ask": []
   },
   "hooks": {
     "PostToolUse": [
@@ -27,5 +27,8 @@
         ]
       }
     ]
+  },
+  "enabledPlugins": {
+    "commit-commands@claude-plugins-official": true
   }
 }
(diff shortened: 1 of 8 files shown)


iancooper:
### Re: Issue #1 & #8 — MsSql "no channels" behaviour

The MsSql implementation is correct as-is, and the apparent divergence from PostgreSQL is intentional, driven by fundamentally different queue semantics.

**MsSql uses atomic read-and-delete.** The `Receive()` SQL is a `DELETE...OUTPUT` statement:

```sql
DELETE FROM cte OUTPUT deleted.Payload, deleted.MessageType, deleted.Topic, deleted.Id
```

The message is permanently removed from the queue table in the same operation that reads it. By the time `Reject()` is called, the message is already gone — there is nothing left to delete. This is also why `Acknowledge()` is a no-op (empty method body). There is no risk of infinite redelivery because the message no longer exists in the source table.

**PostgreSQL uses a visibility timeout model.** `Receive()` marks a row with a visibility timeout but leaves it in the table. Without an explicit `DeleteSourceMessage()`, the row would become visible again after the timeout expires — causing redelivery. That's why PostgreSQL must delete in every path.

So the outcome is identical in both transports: the message is removed from the source queue and will not be redelivered. They differ only in *when* the removal happens (receive-time for MsSql vs. reject/ack-time for PostgreSQL), which is dictated by their underlying queue mechanics.

The `Acknowledge` method signatures confirm this pattern:
- **MsSql**: `public void Acknowledge(Message message) {}` — empty, nothing to do
- **PostgreSQL**: `public void Acknowledge(Message message) { DeleteSourceMessage(...); }` — must explicitly delete

No change needed here.

PR #4018, opened 2026-02-17
Remove Clear Service Bus Calls, Parallelize Tests

iancooper:
Following the move to instance based CommandProcessor dependencies over statics, we can now remove the calls to ClearServiceBus from tests. We can also remove the [Collection("CommandProcessor")] attribute which stopped the tests running in parallel, which should allow them to run faster


iancooper:
I'm going to push this to master, as some of the remaining tests are fixed in #3995 and I need to merge in there

PR #4033, opened 2026-02-27
Pipeline Validation At Startup

### LINKED ISSUE #2176: Enable pipeline validation upon startup
As someone with little experience with Brighter, setting up a pipeline correctly can be a little tricky. For example, it's easy to mistakenly mix sync and async handlers, subscriptions, policies, etc. When building a consumer, Brighter informs me of these issues when I start everything up and send a message to my broker for my application to consume.

Is it possible to implement something that forces Brighter to validate pipelines at startup? Perhaps something configurable so as to only run in certain environments (like on a local machine), to not impact startup times in production?

This would cut down the time for each cycle in the "build everything, run everything, see problem, fix problem, try again"-cycle

iancooper:
We lack validation of our pipelines and the ability to discern what they do easily for diagnostics.

We need both static analysis, with the ability to request validation and receive a report, that follows a similar model to validate methods for ASP.NET or describe other .NET event processing frameworks, and dynamic analysis that extends our recent Roslyn Analyser additions.

This remains one of the weakest parts of our offering, leading to criticism that it is difficult to set Brighter up.

This will close #2176

### DIFF
--- a/docs/adr/0053-pipeline-validation-at-startup.md
+++ b/docs/adr/0053-pipeline-validation-at-startup.md
@@ -649,27 +649,54 @@ Developers opt in through the existing `IBrighterBuilder` interface, which is co
 // Extension methods on IBrighterBuilder
 builder.Services.AddBrighter(options => { /* ... */ })
     .AutoFromAssemblies()
-    .ValidatePipelines()    // enable startup validation
+    .ValidatePipelines()    // enable startup validation (throws on error by default)
     .DescribePipelines();   // enable diagnostic report
 
-// If using producers
+// With explicit flags — developer controls when validation runs
+var isDevelopment = builder.Environment.IsDevelopment();
 builder.Services.AddBrighter(options => { /* ... */ })
     .AutoFromAssemblies()
-    .AddProducers(config => { /* ... */ })
-    .ValidatePipelines()
+    .ValidatePipelines(enabled: isDevelopment)       // only validate in Development
+    .DescribePipelines(enabled: isDevelopment);       // only describe in Development
+
+// Validate but don't terminate — log errors instead of throwing
+builder.Services.AddBrighter(options => { /* ... */ })
+    .AutoFromAssemblies()
+    .ValidatePipelines(throwOnError: false)  // log errors, don't throw
     .DescribePipelines();
 
-// If using consumers (full messaging)
+// Mix and match — validate everywhere but only throw in Development
+var isDev = builder.Environment.IsDevelopment();
 builder.Services.AddBrighter(options => { /* ... */ })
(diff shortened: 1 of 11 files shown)


iancooper:
@claude We have considered and rejected point 2: "Describe() on a Generic Class Ignores Its Type Parameter
PipelineBuilder<TRequest>.Describe(Type requestType)" 

Two issues:

(1) Moving Describe to a separate class carries the risk of model drift between the two. This approach ensures that the path used to build pipelines matches the one we describe.
(2) We do not have an instance of requestType at the point at which we evaluate.

Both these lead us to prefer adding a non-generic method that takes the type

iancooper:
### Code review

Found 2 issues:

1. `ServiceCollectionSubscriberRegistry` does not implement `IAmASubscriberRegistryInspector`, so `ValidatePipelines()` will throw `ConfigurationException` at startup. The factory in `ValidatePipelines()` resolves `ServiceCollectionSubscriberRegistry` and passes it to `PipelineBuilder`, but `PipelineBuilder.Describe()` guards with `if (_subscriberRegistry is not IAmASubscriberRegistryInspector inspector) throw ...`. Only the inner `SubscriberRegistry` field implements that interface. All tests use `SubscriberRegistry` directly, so this bug is not caught by the test suite.

https://github.com/BrighterCommand/Brighter/blob/31ae48cf4ba7dc8cade1bc942b88a1b8d693f915/src/Paramore.Brighter.Extensions.DependencyInjection/BrighterPipelineValidationExtensions.cs#L47-L52

https://github.com/BrighterCommand/Brighter/blob/31ae48cf4ba7dc8cade1bc942b88a1b8d693f915/src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionSubscriberRegistry.cs#L35-L37

2. `PipelineValidationResult` XML doc states "warnings are logged but do not prevent startup", but neither `BrighterValidationHostedService` nor `ServiceActivatorHostedService` actually logs warnings. Both call `result.ThrowIfInvalid()` and discard warnings silently.

https://github.com/BrighterCommand/Brighter/blob/31ae48cf4ba7dc8cade1bc942b88a1b8d693f915/src/Paramore.Brighter/Validation/PipelineValidationResult.cs#L30-L42

https://github.com/BrighterCommand/Brighter/blob/31ae48cf4ba7dc8cade1bc942b88a1b8d693f915/src/Paramore.Brighter.Extensions.DependencyInjection/BrighterValidationHostedService.cs#L66-L77

Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with :+1:. Otherwise, react with :-1:.</sub>

PR #4039, opened 2026-03-01
Database Utilities for our boxes

iancooper:
We had some useful feedback that the two biggest usability issues were the complexity of configuration (covered by a separate ADR) and the management of a box (inbox/outbox).

This PR addresses the management of a box. It derives from lessons from the WebAPI sample and leans into Aspire, because that is the expectation for developers.

## Scope

This PR delivers **both** specifications together, so the bootstrap path is production-ready when the feature first ships:

- **Spec 0023 — Box Database Migration** ([ADR 0053](docs/adr/0053-box-database-migration.md))
  Core provisioning infrastructure: `IAmABoxProvisioner` / `IAmABoxMigrationRunner`, `BoxTableState`, `__BrighterMigrationHistory`, advisory locking per backend, fail-fast hosted service, Aspire connection-name overloads.

- **Spec 0027 — Box Schema Versioning and Migrations** ([ADR 0057](docs/adr/0057-box-schema-versioning-and-migrations.md))
  Proper version chain (V1..V_latest) for outbox (7 versions) and inbox (2 versions) across MSSQL/Postgres/MySQL/SQLite, plus subset-based version detection so existing pre-`DataRef`/`SpecVersion` tables bootstrap correctly. Spanner is fresh-install-only.

- **Spec 0028 — Box Provisioning RDD Role Interfaces** ([ADR 0058](docs/adr/0058-box-provisioning-rdd-role-interfaces.md), [spec dir](specs/0028-box-provisioning-rdd-role-interfaces/))
  A fourth-pass review response on this same PR (per Spec 0028 requirement C1). Restructures the BoxProvisioning surface introduced by Specs 0023 + 0027 around Responsibility-Driven-Design role interfaces (`IAmABoxMigrationDetectionHelper<TConnection, TTransaction>` and its version-detecting extension, `IAmABoxMigrationCatalog`, `IAmABoxPayloadModeValidator<TConnection>`, `IAmAProvisioningUnitOfWork<TTransaction>`) and a template-method `SqlBoxMigrationRunner<TConnection, TTransaction>` that owns the migration algorithm with derived runners supplying only backend-specific hooks. Static helper classes (`{Backend}BoxDetectionHelpers`, `{Backend}{Box}Migrations`, `{Backend}PayloadModeValidator`) become public instance classes; runner ctors gain typed parameters; provisioner ctors gain typed parameters. Behaviour unchanged. Source-breaks licensed by Spec 0028 NF1 (Spec 0027 surface had not shipped). Full source-break enumeration in `release_notes.md` and ADR 0058 §A.1/§A.2/§A.3/§B.

  - **Sub-phase A (post-acceptance review response, 2026-05-12)** — `SqlBoxProvisioner<TConnection, TTransaction>` pull-up consolidating ~640 lines duplicated body across the eight relational provisioners into one ~120-line base. MySQL pre-lock negative-version clamp unified with MsSql/Postgres/Sqlite per F11. ADR 0058 §B.5 + `specs/0028-box-provisioning-rdd-role-interfaces/` sub-phase A appendix.

  - **Sub-phase B (pre-merge rename, post-acceptance, 2026-05-13)** — per [ADR 0059](docs/adr/0059-box-provisioning-abstract-base-naming-symmetry.md) (Accepted 2026-05-13), renamed `RelationalBoxMigrationRunnerBase<TConnection, TTransaction>` → `SqlBoxMigrationRunner<TConnection, TTransaction>` for naming symmetry with sub-phase A's `SqlBoxProvisioner<TConnection, TTransaction>`. Pure Tidy First — zero behavioural delta; build clean on `netstandard2.0;net8.0;net9.0;net10.0`; all BoxProvisioning filters green at the post-13.D floor (MSSQL 64/64, PG 55/55, MySQL 67/67 (test suite net9.0-only; production assembly targets `netstandard2.0;net8.0;net9.0;net10.0` like the other backends), SQLite 46/46, Core 43/43, Spanner 26/26 per TFM). Pre-merge scope — see ADR 0059 §Context. Closes the "Naming asymmetry, time-bounded" risk in ADR 0058.

Bundling these resolves review finding **R1** — without Spec 0027, the Spec 0023 bootstrap path treats any pre-V_latest table as unrecognised and the runner attempts a `CREATE TABLE` that fails. Spec 0027 also addresses **R2** (TOCTOU race on bootstrap insert), **R4** (Spanner concurrency), and **R5** (payload-mode validator test coverage on non-MSSQL backends).

## Breaking Changes

- **`IAmARelationalDatabaseConfiguration.SchemaName`** — a new `string? SchemaName { get; }` member is added to this public interface. The core `Paramore.Brighter` package targets `netstandard2.0`, which does not support default interface members, so this is a **source-breaking** change for any external code that implements `IAmARelationalDatabaseConfiguration`. All in-tree implementors are updated; external implementors must add the property (returning `null` reproduces the previous behaviour).

  ADR 0057 §10 documents the rationale; commit `297ca030f` records the explicit decision to accept the abstract member rather than split target frameworks for one property.

- **Spec 0028 source-breaks** — restructuring the BoxProvisioning surface to RDD role interfaces is purely structural but source-breaks any external implementor of the affected types (static helpers, migration catalogues, payload validators, runner/provisioner ctors). The shipped Brighter call-sites and DI extensions absorb the cascade; `UseBoxProvisioning(opts => opts.Add{Backend}{Box}(config))` users require no change. Full enumeration in `release_notes.md` under "Box Provisioning RDD role-interface refactor (spec 0028)".

## Post-merge follow-up

**(Resolved in-PR by sub-phase B — see Scope above.)** Originally this section committed to a post-merge successor ADR for the §B.2 / §B.5 naming asymmetry. The successor ADR ([0059](docs/adr/0059-box-provisioning-abstract-base-naming-symmetry.md)) and the rename it documents both ship inside PR #4039 itself, eliminating the post-merge migration cost. No outstanding post-merge follow-ups remain for spec 0028.


iancooper:
Re item #4 — the detection algorithm uses `IsSupersetOf(migrations[i].LogicalColumns)` (e.g. `MsSqlBoxDetectionHelper.cs:153, 159`), so extra columns don't fail the check. A V8-shaped table seen by an older Brighter (V1..V7) detects at V7, not 0. In the rolling-deploy scenario the newer replica has already written the V8 history row, so the older replica enters the **normal** path (not bootstrap), reads `maxVersion=8` via `GetMaxVersionAsync` (`MsSqlBoxDetectionHelper.cs:99`), and finds that every V1..V7 migration satisfies `migration.Version ≤ maxVersion` — startup succeeds with zero DDL. Bootstrap is only reached when no history row exists for the box, which can't happen once any rolling-deploy peer has migrated. No code change needed; happy to add a worked example to the ADR if useful.

iancooper:
Re item 2 (`BoxProvisioningOptions.MigrationLockTimeout` is set but never flows to the runners):

Triaged but the wiring already exists. Each relational backend's `Add{Backend}{Box}` registration captures `options.MigrationLockTimeout` inside the late-bound `options.Add(services => …)` lambda and passes it to the runner ctor:

- `MsSqlBoxProvisioningExtensions.cs` lines 30 / 67 / 92 / 129
- `MySqlBoxProvisioningExtensions.cs` lines 30 / 67 / 92 / 129
- `PostgreSqlBoxProvisioningExtensions.cs` lines 30 / 67 / 92 / 129
- `SqliteBoxProvisioningExtensions.cs` lines 36 / 78 / 110 / 152

The lambda runs after the user's `configure(options)` delegate returns, so setting `options.MigrationLockTimeout` anywhere inside `UseBoxProvisioning(o => …)` takes effect regardless of statement order — which matches the contract documented on `UseBoxProvisioning`.

Spanner intentionally does not propagate `MigrationLockTimeout` — it has no advisory-lock concept per ADR 0057 §6.

No code change.

iancooper:
## Batch E response to review comment 4480741284

Validated each item against the current `database_migration` HEAD before
acting. Six commits queued locally, batched per the pattern proven across
Batches A–D. Pushing as one shot when ready.

### Blocker

- **`.gitignore` merge-conflict markers** — fixed (E1 `70045d539`). Markers
  removed and the bottom-of-file `PROMPT.md` duplicate also removed
  (it was already ignored at line 7, the canonical position).

### Issues worth fixing before merge

1. **SQLite UoW narrow rollback catch** — fixed (E2a `968b44963`). Widened
   to `catch (Exception ex)` matching W6's MSSQL stance and the PG UoW's
   release-cleanup catch. New TDD test pins the non-IOE contract via a
   pre-cancelled token (mirrors the W6 MSSQL test).
2. **Marker registration order** — fixed (E2b `8a7216c81`). W7 had
   protected against a throwing `configure`; this commit moves the
   marker registration further, to after the `foreach` over
   `options.Registrations`, so a throw from a per-backend
   `Add{Backend}{Box}` callback also doesn't leak the marker. New TDD
   test pins the registration-callback-throws-then-retry case.
3. **Spanner `VLatest*` hardcoded constants** — fixed (E2c `844a8089a`).
   Switched from `public const int` to `public static readonly int` so a
   partial-rebuild deployment can't silently see a stale V_latest from
   IL-embedded literals. Comment extended to call out the runtime-vs-
   compile-time semantics. All 5 Spanner drift tests still GREEN.

### Observations / minor

4. **PG `hashtext()` collision rationale** — fixed (E3 `f2f9299a1`).
   Doc-only `<para>` block added to the class-level `<remarks>` covering
   the ~1 in 2^32 birthday-bound, the per-deployment population context,
   the failure mode (correctness preserved; widened concurrency
   boundary), and the bigint-overload upgrade path with a SHA-256-derived
   key as future hardening.
5. **Back-compat 2-arg constructors** — deferred to follow-up issue
   [#4139](https://github.com/BrighterCommand/Brighter/issues/4139).
   Spec 0027's DI cascade has not shipped; removing them now would break
   downstream consumers on upgrade.
6. **Sequential `BoxProvisioningHostedService`** — deferred to follow-up
   issue [#4140](https://github.com/BrighterCommand/Brighter/issues/4140).
   No multi-tenant deployment in-scope for PR #4039 exercises the
   linear-startup-cost path.
7. **`Identifiers.AssertSafe` leading underscore** — fixed (E4
   `e9aa73cae`). Tightened the regex to `^[A-Za-z][A-Za-z0-9_]*$` so the
   strictest backend's rule (Spanner rejecting `_`-prefixed identifiers
   as reserved) applies at the framework chokepoint. TDD: flipped the
   existing `_underscore_leading` test case from the safe to the unsafe
   Theory.
8. **Sample `BoxProvisioningFactory` missing Spanner arm** — declining
   to fix in this PR. The WebAPI sample's `Rdbms` enum lacks `Spanner`
   entirely, and the sibling factories (`OutboxFactory`, `InboxFactory`,
   `DbConnectionFactory`, `SchemaCreation`) also lack Spanner branches.
   Adding only the `BoxProvisioningFactory` arm without the cascade
   would be dead code; doing the cascade is scope creep beyond this PR
   and a sample-only change either way (production assemblies are
   unaffected). Sample support for Spanner feels out of scope for
   Brighter's framework boundary — happy to revisit if there's a
   stronger case for the sample as a canonical reference.

iancooper:
## Spec 0027 R1 refactor — landed on this PR (commits `6efe58cef..7e0ed7ca8`)

Per `specs/0027-box-provisioning/requirements.md`, R1 (separating "live builder DDL for fresh install" from "historical V1 baseline" in each catalog) was originally deferred to a future spec. During late self-review on 2026-05-19 I raised this back in-scope: V1's `UpScript` doubling as both contracts was unreadable in code review, and the asymmetry across backends (only 3 of 8 builders have a clean pre-V2 logical baseline; the other 5 were born past the logical-V1 era) needed to be made explicit before merge rather than later. This is the result.

### The deliverable

`V1.UpScript` and the fresh-install DDL are now separate contracts:

- `V1.UpScript` is the literal historical first-commit DDL — what the table *actually* looked like at V1 in production. Five of the eight backends (PG outbox, MSSQL/MySQL/SQLite/PG inbox) were born past the logical-V1 era and carry their V2-equivalent shape in V1; V2's idempotency guard (`COL_LENGTH IS NULL` / `ADD COLUMN IF NOT EXISTS` / `information_schema` probe / `pragma_table_info`) absorbs the redundant column-add on chain replay. Class-level remarks document the asymmetry on every catalog.
- `IAmABoxMigrationCatalog.FreshInstallDdl(config)` is the live builder DDL — what we want a fresh database to look like *today* — sourced from the per-backend `<Backend>{Box}Builder.GetDDL(...)`. The runner takes the catalog at construction; `MigrateAsync` no longer accepts a migrations parameter.
- ADR 0057 §3 rewritten with a "Fresh-install DDL vs V1.UpScript" sub-section and updated architecture diagrams. The two-contract split is documented as the post-Spec-0027 baseline.

### The six-part split (so reviewers don't have to read every commit)

The refactor was deliberately split into structural-only parts (per Beck's Tidy First) so each commit is reviewable in isolation:

1. **`6efe58cef`** — Part 1 — Add `FreshInstallDdl` to `IAmABoxMigrationCatalog`. Pure addition. All 8 relational catalogs implement it by delegating to the existing per-backend builders.
2. **`04a73a47b`** — Part 2 — Runner takes catalog at construction. `IAmABoxMigrationRunner.MigrateAsync` drops the `migrations` parameter; runner sources chain + fresh-install DDL from the injected catalog. Four relational backends + Spanner updated. `RunFreshPathAsync` hook signature changes from `(migrations, ct)` to `(freshInstallDdl, latestVersion, ct)`. Back-compat ctors preserved.
3. **`6ebf8f945`** — Part 3 — Test reshape (99 files, +444/−669). New shared `StubBoxMigrationCatalog`; per-test `MalformedListCatalog` / `BrokenChainCatalog` / `SingleV1Catalog` doubles wire malformed chains for non-monotonic / mid-chain-failure / cancellation tests. Four V1-guardrail tests (`When_<backend>_runner_fresh_path_is_called_with_migrations_not_starting_at_v1_it_should_throw.cs`) deleted — the guardrail no longer exists.
4. **`be1ae9dcb`** — Part 4 — `V1.UpScript` rewritten across all 8 relational catalogs to the literal historical first-commit DDL strings (asymmetric per archaeology). `V1.LogicalColumns` deliberately kept as the logical pre-V2 baseline (the bootstrap-detection contract per ADR 0057 §4).
5. **`40f579645`** — Part 5 — 39 V1 historical-hallmark + `FreshInstallDdl` tripwires added across the four backend drift-test files. Asymmetric per archaeology (MSSQL/MySQL/SQLite outbox `MUST NOT` contain `Dispatched`; PG outbox `MUST` contain `Dispatched`; etc.). ADR 0057 §3 rewritten with the new "Fresh-install DDL vs V1.UpScript" sub-section and updated diagrams.
6. **`7e0ed7ca8`** — Part 6 — Full BoxProvisioning suite GREEN against the new stack. SQLite 53/53 + PG 64/64 + MySQL 73/73 + MSSQL net9.0+net10.0 144/144 + Spanner 31/31. Bootstrap-at-V_k passes on the born-past-V1 backends (V2's idempotency guard absorbs the redundant column-add on chain replay, as designed).

Plus a side-fix discovered during Part 6:

- **`b01dfa38a`** — `fix` — Spanner `EnsureHistoryTableAsync` + `ExecuteDdlSafeAsync` extended to also tolerate `FailedPrecondition` on `CREATE TABLE IF NOT EXISTS` for the BoxProvisioning history-table and box-table DDL paths. Pre-existing emulator-flake unrelated to R1, exposed reliably by the existing `SpannerConcurrentFreshInstallTests`. Box-table schema-drift detection remains the responsibility of the per-backend Drift tests; W7's narrowing for arbitrary DDL is preserved.

### Archaeology — the V1 reality (informational)

| Backend       | Box     | First commit      | Has Dispatched? | Has ContextKey? |
|---------------|---------|-------------------|-----------------|-----------------|
| MSSQL         | Outbox  | `44a405b79` (2019-04) | NO (true V1)        | n/a             |
| PostgreSQL    | Outbox  | `3c30343fa` (2019-07) | YES (born late)     | n/a             |
| MySQL         | Outbox  | `695522367` (2019-03) | NO                  | n/a             |
| SQLite        | Outbox  | `695522367` (2019-03) | NO                  | n/a             |
| MSSQL         | Inbox   | `b7f96957b` (2019-03) | n/a                 | YES (born late)     |
| PostgreSQL    | Inbox   | `1cdc04b60` (2023-11) | n/a                 | YES + composite PK on (CommandId, ContextKey) |
| MySQL         | Inbox   | `b7f96957b` (2019-03) | n/a                 | YES                 |
| SQLite        | Inbox   | `695522367` (2019-03) | n/a                 | YES                 |

3 of 8 builders have a clean "pre-V2 logical baseline" historical state. The other 5 shipped with V2-equivalent columns from inception — V1.UpScript for those backends carries the born-past-V1 shape (with class-level `<remarks>` documenting the asymmetry + V2 idempotency-guard rationale).

Will pick up the new claude[bot] review activity (`4483309864`/`/289`/`/164` from 2026-05-19T00:09 + `4485697019` from 2026-05-19T08:08) in a follow-up batch.

iancooper:
## Batch F triage complete — 13 commits pushed (`7e0ed7ca8..c9ee00311`)

Triaging the four reviewer comments that landed after R1 (review 1/3 strengths `4483309864`, review 2/3 items 1–11 `4483311923`, review 3/3 tests `4483313164`, separate review `4485697019`). One commit per item, RED→GREEN→commit per the established Batch A–E pattern. All four containers brought up (MySQL/PG/MSSQL/Spanner); SQLite needs no infra.

### Fixed (commits)

| Tag | Source | Commit | Title |
|---|---|---|---|
| F1a | M4-1 | `df0e5a847` | MSSQL fresh install honors SchemaName via builder-aware `SqlOutboxBuilder.GetDDL` / `SqlInboxBuilder.GetDDL` — schema-qualifies CREATE TABLE when SchemaName is non-default. V1HistoricalDdl stays unqualified (historical literal). 74/74 GREEN net9.0 + net10.0 |
| F1b | M4-1 | `2c719254e` | PG fresh install honors SchemaName via builder-aware `PostgreSqlOutboxBuilder.GetDDL` / `PostgreSqlInboxBuilder.GetDDL` — schema-qualifies `CREATE TABLE IF NOT EXISTS`. 66/66 GREEN |
| F1c | M4-1 | `c9ee00311` | MySQL fresh install honors SchemaName end-to-end: builder schema-qualifies; V2..V7 ALTERs and `information_schema.columns` probe pin to the configured schema (in MySQL "schema" = "database"); `MySqlBoxDetectionHelper.DoesHistoryExistAsync` now probes the history table against `connection.Database` (where it always lives) while still filtering rows by SchemaName. 76/76 GREEN |
| F2 | M2-3 + M3 | `1e15c7673` | MySQL UoW: `_lockResource = lockResource` swapped to land AFTER `AcquireAsync` succeeds. Regression test `When_mysql_provisioning_uow_begin_acquire_throws_direct_rollback_should_be_clean_no_op` pins the contract — a direct Rollback after a failed Begin is a no-op, no NULL/0 Warning emitted. |
| F3 | M2-4 / M4-2 | `d3c49cfee` | MySQL `EnsureAllowUserVariables` log Debug → Information so operators monitoring connection-string drift see it at default levels. Docstring updated with the security rationale. |
| F4 | M2-6 + M2-8 | `fd18900a7` | `BoxProvisioningOptions.MigrationLockTimeout` XML-doc: MySQL `GET_LOCK` integer-seconds floor (sub-second TimeSpans round up to 1s), and per-replica sequential-phase note linking issue #4140 for the parallelisation follow-up. |
| F5 | M2-7 | `1f5a9aa4e` | `IAmARelationalDatabaseConfiguration.SchemaName` XML doc — see-also `Identifiers.AssertSafe` rule (rejects unsafe identifiers at provisioning entry; uses `<c>...</c>` per the cross-project-cref convention since `Identifiers` lives in `Paramore.Brighter.BoxProvisioning`, not the core assembly). |
| F6 | M2-10 | `8ae02ad26` | Spanner `RunNormalPath` → `VerifyAtLatestVersionOrThrow`. The reviewer suggested `[DoesNotReturn]`; the method has an early `currentVersion == vLatest` return so that annotation would be wrong. Renaming makes the may-throw semantics explicit at the call site. |
| F7 | M2-11 / M4-6 | `84c4f6942` | Restore the `# AI agent session state\nPROMPT.md` block in `.gitignore` (CLAUDE.md references PROMPT.md as session state). The `.claude/settings.json` slopwatch hook removal stays per user direction. |
| F8 | M4-3 | `1e411c57c` | `MsSqlAdvisoryLock.ValidateLockParameters` explicit `ArgumentNullException` guard before the `lockResource.Length` dereference — consistent with the rest of the function's defensive style. |
| F9 | M4-4 | `cc7e7d95d` | MySQL `_activeTableName` `AsyncLocal<string?>` wrapped in `try/finally` so the value is cleared after `CreateUnitOfWorkAsync` reads it — prevents residual ExecutionContext leakage onto continuations after MigrateAsync returns. |
| F10 | M4-Nit | `b726fd817` | PG `EnsureHistoryTableAsync` — explicit `_ = schemaName;` to make the intentional discard observable to IDE unused-parameter warnings (history table lives in `public` regardless of configured box schema). |
| F11 | M4-Nit | `a482593d5` | PG `PostgreSqlBoxDetectionHelper.DoesHistoryExistAsync` UndefinedTable swallow now emits a Debug log — observable if a future contributor adds a DROP-history operation outside test setup. Optional `ILogger` ctor param defaults to `ApplicationLogging.CreateLogger<>()`; backwards-compatible with all existing `new PostgreSqlBoxDetectionHelper()` call-sites. |

### Already-resolved (no commit)

| Tag | Source | Disposition |
|---|---|---|
| F12 | M3 (test gap 2: UseBoxProvisioning double-call) | Test already exists: [`When_use_box_provisioning_is_called_twice_it_should_throw_configuration_exception.cs`](https://github.com/BrighterCommand/Brighter/blob/database_migration/tests/Paramore.Brighter.BoxProvisioning.Tests/When_use_box_provisioning_is_called_twice_it_should_throw_configuration_exception.cs) covers the `BrighterBuilderBoxProvisioningExtensions` double-call guard contract. |

### Declined (with rationale)

| Source | Rationale |
|---|---|
| M2-1 / M4-5 — PR scope/size | Already a deliberate choice — the three specs are bundled per the PR description's R1 stack rationale. For future work of similar size we'll favour stacked PRs. |
| M2-2 — `IAmARelationalDatabaseConfiguration.SchemaName` source-breaking change | Confirmed as a SemVer-major signal; already prominently called out in `release_notes.md` and ADR 0057 §10. |
| M2-5 — PG advisory lock 32-bit hash collision | Reviewer's own assessment: "Right disposition" for the documented <100-tables-per-deployment population. The class-level `<remarks>` already lays out the upgrade path (`(bigint, bigint)` overload with SHA-256-derived 64-bit key) if real collisions surface. |
| M2-9 — `AddWithValue` without explicit DbType | Reviewer's own assessment: "Low priority at this catalog-table volume." Catalog-table cardinality is bounded by the number of provisioned boxes (single-digit); the nvarchar/varchar inference cost the reviewer flagged is real but not observable at this volume. Tracked informally for a future cleanup pass. |

### Test-coverage delta

| Suite | Pre-Batch-F | Post-Batch-F | Notes |
|---|---|---|---|
| SQLite | 53 / 53 | 53 / 53 | unchanged — F1 didn't touch SQLite |
| PG | 64 / 64 | 66 / 66 | +2 F1b non-default-schema tests |
| MySQL | 73 / 73 | 76 / 76 | +1 F2 regression + 2 F1c non-default-schema tests |
| MSSQL net9.0 | 72 / 72 | 74 / 74 | +2 F1a non-default-schema tests |
| MSSQL net10.0 | 72 / 72 | 74 / 74 | +2 F1a non-default-schema tests |

### Provenance note

All four review comments were authored against pre-R1 surface (`3d58dd511` or earlier). Each item was validated against the post-R1 HEAD before action — no items were already-resolved by R1.

Apologies for taking longer than expected — F1c MySQL required more end-to-end work than F1a/F1b because MySQL's "schema = database" semantics meant fixing V2..V7 ALTERs and the detection helper, not just the builders. The reviewer's verdict gate ("non-default SchemaName needs a green integration test before this hits real deployments") is now satisfied on all three backends.

PR #4061, opened 2026-04-04
fix: move inbox attribute addition outside static pipeline cache

holytshirt:
## Summary
- Moves `AddGlobalInboxAttributes` / `AddGlobalInboxAttributesAsync` calls **outside** the static `s_preAttributesMemento` cache block in both `BuildPipeline` and `BuildAsyncPipeline`
- The cache now only stores the handler's declared pipeline attributes — inbox attributes are applied fresh each time based on the current `_inboxConfiguration`
- Previously, inbox attributes were cached alongside handler attributes, causing cross-contamination between pipeline builders with different inbox configurations:
  - Tests **without** inbox config would pick up cached inbox attributes from earlier tests → `ArgumentOutOfRangeException` when the handler factory couldn't create `UseInboxHandlerAsync<T>`
  - Tests **with** inbox config hitting a cache populated without inbox → inbox handler never added → inbox assertions failed
  - On cache miss with inbox config, the async version had a duplicate call that added the inbox handler twice

## Test plan
- [x] CI core tests pass (557 total, 547 passed, 0 failed)
- [x] `AsyncCommandProcessorPublishObservabilityTests` — previously failing, now pass
- [x] `CommandProcessorBuildDefaultInboxPublishAsyncTests` — previously failing, now passes
- [x] No regressions in other pipeline builder tests


iancooper (approved):
TBH I never really liked the global inbox attribute but we came under such pressure from the CTO at Huddle to add it. The problem it creates is why not global outbox, resilience pipeline etc, which gets into fluent over attribute syntax questions

holytshirt:
@iancooper  are we okay with the order being negative, I assume it's not a problem, from reviewing the other code I think it's fine.

iancooper:
I guess

iancooper:
Let us know when you are done

iancooper (approved):

PR #4065, opened 2026-04-15
Draft: TUnit Example

thomhurst:
You mentioned last week that you wondered if TUnit was worth switching to - So thought I'd just do a draft branch to show you what Brighter would look like switched over.

Might be useful to see if test speeds improve or anything. I attempted to enable parallelisation by having tests use unique topic names, or queue names, etc. to better support test isolation. (Though I don't know if limited CI constraints will fight against more parallelisation!)

Don't feel obliged to adopt it or anything, just thought it'd be a bit more useful to see it in something more tangible.


iancooper:
Oh, thanks for this. It's certainly more concrete. I need to let @lillo42 finish his generated tests, because it's hard to hit a moving target, but we could certainly think about this after that.

thomhurst:
@iancooper — question about `CompressPayloadTransformer` sync vs async divergence I ran into while fixing the Core.Tests CI failures on this branch.

`WrapAsync` sets `CharSet` on the compressed `ContentType` (from the source message's charset, falling back to utf-8):

```csharp
contentType.CharSet = message.Header.ContentType?.CharSet ?? CharacterEncoding.UTF8.FromCharacterEncoding();
```

Sync `Wrap` in the same file does **not** set `CharSet`. The two paths produce different headers for the same input.

The compressed payload is binary (stored as `CharacterEncoding.Raw`), and `application/gzip` / `application/deflate` / `application/br` have no registered `charset` parameter per IANA, so attaching one to the compressed `ContentType` looks semantically wrong. The original content type (with charset) is already preserved in `ORIGINAL_CONTENTTYPE_HEADER` for the unwrap side.

This was invisible until commit c76d53c4b switched a sync-origin test from `Wrap` → `WrapAsync`, exposing the mismatch (the existing `CompressLargePayloadTests` assertions expected no charset; the separate `AsyncCompressLargePayloadTests` was written against the async-with-charset behaviour). My fix removed the charset line in `WrapAsync` to match `Wrap`, and updated the duplicate async test file accordingly.

Two reasonable directions here — which do you want?

1. **Keep my change**: sync and async both omit charset on the compressed ContentType (what's on the branch now). Semantically correct per IANA, but a behaviour change for consumers relying on the current async output.
2. **Revert and do the opposite**: update sync `Wrap` to also set `CharSet`. Preserves existing async output, but bakes in the meaningless-charset-on-binary behaviour in both paths.

Happy to flip to (2) if you'd rather preserve the existing on-the-wire format.

thomhurst:
Also noticed that emitted spans may have incorrect durations: 
image

iancooper:
> @iancooper — question about `CompressPayloadTransformer` sync vs async divergence I ran into while fixing the Core.Tests CI failures on this branch.
> 
> `WrapAsync` sets `CharSet` on the compressed `ContentType` (from the source message's charset, falling back to utf-8):
> 
> ```cs
> contentType.CharSet = message.Header.ContentType?.CharSet ?? CharacterEncoding.UTF8.FromCharacterEncoding();
> ```
> 
> Sync `Wrap` in the same file does **not** set `CharSet`. The two paths produce different headers for the same input.
> 
> The compressed payload is binary (stored as `CharacterEncoding.Raw`), and `application/gzip` / `application/deflate` / `application/br` have no registered `charset` parameter per IANA, so attaching one to the compressed `ContentType` looks semantically wrong. The original content type (with charset) is already preserved in `ORIGINAL_CONTENTTYPE_HEADER` for the unwrap side.
> 
> This was invisible until commit [c76d53c](https://github.com/BrighterCommand/Brighter/commit/c76d53c4b6dbfc1ef8072e268f33403ff4facb80) switched a sync-origin test from `Wrap` → `WrapAsync`, exposing the mismatch (the existing `CompressLargePayloadTests` assertions expected no charset; the separate `AsyncCompressLargePayloadTests` was written against the async-with-charset behaviour). My fix removed the charset line in `WrapAsync` to match `Wrap`, and updated the duplicate async test file accordingly.
> 
> Two reasonable directions here — which do you want?
> 
>     1. **Keep my change**: sync and async both omit charset on the compressed ContentType (what's on the branch now). Semantically correct per IANA, but a behaviour change for consumers relying on the current async output.
> 


Let's opt for (1) but we will need to make sure we document. I doubt  anyone is relying on it

PR #4067, opened 2026-04-17
Replay Outgoing Messages When Inbox Receives Duplicate

### LINKED ISSUE #2541: [Feature] Force Replay of Outbox on seeing duplicate Processed Inbox Message
When I receive a command in my inbox that I have processed, I don't want to replay that message if it is not idempotent, as it may have side affects.

But what about producing the same messages downstream that I sent. It may be that the message was resent from upstream because someone downstream of me needs to reprocess it, but they can't do that unless I resend the messages I sent.

So it should be possible to configure the inbox to replay associated messages in the outbox that were raised when we received that command, from the outbox. To do that we simply have to clear the dispatched indicator in the outbox so that the sweeper will resend (we don't need to support immediate send here, if you don't have a sweeper its unlikely you also want this kind of support.

iancooper:
When an Inbox receives a duplicate message, we may choose to replay any outgoing messages that it triggered when it was originally called. This makes us idempotent, in that our behavior matches that of the original call. It also allows us to recreate messages for downstream consumers who might have failed, by triggering from upstream, thus allowing us to replay a whole workflow until we reach the point where it failed.

Closes #2541


iancooper:
Thanks for the thorough review — addressed in `5f966d801`.

**Fixed**
- **#6 (`.gitignore` scope):** restored the `*.db-journal` entry. The `PROMPT.md` change was a deliberate de-dup (it was listed twice) — it's still ignored once, so nothing regresses there.
- **#5 (double `GetOtherHandlersInPipeline()`):** `PipelineBuilder.Describe` now materializes the handler list once and filters it for before/after, instead of two reflection passes.
- **#2 (order-dependent memoized read):** `Add`/`AddAsync` now capture `includeCausation` from the probe once and thread it into `InitAddDbCommand(..., bool includeCausation)`, mirroring the bulk path. The SQL choice and the parameter list now come from a single read rather than relying on call ordering.
- **#3 (replay span mislabelled `MarkDispatched`):** added a dedicated `BoxDbOperation.Replay` (`"replay.causation"`) and applied it to every `ReplayCausation` span — relational (MSSQL/MySQL/Postgres/SQLite), DynamoDB v3/v4, Mongo, Firestore, and InMemory. Left the `WriteToStore` reuse as-is: replay is an `UPDATE ... SET dispatched = NULL`, so the unique/duplicate `catch` branch (and its `DuplicateDetectedInBatch` log) is unreachable in practice.

**Accepted, not changed**
- **#1 (new `IRequestContext.InstrumentationOptions` member is a breaking change):** intentional for the V11 upgrade. Will call it out in the release notes / breaking-changes list.
- **#4 (silent no-op when a custom `IAmARequestContextFactory` returns a non-`RequestContext`):** the fallback degrades safely and is documented in the code comment. Can add a one-time debug log if you'd prefer the diagnostic — happy to follow up.

Verification: core + affected outbox projects build clean; `OnceOnly` + pipeline Core tests pass 194/194 on net9.0 (no test asserts the outbox replay span op, so the relabel is safe).

iancooper:
Follow-up on **#1**: added the breaking-change call-out in `604bf45a0`. `release_notes.md` now has a "Replay Outbox Messages on Inbox Duplicate" section under `## Master` with a `#### Source-breaking change: IRequestContext.InstrumentationOptions` note — flagging that the new required interface member affects any external `IRequestContext` implementor (not just Replay adopters), and that the shipped `RequestContext` already implements it (defaults to `InstrumentationOptions.All`) so normal call sites are unaffected.

iancooper:
@lillo42 Might be worth you casting an eye over the generated tests for causation in this one

lillo42 (approved):

PR #4074, opened 2026-04-24
fix: replace sync-over-async Task.Delay with Thread.Sleep in sync pumps (#4073)

### LINKED ISSUE #4073: Replace sync-over-async Task.Delay().GetAwaiter().GetResult() with Thread.Sleep in sync pumps
## Summary

Several sync code paths use `Task.Delay(x).GetAwaiter().GetResult()` (or `.Wait()`) to pause a dedicated thread. On a dedicated, sync-only thread this is strictly worse than `Thread.Sleep(x)`: it allocates a Task, registers a timer on the TimerQueue, parks on a ManualResetEventSlim, and takes a `HandleNonSuccessAndDebuggerNotification` roundtrip to do what `Thread.Sleep` does with a single OS primitive.

Replacing them would:
- Remove unnecessary allocations and TimerQueue registrations from hot pump loops.
- Eliminate the TimerQueue as a variable when diagnosing pump-shutdown hangs (see context below).
- Match the honest intent of these sites — they explicitly block a dedicated sync thread; there is nothing to yield to.

## Call sites

All of these run on dedicated `LongRunning` threads or the Dispatcher control thread. None are on an async code path:

1. `src/Paramore.Brighter.ServiceActivator/Reactor.cs:123` — `Task.Delay(ChannelFailureDelay).GetAwaiter().GetResult()` (broken-circuit retry)
2. `src/Paramore.Brighter.ServiceActivator/Reactor.cs:131` — `Task.Delay(ChannelFailureDelay).GetAwaiter().GetResult()` (channel-failure retry)
3. `src/Paramore.Brighter.ServiceActivator/Reactor.cs:155` — `Task.Delay(EmptyChannelDelay).GetAwaiter().GetResult()` (empty-channel pause)
4. `src/Paramore.Brighter.ServiceActivator/Dispatcher.cs:348-350` — `Task.Delay(100).GetAwaiter().GetResult()` (Dispatcher.Start spin-wait for DS_RUNNING)
5. `src/Paramore.Brighter.MessagingGateway.RMQ.Sync/RmqMessageGatewayConnectionPool.cs:151` — `Task.Delay(jitter.Next(5, 100)).Wait()` (connect-retry jitter)
6. `src/Paramore.Brighter.MessagingGateway.RMQ.Sync/PullConsumer.cs:78` — `Task.Delay(pause).Wait()` (pull-consumer pause)

## Proposed change

Replace each with `Thread.Sleep(delay)`.

Do **not** touch `Proactor.cs` or any other `await Task.Delay(...)` in genuinely async paths — those correctly yield their pump thread back to the BrighterAsyncContext.

## Context

During investigation of a CI-only test hang (`When_A_Message_Dispatcher_Shuts_A_Connection`, 15-minute timeout before hangdump), the Reactor pumps were observed parked in `Task.InternalWait(Timeout.Infinite)` on a `DelayPromise` at Reactor.cs:155 for the full timeout window. TimerQueue thread was alive, threadpool had idle workers, and no lock contention was visible in the dump. We cannot inspect the Task object directly (the Linux coredump's GC heap is non-walkable), so TimerQueue pathology cannot be ruled out.

Even if TimerQueue is not the cause, these sites have no reason to use async plumbing. `Thread.Sleep` is the right primitive here and may also help narrow down the hang investigation.

## Acceptance

- [ ] All six sites use `Thread.Sleep(delay)` instead of `Task.Delay(delay).GetAwaiter().GetResult()` / `.Wait()`.
- [ ] No new `Task.Delay().GetAwaiter().GetResult()` or `.Wait()` calls introduced elsewhere in sync code.
- [ ] `Proactor.cs` and async paths unchanged.
- [ ] Existing tests pass.

thomhurst:
## Summary

Closes #4073.

Replaces `Task.Delay(x).GetAwaiter().GetResult()` / `.Wait()` with `Thread.Sleep(x)` at sites running on dedicated sync threads (pump loops, Dispatcher control thread, connection-pool retry jitter). Each of these sites explicitly blocks a sync-only thread — there is nothing to yield to — so the async primitive just allocates a `Task`, registers a `TimerQueue` timer, and parks on a `ManualResetEventSlim` to do what `Thread.Sleep` does with a single OS primitive.

Benefits:
- Removes per-iteration allocations and `TimerQueue` registrations from hot pump loops.
- Eliminates the `TimerQueue` as a variable when diagnosing pump-shutdown hangs (see #4073 for context on the Linux CI hang investigation).
- Matches the honest intent of these sites.

## Sites changed

Six from the issue:
- `src/Paramore.Brighter.ServiceActivator/Reactor.cs:123` — broken-circuit retry
- `src/Paramore.Brighter.ServiceActivator/Reactor.cs:131` — channel-failure retry
- `src/Paramore.Brighter.ServiceActivator/Reactor.cs:155` — empty-channel pause
- `src/Paramore.Brighter.ServiceActivator/Dispatcher.cs:342` — `Start()` spin-wait for `DS_RUNNING`
- `src/Paramore.Brighter.MessagingGateway.RMQ.Sync/RmqMessageGatewayConnectionPool.cs:151` — connect-retry jitter
- `src/Paramore.Brighter.MessagingGateway.RMQ.Sync/PullConsumer.cs:78` — pull-consumer pause

One additional found during audit of `src/` for the same anti-pattern:
- `src/Paramore.Brighter.MessagingGateway.MsSql/SqlQueues/MsSqlMessageQueue.cs:105` — `TryReceive(topic, timeout)` sync polling loop

Also dropped `using System.Threading.Tasks;` from the three files where it became unused after the substitution.

`Proactor.cs` and all genuine `await Task.Delay(...)` paths are unchanged — those correctly yield on the async pump. Test-code uses of `Task.Delay(...).Wait()` are out of scope (they run on the test-runner thread, not dedicated pump threads).

## Test plan

- [ ] `Paramore.Brighter.ServiceActivator` builds clean on all TFMs (netstandard2.0, net8.0, net9.0, net10.0) — verified locally.
- [ ] `Paramore.Brighter.MessagingGateway.RMQ.Sync` builds clean on all TFMs — verified locally.
- [ ] `Paramore.Brighter.MessagingGateway.MsSql` builds clean on all TFMs (including net462) — verified locally.
- [ ] Existing ServiceActivator / RMQ.Sync / MsSql test suites pass on CI.
- [ ] Monitor whether `When_A_Message_Dispatcher_Shuts_A_Connection` remains stable in CI after removing the `TimerQueue` from the Reactor pause path.

### DIFF
--- a/src/Paramore.Brighter.MessagingGateway.MsSql/SqlQueues/MsSqlMessageQueue.cs
+++ b/src/Paramore.Brighter.MessagingGateway.MsSql/SqlQueues/MsSqlMessageQueue.cs
@@ -102,9 +102,7 @@ namespace Paramore.Brighter.MessagingGateway.MsSql.SqlQueues
             var timeLeft = timeout.Value.TotalMilliseconds;
             while (!rc.IsDataValid && timeLeft > 0)
             {
-                Task.Delay(RetryDelay)
-                    .GetAwaiter()
-                    .GetResult();
+                Thread.Sleep(RetryDelay);
                 timeLeft -= RetryDelay;
                 rc = TryReceive(topic);
             }
--- a/src/Paramore.Brighter.MessagingGateway.RMQ.Sync/PullConsumer.cs
+++ b/src/Paramore.Brighter.MessagingGateway.RMQ.Sync/PullConsumer.cs
@@ -25,6 +25,7 @@ THE SOFTWARE. */
 
 using System;
 using System.Collections.Concurrent;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Paramore.Brighter.Logging;
@@ -75,7 +76,7 @@ namespace Paramore.Brighter.MessagingGateway.RMQ.Sync
                 }
                 else
                 {
-                    Task.Delay(pause).Wait();
+                    Thread.Sleep(pause);
                 }
                 now = DateTime.UtcNow;
             }
--- a/src/Paramore.Brighter.MessagingGateway.RMQ.Sync/PullConsumer.cs
+++ b/src/Paramore.Brighter.MessagingGateway.RMQ.Sync/PullConsumer.cs
@@ -26,7 +26,6 @@ THE SOFTWARE. */
 using System;
 using System.Collections.Concurrent;
 using System.Threading;
-using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Paramore.Brighter.Logging;
 using RabbitMQ.Client;
--- a/src/Paramore.Brighter.MessagingGateway.RMQ.Sync/RmqMessageGatewayConnectionPool.cs
+++ b/src/Paramore.Brighter.MessagingGateway.RMQ.Sync/RmqMessageGatewayConnectionPool.cs
@@ -24,6 +24,7 @@ THE SOFTWARE. */
 
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Paramore.Brighter.Logging;
@@ -148,7 +149,7 @@ namespace Paramore.Brighter.MessagingGateway.RMQ.Sync
 
         private static void DelayReconnecting()
         {
-            Task.Delay(jitter.Next(5, 100)).Wait();
+            Thread.Sleep(jitter.Next(5, 100));
         }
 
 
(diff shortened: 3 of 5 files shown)


iancooper (approved):

iancooper:
Thanks @thomhurst. You are right, mostly these are deliberate and it may be better to express that over panic folks seeing a GetAwaiter().GetResult().

lillo42 (approved):

PR #4076, opened 2026-04-24
refactor: rewrite BrighterAsyncContext for correctness and modernization

thomhurst:
## Summary

Rewrites `BrighterAsyncContext` and its collaborators in `src/Paramore.Brighter/Tasks/` to fix correctness bugs inherited from the AsyncEx-derived original implementation and modernize the code for current .NET conventions. Adds a thorough test suite covering shutdown races, concurrent-Execute, nested Run, and a 1000-iteration stress test.

## Correctness fixes

- **Outstanding-operations counter leak** — `Enqueue` now balances the counter when `TryAdd` fails on a closed queue. The previous code left the counter stuck and `Run` would hang forever.
- **Late `Post` after shutdown** — `BrighterTaskQueue.TryAdd` / `CompleteAdding` / `GetScheduledTasks` tolerate both `InvalidOperationException` (completed-for-adding) and `ObjectDisposedException` (queue disposed), in the correct subclass-first catch order. Stray async continuations posting back after `Run` returns are absorbed silently.
- **Late `Send` after shutdown** — now throws `ObjectDisposedException` instead of blocking indefinitely on a task the pump will never execute.
- **Exception swallowing in `Run(Func<Task>)` / `Run<T>(Func<Task<T>>)`** — continuation now does `try { WaitAndUnwrapException } finally { OperationCompleted }`, so `OperationCompleted` cannot mask the user's exception.
- **Concurrent `Execute`** — enforces the single-thread invariant via `Interlocked.CompareExchange`, throwing `InvalidOperationException` on re-entry instead of silently violating it.
- **Idempotent `Dispose`** — guarded by `Interlocked.Exchange(ref _disposed, 1)`; double-dispose no longer throws `ObjectDisposedException` from the underlying `BlockingCollection`.
- **`GetScheduledTasks`** — takes a `ToArray()` snapshot rather than iterating the live `BlockingCollection` (which is undocumented against concurrent Add/Take), and absorbs `ObjectDisposedException`.
- **`TaskExtensions.WaitAndUnwrapException(CancellationToken)`** — uses `Task.WaitAsync(ct)` on `net6+`; on `netstandard2.0` unwraps via `ex.InnerException` (the previous code captured the wrapping `AggregateException` instead, double-wrapping the thrown exception).

## Modernization

- `BrighterAsyncContext` and `BrighterSynchronizationContext` are now `sealed`.
- `Enqueue`, `OperationStarted`, `OperationCompleted`, `GetScheduledTasks` on `BrighterAsyncContext` moved from `public` to `internal`.
- `Tuple<Task, bool>` replaced with `(Task, bool)` `ValueTuple` on the hot path.
- Static lambdas with state params eliminate per-`Enqueue` closure allocations.
- `ArgumentNullException.ThrowIfNull` under `#if NET` with netstandard2.0 fallbacks.
- XML docs rewritten for accuracy; MIT attribution block added to `TaskExtensions`.
- Removed dead code: `ExecuteImmediately`, `TryExecuteNewThread`, `BrighterSynchronizationContext.Timeout`, `OutstandingOperations` setter.

## ⚠️ Breaking changes (major bump)

- `BrighterAsyncContext` and `BrighterSynchronizationContext` are now `sealed`.
- `Enqueue`, `OperationStarted`, `OperationCompleted`, `GetScheduledTasks` went from `public` to `internal`.
- `ExecuteImmediately` removed.
- `BrighterSynchronizationContext.Timeout` property removed.
- `OutstandingOperations` setter removed (getter kept and now returns a correct `Volatile.Read` of the internal counter; before, the setter was a dead auto-prop that always returned 0).

No in-repo consumers use any of the removed/tightened surfaces beyond `Run(...)` and `Current`.

## Test coverage

**Added (6 cases):**
- `Post_AfterContextDisposed_DoesNotThrow`
- `Send_AfterContextDisposed_Throws`
- `Send_AfterShutdown_StressIterations_NeverHangsAndOnlyThrowsObjectDisposed` — 1000 iterations, 5s bounded wait per iteration
- `Dispose_CalledTwice_DoesNotThrow`
- `Run_NestedInsideOuterRun_DoesNotDeadlock`
- `Execute_CalledConcurrently_ThrowsInvalidOperationException`

## Test plan

- [x] Build clean across all TFMs: `netstandard2.0`, `net8.0`, `net9.0`, `net10.0`.
- [x] `BrighterSynchronizationContextsTests` — **31/31 passed** in ~1s.
- [x] Proactor regression smoke — **38/38 passed** in ~30s.

## Commits

1. `build: bump OpenTelemetry packages to 1.15.3 for GHSA-g94r-2vxg-569j` — CVE fix required for master to restore. Happy to split into a separate PR if preferred.
2. `refactor: rewrite BrighterAsyncContext for correctness and modernization` — the rewrite itself.

## Review process

This PR was developed with multiple review passes (code-reuse, code-quality, efficiency, and three thorough correctness passes). Each pass surfaced specific findings that were addressed before the next. Review trail visible in the commit history if useful.

### DIFF
--- a/src/Paramore.Brighter/Tasks/BrighterAsyncContext.cs
+++ b/src/Paramore.Brighter/Tasks/BrighterAsyncContext.cs
@@ -125,24 +125,29 @@ public sealed class BrighterAsyncContext : IDisposable
 
         OperationStarted();
 
-        // Attach completion continuation before queueing. The task cannot start until the
-        // scheduler pulls it from the queue, so no completion race is possible here.
+        if (!_taskQueue.TryAdd(task, propagateExceptions))
+        {
+            // Queue is already completed for adding: the task will never be pulled, so we
+            // must not attach the completion continuation - if a caller later inline-executes
+            // the stranded task via TryExecuteTaskInline, the continuation would fire and
+            // double-decrement the outstanding-operations counter. Balance manually instead.
+            // Record shutdown so Send can distinguish "pump will process this task" from
+            // "task will never run".
+            Volatile.Write(ref _shutdown, 1);
+            OperationCompleted();
+            return;
+        }
+
+        // Task is queued. Attach completion continuation to balance OperationStarted when
+        // the task finishes. ExecuteSynchronously runs the continuation inline if the task
+        // has already completed by the time we attach (e.g. the pump pulled and ran it
+        // between TryAdd and here), so the continuation always fires exactly once.
         task.ContinueWith(
             static (_, state) => ((BrighterAsyncContext)state!).OperationCompleted(),
             this,
             CancellationToken.None,
             TaskContinuationOptions.ExecuteSynchronously,
             _taskScheduler);
-
-        if (!_taskQueue.TryAdd(task, propagateExceptions))
-        {
-            // Queue is already completed for adding: the task will never be pulled, so the
-            // continuation will never fire. Balance the outstanding-operations counter
-            // manually so callers of Execute do not hang. Record shutdown so Send can
-            // distinguish "pump will process this task" from "task will never run".
-            Volatile.Write(ref _shutdown, 1);
-            OperationCompleted();
-        }
     }
 
     /// <summary>
--- a/src/Paramore.Brighter/Tasks/BrighterAsyncContext.cs
+++ b/src/Paramore.Brighter/Tasks/BrighterAsyncContext.cs
@@ -1,8 +1,8 @@
-﻿#region Sources
+#region Sources
 
-// This class is based on Stephen Cleary's AyncContext in https://github.com/StephenCleary/AsyncEx
-// The original code is licensed under the MIT License (MIT) <a href="https://github.com/StephenCleary/AsyncEx/blob/master/LICENSE>AyncEx license</a>
-// Modifies the original approach in Brighter which only provided a synchronization synchronizationHelper, not a scheduler, and thus would
+// This class is based on Stephen Cleary's AsyncContext in https://github.com/StephenCleary/AsyncEx
+// The original code is licensed under the MIT License (MIT) <a href="https://github.com/StephenCleary/AsyncEx/blob/master/LICENSE>AsyncEx license</a>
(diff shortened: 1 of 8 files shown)


iancooper:
Thanks @thomhurst I'll take a peek.

iancooper (approved):
Thanks @thomhurst. I was a aware of a couple of cases where we seemed to block the thread, so its great that you managed to track them down. The additional operation is an interesting solution to one of the problems

thomhurst:
### Code review

Found 1 issue:

1. `BrighterAsyncContext.Enqueue` attaches the `OperationCompleted` continuation **before** calling `_taskQueue.TryAdd`. If `TryAdd` fails, the code calls `OperationCompleted()` manually to balance the counter, but the `ContinueWith` is still wired to the task. The stranded task can later be run via `BrighterTaskScheduler.TryExecuteTaskInline` (e.g. a caller doing `task.Wait()` while `BrighterAsyncContext.Current == _asyncContext`), which fires the continuation a second time and decrements `_outstandingOperations` below zero. The next `OperationCompleted()` that hits zero in normal flow will spuriously call `CompleteAdding()` while operations are still in flight, closing the queue prematurely.

   `TryExecuteTaskInline` only checks `BrighterAsyncContext.Current == _asyncContext`; it does not require the task to be in the queue:

   https://github.com/BrighterCommand/Brighter/blob/4a2523e0a745de517c483cebcf4a2307ee14df9d/src/Paramore.Brighter/Tasks/BrighterTaskScheduler.cs#L58-L67

   Enqueue ordering (continuation attached before `TryAdd`, manual decrement on failure):

   https://github.com/BrighterCommand/Brighter/blob/4a2523e0a745de517c483cebcf4a2307ee14df9d/src/Paramore.Brighter/Tasks/BrighterAsyncContext.cs#L109-L146

   Suggested fix: detach/cancel the continuation on `TryAdd` failure, or attach it only after a successful `TryAdd` (with the task observed for completion separately).

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #4079, opened 2026-04-25
fix: pin Task.Factory.StartNew to TaskScheduler.Default (#4071)

### LINKED ISSUE #4071: ServiceActivator Dispatcher/Performer can deadlock under non-default TaskScheduler
### Summary

`Dispatcher.Start()` and `Performer.Run()` call `Task.Factory.StartNew(..., TaskCreationOptions.LongRunning)` without passing an explicit `TaskScheduler`. `StartNew` inherits `TaskScheduler.Current`, and `LongRunning` is only a hint that custom schedulers aren't required to honor. When a caller constructs or starts a `Dispatcher` from inside a limited-concurrency / single-threaded scheduler — including Brighter's own `BrighterAsyncContext.Run(...)` used on the Proactor path — the dispatcher control loop and the performer message pump queue on that blocked scheduler. `End().Wait()` / `await End()` then never returns.

The same pattern exists in the obsolete `Paramore.Brighter.Policies.Handlers.TimeoutPolicyHandler`, which passes `scheduler: TaskScheduler.Current` outright.

### Reproduction

Call the dispatcher from within a non-default scheduler context. The simplest reproduction exists in the test suite after migrating `MessageDispatcherMultiplePerformerTests.WhenAMessageDispatcherStartsMultiplePerformers` from xUnit `[Fact] public void` to TUnit `[Test] public async Task`:

```csharp
[Before(Test)]
public async Task Setup()
{
    await Assert.That(_dispatcher.State).IsEqualTo(DispatcherState.DS_AWAITING);
    _dispatcher.Receive();
}

[Test]
public async Task WhenAMessageDispatcherStartsMultiplePerformers()
{
    await Assert.That(_dispatcher.State).IsEqualTo(DispatcherState.DS_RUNNING);
    await Assert.That(_dispatcher.Consumers.Count()).IsEqualTo(3);
    await _dispatcher.End();            // <-- hangs indefinitely
    await Assert.That(_bus.Stream(new RoutingKey(Topic))).IsEmpty();
    await Assert.That(_dispatcher.State).IsEqualTo(DispatcherState.DS_STOPPED);
}
```

Observed failure: test times out at TUnit's 30-minute default; the diagnostic trace shows the control task parked on `Task.WaitAny` while the performer tasks were never scheduled.

### Why it hangs

`Task.Factory.StartNew` without an explicit scheduler captures `TaskScheduler.Current`. Under any custom scheduler (including `BrighterTaskScheduler` installed by `BrighterAsyncContext.Run`, plus ASP.NET / WPF / async-test host ambient schedulers), the captured scheduler enqueues the LongRunning work behind the very task that is waiting for it. `TaskCreationOptions.LongRunning` does not force a new thread — that is a contract only `TaskScheduler.Default` implements.

This is a production-visible deadlock, not just a test-only issue: any app code that kicks off a `Dispatcher` from inside `BrighterAsyncContext.Run`, from nested `Run` calls, or from a scheduler with limited concurrency, is exposed.

### Proposed fix

Pass `TaskScheduler.Default` explicitly in the affected call sites:

- `src/Paramore.Brighter.ServiceActivator/Dispatcher.cs` — the `_controlTask = Task.Factory.StartNew(...)` in `Start()`.
- `src/Paramore.Brighter.ServiceActivator/Performer.cs` — `Run()` should return `Task.Factory.StartNew(() => _messagePump.Run(), CancellationToken.None, TaskCreationOptions.LongRunning, TaskScheduler.Default)` (also drop the unnecessary `async`/`await` wrapper).
- `src/Paramore.Brighter/Policies/Handlers/TimeoutPolicyHandler.cs` — replace `scheduler: TaskScheduler.Current` with `TaskScheduler.Default`.

`InMemoryBox` already follows this pattern and is a useful reference.

### Follow-ups worth doing alongside the fix

- **Audit**: grep the codebase for `Task.Factory.StartNew` without an explicit `TaskScheduler` and decide per call site whether `Default` or `Current` is correct. The right default is almost always `Default` unless the code genuinely wants to inherit the caller's context.
- **Analyzer rule**: add a Roslyn analyzer / editorconfig pattern against bare `Task.Factory.StartNew` to catch this class of bug on future contributions.
- **Regression test**: start a `Dispatcher` from inside `BrighterAsyncContext.Run(...)`, enqueue a message, and verify `End()` completes within a reasonable bound — complements the existing `When_proactor_shutdown_inside_async_context_should_not_deadlock` test which addressed a narrower case (issue #3684).

### Context

Surfaced during the TUnit migration in #4065 — xUnit's `[Fact] public void` tests invoked the dispatcher on the calling thread with `TaskScheduler.Current == Default`, hiding the bug. TUnit's `async Task` tests exercise the async continuation path and exposed it. Fix is being pushed on the `feature/tunit-migration` branch but the bug itself predates the migration and applies to production callers.

thomhurst:
Closes #4071.

## Summary
- `Dispatcher.Start`, `Performer.Run`, `TimeoutPolicyHandler`, `Mediator.Runner`, and `Mediator.Waker` called `Task.Factory.StartNew` without an explicit `TaskScheduler`, inheriting `TaskScheduler.Current`. Under any limited-concurrency scheduler (`BrighterAsyncContext`, ASP.NET, async test hosts), the control task / message pump queued behind the very thread waiting for them and `End()` / `await` never returned. `TaskCreationOptions.LongRunning` is only a hint — only `TaskScheduler.Default` guarantees a fresh thread.
- Pass `TaskScheduler.Default` explicitly at every affected call site. `Performer.Run` also dropped its redundant `async`/`await` wrapper.
- Added regression test `DispatcherOnLimitedConcurrencySchedulerTests` using `ConcurrentExclusiveSchedulerPair` — verified red without the fix (30 s deadlock) and green with it (~600 ms).

## Test plan
- [x] New regression test fails with `TaskCreationOptions.LongRunning`-only call (deadlock at 30 s) and passes with `TaskScheduler.Default` pin.
- [x] All 81 `MessageDispatch` tests pass on `net9.0`.
- [x] All 40 Policy / Mediator / Workflow tests pass on `net9.0` (2 pre-existing skips on obsolete Timeout API).
- [x] `Paramore.Brighter.ServiceActivator` and `Paramore.Brighter.Mediator` build clean.

### DIFF
--- a/src/Paramore.Brighter.Mediator/Runner.cs
+++ b/src/Paramore.Brighter.Mediator/Runner.cs
@@ -71,12 +71,12 @@ public class Runner<TData>
         var task = Task.Factory.StartNew(async () =>
         {
             cancellationToken.ThrowIfCancellationRequested();
-            
+
             await ProcessJobs(cancellationToken);
-            
+
             cancellationToken.ThrowIfCancellationRequested();
-            
-        }, cancellationToken);
+
+        }, cancellationToken, TaskCreationOptions.DenyChildAttach, TaskScheduler.Default);
         
         Task.WaitAll([task], cancellationToken);
         
--- a/src/Paramore.Brighter.Mediator/Waker.cs
+++ b/src/Paramore.Brighter.Mediator/Waker.cs
@@ -72,7 +72,7 @@ public class Waker<TData>
             if (cancellationToken.IsCancellationRequested)
                 cancellationToken.ThrowIfCancellationRequested();
 
-        }, cancellationToken);
+        }, cancellationToken, TaskCreationOptions.DenyChildAttach, TaskScheduler.Default);
         
         Task.WaitAll([task], cancellationToken);
         
--- a/src/Paramore.Brighter.ServiceActivator/Dispatcher.cs
+++ b/src/Paramore.Brighter.ServiceActivator/Dispatcher.cs
@@ -27,6 +27,7 @@ using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Paramore.Brighter.Extensions;
@@ -335,7 +336,9 @@ namespace Paramore.Brighter.ServiceActivator
                     Log.DispatcherStopped(s_logger);
                 }
             },
-            TaskCreationOptions.LongRunning);
+            CancellationToken.None,
+            TaskCreationOptions.LongRunning,
+            TaskScheduler.Default);
 
             while (State != DispatcherState.DS_RUNNING)
             {
--- a/src/Paramore.Brighter.ServiceActivator/Performer.cs
+++ b/src/Paramore.Brighter.ServiceActivator/Performer.cs
@@ -23,6 +23,7 @@ THE SOFTWARE. */
 #endregion
 
 using System;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Paramore.Brighter.ServiceActivator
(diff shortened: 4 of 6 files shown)


iancooper (approved):
Thanks @thomhurst

PR #4080, opened 2026-04-25
fix: Mediator Runner/Waker RunAsync awaits inner work (#4078)

### LINKED ISSUE #4078: Mediator Runner/Waker fire-and-forget: RunAsync returns before background work finishes
## Summary

`Paramore.Brighter.Mediator.Runner.RunAsync` and `Waker.RunAsync` use `Task.Factory.StartNew(async () => ...)` without calling `.Unwrap()`. `StartNew` returns a `Task<Task>` whose **outer** task completes the moment the lambda hits its first `await` — not when the inner work finishes. The subsequent `Task.WaitAll([task], cancellationToken)` therefore waits on the outer task, which signals completion almost immediately, and `RunAsync` returns (logging "Finished") while `ProcessJobs` / `Wake` is still running on the threadpool.

## Affected sites

- `src/Paramore.Brighter.Mediator/Runner.cs:71-79` — `Task.Factory.StartNew(async () => { ... await ProcessJobs(...); ... })` is not unwrapped.
- `src/Paramore.Brighter.Mediator/Waker.cs:66-77` — `Task.Factory.StartNew(async () => { ... await Wake(...); ... })` is not unwrapped.

## Symptom

```
Starting runner {RunnerName}
Finished runner {RunnerName}   <-- logged immediately, before ProcessJobs has finished
```

The "Finished" log fires the moment the lambda first awaits, even though the loop continues running in the background. Hosts that rely on `RunAsync` returning to know when shutdown has completed will see a false signal. Cancellation does eventually propagate through the inner task, but observability and shutdown ordering are broken.

## Proposed fix

Append `.Unwrap()` so `WaitAll` waits on the inner work:

```csharp
var task = Task.Factory.StartNew(async () =>
{
    cancellationToken.ThrowIfCancellationRequested();
    await ProcessJobs(cancellationToken);
    cancellationToken.ThrowIfCancellationRequested();
}, cancellationToken, TaskCreationOptions.DenyChildAttach, TaskScheduler.Default).Unwrap();

Task.WaitAll([task], cancellationToken);
```

Same change applies to `Waker.RunAsync`.

## Related quality issues in the same files (smaller scope, fixable alongside)

- `Runner.cs:67` and `Waker.cs:62`: `RunAsync` returns `void` despite the `Async` suffix and the fact that it blocks via `Task.WaitAll`. Either rename to `Run` or change the signature to return a `Task` and let callers await.
- `Runner.cs:77`: redundant `cancellationToken.ThrowIfCancellationRequested()` after `await ProcessJobs(cancellationToken)` — `await` already propagates cancellation.
- `Waker.cs:72-73`: `if (cancellationToken.IsCancellationRequested) cancellationToken.ThrowIfCancellationRequested();` — the `if` is pointless because `ThrowIfCancellationRequested` already performs that check.

## Context

Surfaced during code review of the fix for #4071 (the `TaskScheduler.Current` deadlock). Those two files were touched to apply the same `TaskScheduler.Default` pin, and this fire-and-forget bug was visible in the surrounding code. It is pre-existing and out of scope for #4071, hence this separate report.

thomhurst:
## Summary

Fixes #4078.

`Runner.RunAsync` and `Waker.RunAsync` used `Task.Factory.StartNew(async () => ...)` without `.Unwrap()`. The outer `Task<Task>` signaled completion at the lambda's first `await`, so `Task.WaitAll` returned and "Finished" was logged while `ProcessJobs` / `Wake` was still running on the threadpool. Hosts relying on `RunAsync` returning to know when shutdown completes saw a false signal.

## Changes

- `Runner.RunAsync` / `Waker.RunAsync` now return `Task` and `await Task.Run(() => ProcessJobs(ct), ct)`, so callers actually observe completion. `Task.Run` pins to `TaskScheduler.Default`, also guarding against the deadlock pattern reported in #4071.
- Removed redundant `cancellationToken.ThrowIfCancellationRequested()` calls after `await` and the pointless `if (ct.IsCancellationRequested) ct.ThrowIfCancellationRequested()` guard.
- Moved the "Finished" log into a `finally` block so it fires on both clean exit and cancellation.
- Updated 12 workflow tests: `_runner.RunAsync(ct.Token);` → `_ = _runner.RunAsync(ct.Token); await Task.Delay(100ms);` to preserve the implicit settle window the bug was accidentally providing.

## Test plan

- [x] `dotnet test --filter "FullyQualifiedName~Workflows" -f net9.0` — 12/12 passing
- [x] `dotnet build src/Paramore.Brighter.Mediator` — clean

### DIFF
--- a/src/Paramore.Brighter.Mediator/Runner.cs
+++ b/src/Paramore.Brighter.Mediator/Runner.cs
@@ -64,23 +64,19 @@ public class Runner<TData>
     /// Runs the job processing loop.
     /// </summary>
     /// <param name="cancellationToken">A token to monitor for cancellation requests.</param>
-    public void RunAsync(CancellationToken cancellationToken = default(CancellationToken))
+    /// <returns>A task that completes when the job processing loop exits (via cancellation or channel closure).</returns>
+    public async Task RunAsync(CancellationToken cancellationToken = default(CancellationToken))
     {
         s_logger.LogInformation("Starting runner {RunnerName}", _runnerName);
-        
-        var task = Task.Factory.StartNew(async () =>
+
+        try
         {
-            cancellationToken.ThrowIfCancellationRequested();
-
-            await ProcessJobs(cancellationToken);
-
-            cancellationToken.ThrowIfCancellationRequested();
-
-        }, cancellationToken, TaskCreationOptions.DenyChildAttach, TaskScheduler.Default);
-        
-        Task.WaitAll([task], cancellationToken);
-        
-        s_logger.LogInformation("Finished runner {RunnerName}", _runnerName);
+            await Task.Run(() => ProcessJobs(cancellationToken), cancellationToken);
+        }
+        finally
+        {
+            s_logger.LogInformation("Finished runner {RunnerName}", _runnerName);
+        }
     }
 
     private async Task Execute(Job<TData>? job, CancellationToken cancellationToken = default(CancellationToken))
--- a/src/Paramore.Brighter.Mediator/Waker.cs
+++ b/src/Paramore.Brighter.Mediator/Waker.cs
@@ -58,25 +58,19 @@ public class Waker<TData>
     /// This will periodically wake up and trigger due jobs in the scheduler.
     /// </summary>
     /// <param name="cancellationToken">A token to monitor for cancellation requests.</param>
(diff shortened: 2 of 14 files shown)


iancooper (approved):
Thanks @thomhurst for the fixes. All good stuff, just want to note that Mediator is a WiP, and so may change significantly

PR #4081, opened 2026-04-26
fix: Dispatcher shutdown race leaves late-opened Reactor consumers running (#4075)

### LINKED ISSUE #4075: Dispatcher shutdown race can leave late-opened Reactor consumers running after Shut()/End()
### Describe the bug
`Dispatcher.Receive()` can return after `State` becomes `DS_RUNNING` but before all consumers have actually been opened. If `Shut()` or `End()` is called in that window, `Consumer.Shut()` no-ops for consumers that are still in `State = Shut`. The control task then opens those consumers afterwards, so they start running without ever receiving a quit message. `Dispatcher.End()` then waits indefinitely for performer tasks that were never told to stop.

This appears to be distinct from #4073. The `Task.Delay(...).GetAwaiter().GetResult()` frames in the hang dump are where the orphaned Reactor performers were parked after the race, not the primary cause.

### To Reproduce
Repository self-repro from CI:
1. Run `dotnet test --project tests/Paramore.Brighter.Core.Tests/Paramore.Brighter.Core.Tests.csproj --treenode-filter "/*/*/*/*[Fragile!=CI]" -c Release --no-restore --no-build --verbosity d --hangdump --hangdump-timeout 15m --hangdump-type Mini`
2. On GitHub Actions run `24894590307` on April 24, 2026, `When_A_Message_Dispatcher_Restarts_A_Connection_After_All_Connections_Have_Stopped` failed first in setup while waiting for `_dispatcher.Consumers.Any()` to become false.
3. The same run later hung in `When_A_Message_Dispatcher_Shuts_A_Connection` until the 15-minute hangdump timeout.

Links:
- Run: https://github.com/BrighterCommand/Brighter/actions/runs/24894590307
- Hangdump artifact: https://github.com/BrighterCommand/Brighter/actions/runs/24894590307/artifacts/6626956672

Minimal race pattern in code:
1. Create a Reactor `Dispatcher` with multiple performers.
2. Call `Receive()`.
3. Call `Shut(subscription)` and/or `await End()` immediately afterwards.
4. Intermittently, shutdown hangs because only a subset of consumers were open when `Shut()` ran.

Local confirmation:
- I reproduced this with a small console harness against current source (`Version 10.0.0`) that loops the exact `Receive()` then `Shut()/End()` sequence.
- The first reproduced hang was on iteration 16 with:

```text
beforeShut consumers=3, beforeShut open=1, final state=DS_RUNNING, final consumers=2, final openConsumers=2
```

That matches the dump shape: one consumer stopped, two late-opened consumers remained running.

### Exceptions (if any)
No managed exception; this is a hang.

The hang dump from Actions run `24894590307` / artifact `6626956672` showed:
- the test host waiting for `When_A_Message_Dispatcher_Shuts_A_Connection` to finish
- the dispatcher control task blocked in `Dispatcher.Start()` -> `Task.WaitAny(runningTasks)`
- two Reactor performers blocked in `Reactor.Run()` -> `Task.Delay(EmptyChannelDelay).GetAwaiter().GetResult()`

### Further technical details
- Brighter version: `10.0.0`
- Suspect code path:
  - `Dispatcher.Start()` sets `State = DS_RUNNING` before `consumer.Open()`
  - `Dispatcher.Receive()` returns once that state flip is observed
  - `Consumer.Shut()` only enqueues quit when `State == ConsumerState.Open`
- Relevant tests:
  - `tests/Paramore.Brighter.Core.Tests/MessageDispatch/Reactor/When_a_message_dispatcher_shuts_a_connection.cs`
  - `tests/Paramore.Brighter.Core.Tests/MessageDispatch/Reactor/When_a_message_dispatcher_restarts_a_connection_after_all_connections_have_stopped.cs`
- Related issue: #4073

`dotnet --info` (local repro environment):

```text
.NET SDK:
 Version:           11.0.100-preview.2.26159.112
 Commit:            17d11de66c

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.26200
 OS Platform: Windows
 RID:         win-x64

.NET SDKs installed:
  7.0.120
  9.0.313
  10.0.107
  10.0.200-preview.0.26103.119
  10.0.203
  11.0.100-preview.2.26159.112

.NET runtimes installed include:
  Microsoft.NETCore.App 9.0.15
  Microsoft.NETCore.App 10.0.7
  Microsoft.NETCore.App 11.0.0-preview.2.26159.112
```

thomhurst:
Fixes #4075.

## Root cause

`Dispatcher.Start()` flipped `State = DS_RUNNING` *before* opening consumers, so `Receive()`'s 100ms busy-wait could return while some consumers were still in `ConsumerState.Shut`. A `Shut(subscription)` or `End()` call racing into that window would no-op against those consumers — `Consumer.Shut()` only enqueues a quit when `State == Open`. The control task then opened them, leaving orphaned Reactor performers parked in `Task.Delay(EmptyChannelDelay).GetAwaiter().GetResult()` with no way to receive a quit message. `End()` returned the control task, which was blocked in `Task.WaitAny` waiting for performers that would never stop.

This matches the hang dump in #4075 exactly: control task in `WaitAny`, two Reactor performers parked in `Task.Delay`.

## Fix

Two layers:

1. **`Dispatcher.Start()`** — open every consumer and register its task *before* publishing `State = DS_RUNNING`. Replaced the 100ms busy-wait poll with a `TaskCompletionSource<bool>`, which gives callers of `Start()` an explicit happens-before edge over the opens — `Receive()` cannot return until every performer is `Open`. Exceptions during open propagate to the caller via `TrySetException` so it never deadlocks.

2. **`Consumer`** — added a state lock + `_shutRequested` flag. A `Shut()` arriving before `Open()` records intent; the subsequent `Open()` honours it and stays `Shut` (no performer is started). `Dispatcher.Start()` filters out consumers whose `Job` is null when registering tasks. This is defence in depth for any future caller path that opens consumers off the control task.

## Verification

New regression test: `tests/Paramore.Brighter.Core.Tests/MessageDispatch/Reactor/When_dispatcher_shuts_immediately_after_receive_should_not_hang.cs` — three invariants, each looped 25–50 iterations to amplify the race:

- `Receive()` post-condition: every consumer is `Open`.
- `Shut` immediately after `Receive` → `End` finishes within 10s.
- `End` immediately after `Receive` → finishes within 10s.

Pre-fix: all three fail (matches reporter's iter-16 local repro). Post-fix: 3/3 green. Full `MessageDispatch` suite (83 tests) passes on both `net9.0` and `net10.0`.

## Test plan

- [x] `dotnet test --filter FullyQualifiedName~DispatcherShutImmediatelyAfterReceiveTests` — green
- [x] `dotnet test --filter FullyQualifiedName~MessageDispatch` — 83/83 green on net9.0 + net10.0
- [x] `dotnet test --filter "FullyQualifiedName~ControlBus|FullyQualifiedName~ServiceActivator"` — green
- [ ] CI run for the originally-flaky `When_A_Message_Dispatcher_Shuts_A_Connection` and `When_A_Message_Dispatcher_Restarts_A_Connection_After_All_Connections_Have_Stopped`

### DIFF
--- a/src/Paramore.Brighter.ServiceActivator/Consumer.cs
+++ b/src/Paramore.Brighter.ServiceActivator/Consumer.cs
@@ -78,6 +78,13 @@ namespace Paramore.Brighter.ServiceActivator
 
         public int JobId { get; set; }
 
+        // Guards Open/Shut transitions and the _shutRequested intent so that a Shut() arriving
+        // before Open() is honored when Open() runs (the consumer stays Shut and no performer is
+        // started). Without this, a racing Shut on a not-yet-Open consumer would silently no-op
+        // and the late-opened performer would leak — see Dispatcher.Start() and issue #4075.
+        private readonly object _stateLock = new();
+        private bool _shutRequested;
+
 
         /// <summary>
         /// Initializes a new instance of the <see cref="Consumer"/> class.
@@ -99,9 +106,14 @@ namespace Paramore.Brighter.ServiceActivator
         /// </summary>
         public void Open()
         {
-            State = ConsumerState.Open;
-            Job = Performer.Run();
-            JobId = Job.Id;
+            lock (_stateLock)
+            {
+                if (_shutRequested)
+                    return;
+                State = ConsumerState.Open;
+                Job = Performer.Run();
+                JobId = Job.Id;
+            }
         }
 
         /// <summary>
@@ -110,8 +122,11 @@ namespace Paramore.Brighter.ServiceActivator
         /// <param name="topic">The topic we post the quit message to, in order to shut the perfomer</param>
         public void Shut(RoutingKey topic)
         {
-            if (State == ConsumerState.Open)
+            lock (_stateLock)
             {
+                _shutRequested = true;
+                if (State != ConsumerState.Open)
+                    return;
                 Performer.Stop(topic);
                 State = ConsumerState.Shut;
             }
--- a/src/Paramore.Brighter.ServiceActivator/Consumer.cs
+++ b/src/Paramore.Brighter.ServiceActivator/Consumer.cs
@@ -78,10 +78,8 @@ namespace Paramore.Brighter.ServiceActivator
 
         public int JobId { get; set; }
 
-        // Guards Open/Shut transitions and the _shutRequested intent so that a Shut() arriving
-        // before Open() is honored when Open() runs (the consumer stays Shut and no performer is
-        // started). Without this, a racing Shut on a not-yet-Open consumer would silently no-op
-        // and the late-opened performer would leak — see Dispatcher.Start() and issue #4075.
+        // A Shut() arriving before Open() must keep the consumer Shut, otherwise the
+        // late-opened performer leaks with no way to receive a quit message.
         private readonly object _stateLock = new();
         private bool _shutRequested;
 
--- a/src/Paramore.Brighter.ServiceActivator/Dispatcher.cs
+++ b/src/Paramore.Brighter.ServiceActivator/Dispatcher.cs
@@ -306,8 +306,18 @@ namespace Paramore.Brighter.ServiceActivator
 
             Log.DispatcherStarting(s_logger);
 
-            if (!TryOpenConsumers(startup))
-                return;
+            try
+            {
+                OpenConsumers();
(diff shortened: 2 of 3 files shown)


iancooper (approved):
Thanks. Useful fix.

PR #4088, opened 2026-04-26
fix: use TimeProvider for span end time in BrighterTracer (#4087)

### LINKED ISSUE #4087: BrighterTracer span end time ignores TimeProvider — durations broken under FakeTimeProvider
## Summary

`BrighterTracer` uses an injected `TimeProvider` to set the `startTime` on every `Activity` it creates, but the corresponding end time is taken from raw `DateTime.UtcNow` because `Activity.Stop` (called from `Activity.Dispose`) has no `TimeProvider` plug. Under `TimeProvider.System` (production default) both are wall clock and durations are correct. Under any non-system `TimeProvider` (e.g. `Microsoft.Extensions.Time.Testing.FakeTimeProvider` in tests) the start uses simulated time and the end uses real wall clock — `Activity.Duration` becomes meaningless (often huge or negative).

## Affected sites

`src/Paramore.Brighter/Observability/BrighterTracer.cs` snapshots `_timeProvider.GetUtcNow()` and passes it as `startTime` in 10 places:

- Line 113 — `CreateSpan<TRequest>` (CommandProcessor)
- Line 164 — `CreateSpan(MessagePumpSpanOperation.Receive, ...)`
- Line 238 — `CreateArchiveSpan`
- Line 281 — `CreateBatchSpan`
- Line 313 — `CreateClaimCheckSpan`
- Line 369 — `CreateClearSpan`
- Line 409 — `CreateDbSpan`
- Line 470 — `CreateMessagePumpSpan`
- Line 514 — `CreateMessagePumpExceptionSpan`
- Line 564 — `CreateProducerSpan`

End side, `BrighterTracer.cs:621-626`:

```csharp
public void EndSpan(Activity? span)
{
    if (span?.Status == ActivityStatusCode.Unset)
        span.SetStatus(ActivityStatusCode.Ok);
    span?.Dispose();
}
```

`Activity.Dispose` → `Activity.Stop` uses `Activity.GetUtcNow()` which is hard-coded to `DateTime.UtcNow`. The injected `TimeProvider` is not consulted.

## Symptom

- In tests using `FakeTimeProvider` (or any non-system `TimeProvider`), `Activity.Duration` does not reflect the simulated elapsed time. Asserting on durations from observability tests is unreliable.
- In production with `TimeProvider.System` the bug does not manifest because `_timeProvider.GetUtcNow()` and `DateTime.UtcNow` come from the same wall clock.
- Latent risk: any future code path that constructs `BrighterTracer` with a custom `TimeProvider` for legitimate reasons (deterministic replay, simulated load tests, scheduling integration) would silently produce wrong durations.

## Minor secondary issue (same root cause)

The `now` snapshot is taken **before** tag construction, which in some `Create*Span` methods includes `JsonSerializer.Serialize(message.Header, ...)` (e.g. `CreateProducerSpan` at line 564 vs `StartActivity` at 603, and `CreateSpan(Receive)` at line 164 vs 200). The span's `startTime` therefore backdates to before serialization — durations include serialization overhead. Negligible in production, but it widens the start/end inconsistency further.

## Proposed fix

Set the end time explicitly through the `TimeProvider` before disposing:

```csharp
public void EndSpan(Activity? span)
{
    if (span is null) return;
    if (span.Status == ActivityStatusCode.Unset)
        span.SetStatus(ActivityStatusCode.Ok);
    span.SetEndTime(_timeProvider.GetUtcNow().UtcDateTime);
    span.Dispose();
}
```

`Activity.SetEndTime` records the explicit end and `Stop` then preserves it (rather than overwriting with `DateTime.UtcNow`). Same change applies to `EndSpans`.

Optionally, move the `now` snapshot to **immediately before** `StartActivity(...)` in each `Create*Span` so the start time excludes tag-construction overhead.

## Acceptance

- [ ] A test that creates and ends a Brighter span under a `FakeTimeProvider` advanced by N seconds asserts `Activity.Duration == TimeSpan.FromSeconds(N)`.
- [ ] Production behaviour with `TimeProvider.System` is unchanged (durations remain wall-clock accurate).
- [ ] Existing observability tests still pass.

## Context

Surfaced while reviewing PR #4065 (TUnit migration) — pre-existing, unrelated to that PR.

thomhurst:
## Summary

Fixes #4087.

`BrighterTracer.EndSpan` disposed the `Activity` without setting an explicit end time, so `Activity.Stop` (called from `Dispose`) fell back to raw `DateTime.UtcNow` and ignored the injected `TimeProvider`. Span start times use `_timeProvider.GetUtcNow()`, but end times did not — under any non-system `TimeProvider` (e.g. `FakeTimeProvider` in tests) the start used simulated time and the end used real wall clock, so `Activity.Duration` was wildly wrong (often hundreds of days off, depending on how far the simulated clock was from real time).

## Change

`src/Paramore.Brighter/Observability/BrighterTracer.cs` — `EndSpan` now calls `span.SetEndTime(_timeProvider.GetUtcNow().UtcDateTime)` before `Dispose()`. `Activity.Stop` honours the explicit end time when one has been set, so the duration is now `(end via TimeProvider) - (start via TimeProvider)` and is correct under both `TimeProvider.System` (production) and any custom provider (tests, deterministic replay, scheduling integration).

`EndSpans` flows through `EndSpan`, so it is covered without further change.

## Test plan

- [x] New test `BrighterTracerEndSpanTimeProviderTests.When_Ending_A_Span_Duration_Reflects_TimeProvider` creates a span under `FakeTimeProvider`, advances by 5s, ends the span, and asserts `Duration == TimeSpan.FromSeconds(5)`.
- [x] Confirmed RED before the fix: failed with `Expected 00:00:05, Actual 115.13:19:00…` (wall-clock leak).
- [x] Confirmed GREEN after the fix.
- [x] Full Observability suite still passes (61 passed, 2 pre-existing skips, 0 failures) on `net9.0`.

## Notes

The issue also called out an optional secondary fix — moving the `now` snapshot to immediately before `StartActivity` in each `Create*Span` method so tag-construction overhead (`JsonSerializer.Serialize(...)` for `Header`/`Body`) is excluded from span start. That is acknowledged in the issue as "negligible in production" and is not addressed here to keep this change minimal and focused on the correctness bug. Happy to follow up in a separate PR if desired.

### DIFF
--- a/src/Paramore.Brighter/Observability/BrighterTracer.cs
+++ b/src/Paramore.Brighter/Observability/BrighterTracer.cs
@@ -620,9 +620,11 @@ public class BrighterTracer : IAmABrighterTracer
     /// <param name="span">The span to end</param>
     public void EndSpan(Activity? span)
     {
-        if (span?.Status == ActivityStatusCode.Unset)
+        if (span is null) return;
+        if (span.Status == ActivityStatusCode.Unset)
             span.SetStatus(ActivityStatusCode.Ok);
-        span?.Dispose();
+        span.SetEndTime(_timeProvider.GetUtcNow().UtcDateTime);
+        span.Dispose();
     }
 
     /// <summary>
--- a/tests/Paramore.Brighter.Core.Tests/Observability/Common/When_Ending_A_Span_Use_TimeProvider_For_End_Time.cs
+++ b/tests/Paramore.Brighter.Core.Tests/Observability/Common/When_Ending_A_Span_Use_TimeProvider_For_End_Time.cs
@@ -1,0 +1,58 @@
+using System;
+using System.Diagnostics;
+using Microsoft.Extensions.Time.Testing;
+using OpenTelemetry;
+using OpenTelemetry.Resources;
+using OpenTelemetry.Trace;
+using Paramore.Brighter.Core.Tests.CommandProcessors.TestDoubles;
+using Paramore.Brighter.Observability;
+using Xunit;
+
+namespace Paramore.Brighter.Core.Tests.Observability.Common;
+
+public class BrighterTracerEndSpanTimeProviderTests : IDisposable
+{
+    private readonly TracerProvider _traceProvider;
+    private readonly Activity _parentActivity;
+
+    public BrighterTracerEndSpanTimeProviderTests()
+    {
+        _traceProvider = Sdk.CreateTracerProviderBuilder()
+            .AddSource("Paramore.Brighter.Tests", "Paramore.Brighter")
+            .ConfigureResource(r => r.AddService("in-memory-tracer"))
+            .Build();
+
+        _parentActivity = new ActivitySource("Paramore.Brighter.Tests")
+            .StartActivity("BrighterTracerEndSpanTimeProviderTests");
+    }
+
+    [Fact]
+    public void When_Ending_A_Span_Duration_Reflects_TimeProvider()
+    {
+        //arrange
+        var fakeTime = new FakeTimeProvider(new DateTimeOffset(2026, 1, 1, 0, 0, 0, TimeSpan.Zero));
+        var tracer = new BrighterTracer(fakeTime);
+        var command = new MyCommand { Value = "duration" };
+
+        //act
+        var span = tracer.CreateSpan(
+            CommandProcessorSpanOperation.Send,
+            command,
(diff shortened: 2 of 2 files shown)


iancooper (approved):
Thanks, and thanks for the test

Issue #4090, opened 2026-04-26
Pump observability: pumpSpan and processSpan can leak on exception paths


thomhurst:
## Summary

The message pumps (`Reactor` and `Proactor`) start two `Activity` spans whose lifetimes are not bounded by a `try/finally`, so they can leak (started but never ended) on certain exception paths. Both are pre-existing — the second was surfaced during review of the fix for #4085 but exists in the same shape in master.

### 1. `pumpSpan` — leaks on any throw out of the receive loop

`src/Paramore.Brighter.ServiceActivator/Reactor.cs` (`Run`, around line 97 and line 358):
```csharp
public void Run()
{
    var pumpSpan = Tracer?.CreateMessagePumpSpan(MessagePumpSpanOperation.Begin, ...);
    Status = MessagePumpStatus.MP_STARTED;
    do
    {
        // ...
        // throws on the "message is null" branch (around line 152)
        if (message is null)
            throw new Exception("Could not receive message...");
        // ...
    } while (true);

    Tracer?.EndSpan(pumpSpan);  // unreachable if the loop threw
}
```

`src/Paramore.Brighter.ServiceActivator/Proactor.cs` (`EventLoop`, lines 138 and 393): identical pattern.

When the pump throws (the `message is null` path, or any uncaught exception inside the loop), `EndSpan(pumpSpan)` never runs. The `Begin` activity is leaked.

### 2. `processSpan` — leaks if `CreateSpan` itself throws

`src/Paramore.Brighter.ServiceActivator/Reactor.cs` (around line 182):
```csharp
Activity? processSpan = Tracer?.CreateSpan(MessagePumpSpanOperation.Process, message, ...);
try
{
    // dispatch
}
// catches
finally
{
    Tracer?.EndSpan(processSpan);
}
```

`CreateSpan` runs **before** the `try`. Inside `CreateSpan`, `ActivitySource.StartActivity` runs first, then tag serialization (including `JsonSerializer.Serialize(message.Header, ...)`) and `Activity.Current = activity`. If anything between `StartActivity` and the function return throws — or if the same call site throws on subsequent code before reaching the `try` — the activity has been started but the `finally` never engages. `EndSpan` is never called.

`Proactor.cs` mirrors this around line 226.

The receive span (added in #4085 fix) is correctly bounded by an outer `try/finally`. The two spans flagged here are not.

## Impact

[...]

Issue #4089, opened 2026-04-26
Perf: message header serialized twice per message in pump observability hot path


thomhurst:
## Summary

After the fix for #4085, every dispatched message causes `JsonSerializer.Serialize(message.Header, JsonSerialisationOptions.Options)` to run **twice** — once for the receive span, once for the process span. This was a single serialization before the fix.

The double serialization is unavoidable today because the receive span must carry the headers for diagnostics on `MT_UNACCEPTABLE` (failed-to-parse) messages, where there is no process span — but for *serviceable* messages (the steady-state hot path) it is pure duplication.

## Affected sites

- `src/Paramore.Brighter/Observability/BrighterTracer.cs` — `EnrichReceiveSpan` (`MessageHeaders` tag, `Messaging` flag)
- `src/Paramore.Brighter/Observability/BrighterTracer.cs` — `CreateSpan(MessagePumpSpanOperation, Message, ...)` for the process span (`MessageHeaders` tag, `Messaging` flag)
- Triggered per message from `Paramore.Brighter.ServiceActivator/Reactor.cs` and `Proactor.cs`.

## Impact

- Per-message hot path. Multiplies by message volume — measurable under load.
- Reflection-driven JSON of the full `MessageHeader` (topic, partition key, correlation id, content type, all CloudEvents fields, plus the `Bag` dictionary, plus `Baggage`).
- Doubles allocation pressure for the serialized string and intermediate writers.
- Same data, same call, every message — no semantic value to running it twice.

The same pattern likely affects `MessageBody` (RequestBody flag), though strings don't pay reflection cost.

## Suggested approaches

1. **Cache the serialized header on the activity / context for one message lifetime.** Compute once, attach to both spans. Simplest. Memory footprint is one string per in-flight message.
2. **Store the serialized header on `Message` / `MessageHeader` lazily.** Risk: bleeds an observability concern into the domain type.
3. **Skip `MessageHeaders` on the receive span and accept the diagnostic gap on `MT_UNACCEPTABLE`.** Trade-off: operators investigating bad messages lose the raw headers. Not preferred.
4. **Conditional enrichment based on message type.** After `Channel.Receive` returns, inspect `MessageType` — only serialize headers onto the receive span when it's `MT_UNACCEPTABLE` (or an error path), since the process span handles serviceable messages. Best perf, modest complexity.

[...]

PR #4091, opened 2026-04-26
fix: receive span covers broker call; add process span (#4085)

### LINKED ISSUE #4085: Receive span starts after Channel.Receive — duration excludes broker call
## Summary

`BrighterTracer.CreateSpan(MessagePumpSpanOperation.Receive, ...)` is invoked **after** the synchronous `Channel.Receive` / `Channel.ReceiveAsync` call returns. The span is then closed in the loop's `finally`, after dispatch and ack. Net effect: a span named `<topic> receive` measures dispatch + ack work, not the broker receive — and the `messaging.client.operation.duration{messaging.operation.name="receive"}` histogram derived from it lies about receive latency.

## Affected sites

- `src/Paramore.Brighter.ServiceActivator/Reactor.cs:113-115` — `message = Channel.Receive(TimeOut);` then `span = Tracer?.CreateSpan(MessagePumpSpanOperation.Receive, message, ...);`
- `src/Paramore.Brighter.ServiceActivator/Proactor.cs:155-156` — same pattern with `Channel.ReceiveAsync(TimeOut)`.
- `src/Paramore.Brighter.ServiceActivator/Reactor.cs:350` and the equivalent in `Proactor.cs` — `finally { Tracer?.EndSpan(span); }` after dispatch returns.

## How the metric is consumed

`src/Paramore.Brighter/Observability/BrighterMetricsFromTracesProcessor.cs:66-69`:

```csharp
case "receive":
    messagingMeter.RecordClientOperation(activity);
    messagingMeter.AddClientConsumedMessage(activity);
    break;
```

`MessagingMeter.RecordClientOperation` records `activity.Duration.TotalSeconds` into `messaging.client.operation.duration`. Because the activity was started after the receive call, that duration reflects dispatch and ack, not the time spent waiting on the broker.

## Symptom

- Trace spans labeled `receive` appear shorter than the actual broker call when the broker is slow (because the broker time isn't in the span at all).
- Spans appear **longer** than expected when dispatch is slow, because dispatch work is wrongly attributed to "receive".
- The `messaging.client.operation.duration` histogram is consequently mis-bucketed for the `receive` operation.

## Proposed fix

Two reasonable shapes — either:

1. **Move span creation earlier**: snapshot tags from headers we already have (topic, messaging system) before calling `Channel.Receive`, start the span around the receive call, then enrich it with message-derived tags after the call returns. The error path in the surrounding `catch` blocks already creates a span via `CreateMessagePumpExceptionSpan` — those would feed off the same earlier-started span instead.
2. **Keep current `receive` span but add a separate `process` span** for dispatch (the `MessagePumpSpanOperation.Process` enum value already exists at `src/Paramore.Brighter/Observability/MessagePumpSpanOperation.cs:10` and is handled by the metrics processor's `"process"` case at `BrighterMetricsFromTracesProcessor.cs:70-72`, but is never created anywhere in production code).

Option 2 is closer to the OpenTelemetry messaging conventions (separate `receive` and `process` operations) and would also fix the dead-enum/dead-histogram issue (`messaging.process.duration` is currently never recorded).

## Acceptance

- [ ] `<topic> receive` span's `Duration` covers the broker `Channel.Receive` / `ReceiveAsync` call.
- [ ] `messaging.client.operation.duration{operation=receive}` reflects broker latency, not dispatch latency.
- [ ] If option 2 is taken: a `<topic> process` span is created around dispatch and `messaging.process.duration` records non-zero values for handled messages.
- [ ] Existing observability tests pass (`tests/Paramore.Brighter.Core.Tests/Observability/MessageDispatch/*`).

## Context

Surfaced while reviewing PR #4065 (TUnit migration) — the issue is pre-existing and unrelated to that PR.

thomhurst:
Fixes #4085.

## Problem

`BrighterTracer.CreateSpan(Receive, ...)` was called **after** `Channel.Receive` / `ReceiveAsync` returned, then ended after dispatch + ack. Net effect: the span named `<topic> receive` measured dispatch and ack work — not the broker call — and the `messaging.client.operation.duration{operation=receive}` histogram derived from it lied about broker latency.

The `MessagePumpSpanOperation.Process` enum value and the matching `messaging.process.duration` histogram had been defined for some time but were never produced anywhere in production code.

## Fix

- The **receive span** is now started before the broker call and enriched with message-derived tags after the call returns. Its `Duration` reflects only broker latency.
- A new **process span** (using the existing `MessagePumpSpanOperation.Process` value) wraps dispatch — translation, handler invocation, and the ack decision. Feeds the previously-dead `messaging.process.duration` histogram.
- Process span carries the producer's `traceparent` so handler spans descend from the producer trace; receive span has a local pump-span parent. Matches the OpenTelemetry messaging conventions of separate `receive` and `process` operations.

## Public API

Two new members on `IAmABrighterTracer`:

- `CreateReceiveSpan(RoutingKey, MessagingSystem, InstrumentationOptions)` — starts the receive span before the broker call.
- `EnrichReceiveSpan(Activity?, Message, InstrumentationOptions)` — adds message-derived tags and propagates producer tracestate/baggage after the broker call returns.

### ⚠️ Interface break

`IAmABrighterTracer` is no longer source-compatible: external implementations must add the two new members. Per review discussion the risk is acceptable — the interface is not expected to have external implementers (none exist in this repo), and the cleaner pattern (separation of concerns: span work belongs with the tracer, not the pump) outweighs preserving binary compatibility for an unlikely consumer. Release notes should call this out.

The earlier extension-method approach (`BrighterTracerReceiveExtensions`) is gone. Reasons interface members were chosen over extensions:

- Separation of concerns — addresses original review feedback that span helpers belong on `BrighterTracer`, not the pump or a detached extensions class.
- Symmetry with every other span-creating method on the interface (`CreateSpan`, `CreateProducerSpan`, `CreateMessagePumpSpan`, etc.).
- Single `TimeProvider` source — the tracer's `_timeProvider` drives both span start and `EndSpan`, fixing an inconsistency where the extension took the pump's `TimeProvider` for start time but `EndSpan` used the tracer's for end time.
- Drops the redundant `TimeProvider?` parameter — resolves the CodeScene "Excess Number of Function Arguments" advisory (5→3 args).
- Mockable by test doubles; static extensions are not.

Pump call sites use the null-conditional operator (`Tracer?.CreateReceiveSpan(...)` / `Tracer?.EnrichReceiveSpan(...)`) since `Tracer` is `IAmABrighterTracer?`.

## Behaviour preservation

Walked every exit path of the receive section to confirm no regressions vs the pre-fix code:

| Path | Receive span ended? | Status set? | Tags preserved? |
|------|---------------------|-------------|-----------------|
| `BrokenCircuitException` | ✓ via `finally` | `Error` + exception attached | tags from before broker call |
| `ChannelFailureException` | ✓ via `finally` | `Error` + exception attached | tags from before broker call |
| Generic `Exception` | ✓ via `finally` | `Error` + exception attached | tags from before broker call |
| `message is null` (throws) | ✓ via `finally` | `Error` + description | from broker call |
| `MT_NONE` (empty) | ✓ via `finally` | `Ok` (default via `EndSpan`) | enriched |
| `MT_UNACCEPTABLE` | ✓ via `finally` | `Error` + parse-failure description | enriched, **incl. `MessageHeaders` and `MessageBody`** for diagnostic |
| `MT_QUIT` | ✓ via `finally` | `Ok` (default via `EndSpan`) | enriched |
| Serviceable | ✓ via `finally`, before dispatch | `Ok` | enriched |

Receive span continues to carry **all** the message tags it carried pre-fix (`MessageHeaders`, `MessageBody`, all CloudEvents fields, etc.) plus producer `traceState` and `Baggage` (with `correlationId`) — verified necessary for the `MT_UNACCEPTABLE` path where there is no process span and operators need full diagnostic context.

## Acceptance (per #4085)

- [x] `<topic> receive` span's `Duration` covers the broker call
- [x] `messaging.client.operation.duration{operation=receive}` reflects broker latency
- [x] `<topic> process` span created around dispatch; `messaging.process.duration` records non-zero values for handled messages
- [x] Existing observability tests pass (`tests/Paramore.Brighter.Core.Tests/Observability/MessageDispatch/*`)

## Tests

- Updated three existing tests in `Observability/MessageDispatch/`:
  - Span counts adjusted for the new process span (6→7 dispatch, 7→8 channel failure cases).
  - `ParentId` assertion moved from receive span to process span (process span inherits the producer `traceparent`; receive span has a local parent).
  - Both spans now asserted to exist with their respective `messaging.operation.type` tag.
- New test `When_A_Message_Is_Processed_It_Should_Have_A_Process_Span` covers the acceptance criteria explicitly.

Test results on net10.0: 62 passed (2 pre-existing skips) for Observability filter; 85 passed for MessageDispatch filter.

## Follow-ups (separate issues)

Surfaced during review of this fix, filed separately to keep this PR focused:

- **#4089** — Perf: `JsonSerializer.Serialize(message.Header, ...)` runs twice per serviceable message (once for receive span, once for process span). Acceptable trade-off here for diagnostic completeness on `MT_UNACCEPTABLE`; the issue tracks recovering throughput with conditional enrichment based on `MessageType`.
- **#4090** — `pumpSpan` and `processSpan` can leak (started but never ended) on certain throw paths because their lifetimes aren't bounded by `try/finally`. Pre-existing in master; receive span now does the right thing and the same pattern should be applied to the other two spans.
- Per review on this PR: split `BrighterTracer` by concern (Producer / Consumer / Inbox / Outbox) once more receive-side helpers accumulate.


src/Paramore.Brighter.ServiceActivator/MessagePump.cs
```diff
@@ -199,6 +204,77 @@ protected void ValidateMessageType(MessageType messageType, IRequest request)
             }
         }
 
+        // Pump-internal helpers for the receive span. Live here (not on BrighterTracer) to avoid growing the public API.
```
iancooper: I would disagree with that on the basis of the separation of concerns. These would be better on BrighterTracer. If BrighterTracer grows unwieldy, the answer is to break that up by concern, not to locate the helper in the Pump

iancooper:
I checked https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans, and this seems correct. The receive span should check the time for the call to the broker and process our time accordingly.

So this looks correct

iancooper (commented):
Can you push those tracer methods to BrighterTracer, not the Pump? We are mixing concerns. If we think that BrighterTracer is too large, let's raise a separate issue and split it by Producer/Consumer/Inbox/Outbox or similar convention.

src/Paramore.Brighter.ServiceActivator/MessagePump.cs
thomhurst: @iancooper In order to use the new methods on the interface, we'd either need to add those new methods to the interface, which would be breaking if anyone else implements it, or add them as extension methods.

I think the latter would be less disruptive?

src/Paramore.Brighter/Observability/BrighterTracerReceiveExtensions.cs
```diff
@@ -0,0 +1,136 @@
...
+            tags.Add(BrighterSemanticConventions.MessagingOperationType, operation.ToSpanName());
+        if (options.HasFlag(InstrumentationOptions.Messaging))
+        {
+            tags.Add(BrighterSemanticConventions.MessagingDestination, topic);
+            tags.Add(BrighterSemanticConventions.MessagingSystem, messagingSystem.ToMessagingSystemName());
+            tags.Add(BrighterSemanticConventions.Operation, operation.ToSpanName());
+        }
+
+        var activity = tracer.ActivitySource.StartActivity(
+            name: $"{topic} {operation.ToSpanName()}",
```
lillo42: Maybe  we should use the subscription name, not the topic

```suggestion
            name: $"{subscriptionName} {operation.ToSpanName()}",
```

src/Paramore.Brighter.ServiceActivator/MessagePump.cs
iancooper: I doubt anyone else implements it, so I think the risk is acceptable, provided we add to the release notes.

iancooper:
@thomhurst Sorry, my Span<T> fix has made this stale, can you fix (warnings as errors issue):

"Error: /home/runner/work/Brighter/Brighter/src/Paramore.Brighter/Observability/BrighterTracerReceiveExtensions.cs(118,70): error CS0618: 'MessageBody.Bytes' is obsolete: 'Use Memory for zero-copy access. This property allocates on every call.' [/home/runner/work/Brighter/Brighter/src/Paramore.Brighter/Paramore.Brighter.csproj::TargetFramework=netstandard2.0]"

thomhurst:
@iancooper should be fixed. Updated the tracer interface. Technically a breaking change, but like you say, probably no consumers

iancooper (approved):

iancooper:
@thomhurst Going to merge for you to try and clear our queue

PR #4103, opened 2026-04-26
fix: each RedisMessageGateway owns its own pool (#4023)

### LINKED ISSUE #4023: [Bug] Redis: static s_pool in RedisMessageGateway is clobbered when multiple gateway instances coexist
## Summary

`RedisMessageGateway` uses a **static** `Lazy<RedisManagerPool>? s_pool` field that is unconditionally overwritten in every constructor call (`s_pool = _myPool`). When a consumer creates a lazily-initialized requeue producer (introduced in the universal scheduler delay feature), the producer's constructor overwrites `s_pool`, causing the consumer to silently switch to the producer's connection pool mid-lifetime.

## Root Cause

```csharp
// RedisMessageGateway.cs constructor
_myPool = new Lazy<RedisManagerPool>(() =>
{
    OverrideRedisClientDefaults();
    return new RedisManagerPool(_gatewayConfiguration.RedisConnectionString, new RedisPoolConfig());
});
s_pool = _myPool;  // <-- unconditionally clobbers the shared static
```

Every `RedisMessageGateway` subclass (`RedisMessageConsumer`, `RedisMessageProducer`) runs this constructor. The last instance constructed wins ownership of `s_pool`. `GetClient()` and `GetClientAsync()` both read from `s_pool`, so all gateway instances share whichever pool was most recently assigned.

### Triggered by `EnsureRequeueProducer()` (new in universal_delay branch)

When `RedisMessageConsumer.EnsureRequeueProducer()` creates a `new RedisMessageProducer(...)`, the producer's base constructor overwrites `s_pool` with its own `_myPool`. From that point:

1. The **consumer** starts getting clients from the **producer's pool** instead of its own
2. The consumer's original pool may have already been lazily created with active connections — those connections are orphaned until disposal
3. `DisposePool()` uses `Interlocked.CompareExchange(ref s_pool, null, _myPool)` — the consumer's dispose will **fail to null `s_pool`** because `s_pool` no longer equals the consumer's `_myPool`

### Pre-existing (affects any multi-instance scenario)

The same issue exists whenever multiple consumers or producers are created in the same process — each constructor clobbers the previous `s_pool`. This affects:
- Multiple subscriptions with different topics
- Consumer + producer pairs
- Any scenario with more than one `RedisMessageGateway` subclass instance

### Why it doesn't cause failures today

Both pools connect to the same Redis instance with the same configuration, so the pool switch is functionally transparent. Both the consumer's and producer's pools **are** disposed on cleanup (`DisposePool` always disposes `_myPool.Value` regardless of whether `CompareExchange` succeeds). The issue is architectural — the static mutable state is a code smell that makes reasoning about lifetime and ownership difficult.

## Suggested Fix

Replace the static `s_pool` with instance-based pool management. Options:

### Option A: Instance-based pool (simplest)
Remove `s_pool` entirely. Each gateway instance uses its own `_myPool` directly in `GetClient()`/`GetClientAsync()`. This trades connection pooling efficiency for correctness — each consumer/producer gets its own pool.

### Option B: Shared pool via DI
Inject a shared `RedisManagerPool` (or a factory for one) via constructor parameter. The DI container manages the singleton lifetime. This preserves pool sharing without static mutable state.

### Option C: Thread-safe static initialization
Use `Interlocked.CompareExchange` in the constructor to only set `s_pool` if it's currently null:
```csharp
Interlocked.CompareExchange(ref s_pool, _myPool, null);
```
First instance wins; subsequent instances reuse the existing pool. Requires all instances to use the same Redis configuration (which is currently assumed).

## Related

- Branch `universal_delay` / Spec 0002 (universal scheduler delay) — `EnsureRequeueProducer` is the new trigger
- Commit `80c2117a4` — review fixes noted this as out-of-scope for the current PR
- `DisposePool()` / `DisposePoolAsync()` use `CompareExchange` to guard against cross-instance disposal, which partially mitigates but doesn't solve the root cause

## Files

- `src/Paramore.Brighter.MessagingGateway.Redis/RedisMessageGateway.cs` (static `s_pool`, constructor, `DisposePool`)
- `src/Paramore.Brighter.MessagingGateway.Redis/RedisMessageConsumer.cs` (`EnsureRequeueProducer`, `GetClient`, `GetClientAsync`)
- `src/Paramore.Brighter.MessagingGateway.Redis/RedisMessageProducer.cs` (extends `RedisMessageGateway`)

thomhurst:
## Summary

Closes #4023.

`RedisMessageGateway` stored its connection pool in a `protected static Lazy<RedisManagerPool>? s_pool` that every constructor overwrote unconditionally. Multiple coexisting gateways silently shared whichever pool was constructed last, and disposing one could leave siblings without a pool. The `EnsureRequeueProducer` path (consumer lazily constructs a producer for delayed requeue) was the most visible trigger but the issue affected any multi-instance scenario.

This change adopts the issue's Option A: each gateway owns its own `Lazy<RedisManagerPool> Pool`. The `Interlocked.CompareExchange` bookkeeping in `DisposePool`/`DisposePoolAsync` is removed (no longer needed once lifetimes are decoupled). `s_pool == null` guards in the consumer/producer call sites are removed since `Pool` is always non-null; the existing `ObjectDisposedException` catch still maps reuse-after-dispose to `ChannelFailureException`.

Why not Option C (CAS-init): the CAS winner still owns the pool, so disposing it strands every sibling — a worse failure mode than the current bug.

### Behaviour change worth flagging

- `MaxPoolSize` is now a per-gateway cap, not process-wide. A process with N gateways may hold up to `N * MaxPoolSize` connections (was effectively `MaxPoolSize` because only the latest s_pool was active). Doc updated.
- `protected static s_pool` is removed — binary-breaking for any external subclass that referenced it. Inside this repo only `RedisMessageProducer`, `RedisMessageConsumer`, and the test double `TestRedisGateway` extend the base; all updated.

## Test plan

- [x] `Paramore.Brighter.MessagingGateway.Redis` builds clean (netstandard2.0/net8.0/net9.0/net10.0).
- [x] All 17 non-broker tests in `Paramore.Brighter.Redis.Tests` pass on net9.0 and net10.0, including the two new regression facts in `When_multiple_redis_gateways_each_owns_its_own_pool`.
- [ ] CI runs the `[Trait("Category","Redis")]` integration suite against a live broker — touched call sites are only the `s_pool` -> `Pool` rename.


iancooper (approved):
Thanks Tom

lillo42 (approved):

PR #4108, opened 2026-04-29
feat: move AWS resource cleanup to dedicated workflow with scheduler support

holytshirt:
## Summary

- Moves AWS test resource cleanup from inline CI steps to a dedicated workflow (`aws-cleanup.yml`) triggered on CI completion, every 6 hours via cron, and manually via `workflow_dispatch`
- Adds EventBridge Scheduler resource cleanup (schedule groups and schedules) to the cleanup script — these were previously never cleaned up
- Fixes a `set -u` bug and adds proper dry-run support for all resource types
- Adds scheduler resource coverage to the integration test and increases wait times for tag propagation consistency

## Motivation

CI was hitting AWS SNS topic quota limits because:
1. Cleanup only ran as a post-step within each job — couldn't catch resources from concurrent runs
2. EventBridge Scheduler groups/schedules were never cleaned up
3. Tag propagation delay meant recently-created resources could be missed

## Changes

| File | Change |
|------|--------|
| `.github/workflows/aws-cleanup.yml` | New dedicated cleanup workflow |
| `.github/workflows/ci.yml` | Remove inline cleanup steps from `aws-ci` and `aws-scheduler-ci` |
| `clean_failed_tests_aws_assets.sh` | Add scheduler cleanup, fix `set -u` bug, add dry-run for default group |
| `test_clean_failed_tests_aws_assets.sh` | Add scheduler test coverage, increase sleep timers |

### DIFF
--- a/.github/workflows/aws-cleanup.yml
+++ b/.github/workflows/aws-cleanup.yml
@@ -9,6 +9,9 @@ on:
     - cron: '0 */6 * * *'
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   cleanup:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


holytshirt:
@claude  checkout is at v6 https://github.com/actions/checkout

PR #4110, opened 2026-04-29
Fix RMQ producer channel disposal

### LINKED ISSUE #4109: RmqMessageProducer async Dispose does not release base gateway resources
## Summary

`Paramore.Brighter.MessagingGateway.RMQ.Async.RmqMessageProducer.Dispose()` only detaches publisher-confirm event handlers and does not call the base gateway disposal path. That means the producer's per-instance channel/base gateway resources are not released by sync disposal.

This is separate from, but related to, #4102. #4102 covers per-gateway disposal evicting a shared pooled `IConnection`. This issue covers the opposite problem in the async producer: disposal does not release the producer's own gateway resources at all.

## Where

Async producer:

- `src/Paramore.Brighter.MessagingGateway.RMQ.Async/RmqMessageProducer.cs:197`

```csharp
public sealed override void Dispose()
{
    Channel?.BasicAcksAsync -= OnPublishSucceeded;
    Channel?.BasicNacksAsync -= OnPublishFailed;
}
```

Sync producer behaves differently and does run the base disposal path:

- `src/Paramore.Brighter.MessagingGateway.RMQ.Sync/RmqMessageProducer.cs:200`
- `src/Paramore.Brighter.MessagingGateway.RMQ.Sync/RmqMessageProducer.cs:228`

## Why it matters

Today this may partly mask #4102 for async producers because it avoids calling the base disposal path that currently removes the shared pooled connection. But once #4102 is fixed with a lease/ref-counted connection pool, this becomes a resource leak: producer disposal would detach event handlers but fail to release its channel/connection lease.

It also means async producer sync disposal is inconsistent with the sync RMQ producer and with the base `RmqMessageGateway` contract.

## Expected behaviour

Disposing an async RMQ producer should:

- detach publisher-confirm handlers;
- release its per-instance channel/base gateway resources;
- after #4102, release its connection-pool lease rather than evicting the shared connection;
- suppress finalization consistently.

`DisposeAsync()` should also be checked. The inherited base async disposal releases gateway resources, but it does not detach the producer's publisher-confirm handlers first.

## Relationship to #4102

This should probably be fixed after or alongside #4102. The correct end state is:

- per-producer disposal releases the producer's channel and lease;
- it does not tear down a shared pooled `IConnection` still used by sibling producers/consumers;
- both sync and async disposal paths have equivalent semantics.

thomhurst:
## Summary

Fix RMQ async producer shutdown so producer-owned channels are disposed in both sync and async dispose paths. Shutdown now drains active sends, waits for pending publisher confirmations, detaches confirm handlers before channel teardown, disposes the producer channel, and then delegates to the base gateway cleanup so pooled connection lifetime is still handled by the base class.

Additional correctness fixes included in this PR:

- Handle RabbitMQ publisher confirms with `multiple=true`, resolving all pending delivery tags up to the broker-confirmed tag instead of leaving earlier confirmations stuck indefinitely.
- Capture the failed channel before `ResetConnectionToBrokerAsync` on `IOException`, so confirm handlers are detached from the failed channel rather than accidentally detached from a newly recovered channel.
- Harden producer disposal with atomic guards and explicit `ObjectDisposedException` behavior for sends after disposal.
- Preserve the `BeginSend` / `CompleteSend` invariant so rejected sends after disposal do not decrement active-send tracking.

## Verification

- RMQ async Release build passed.
- `Paramore.Brighter.RMQ.Async.Tests` builds for `net10.0`.
- Targeted `RmqMessageProducerDisposedSendTests` passed: 2/2.

Fixes #4109.

### DIFF
--- a/src/Paramore.Brighter.MessagingGateway.RMQ.Async/RmqMessageGateway.cs
+++ b/src/Paramore.Brighter.MessagingGateway.RMQ.Async/RmqMessageGateway.cs
@@ -58,7 +58,7 @@ public partial class RmqMessageGateway : IDisposable, IAsyncDisposable
     private readonly AsyncPolicy _circuitBreakerPolicy;
     private readonly ConnectionFactory _connectionFactory;
     private readonly AsyncPolicy _retryPolicy;
-    private bool _disposed;
+    private int _disposed;
     protected readonly RmqMessagingGatewayConnection Connection;
     protected IChannel? Channel;
 
@@ -190,8 +190,7 @@ public partial class RmqMessageGateway : IDisposable, IAsyncDisposable
 
     public virtual async ValueTask DisposeAsync()
     {
-        if (_disposed) return;
-        _disposed = true;
+        if (Interlocked.Exchange(ref _disposed, 1) != 0) return;
 
         if (Channel != null)
         {
@@ -205,6 +204,8 @@ public partial class RmqMessageGateway : IDisposable, IAsyncDisposable
 
     protected virtual void Dispose(bool disposing)
     {
+        if (Interlocked.Exchange(ref _disposed, 1) != 0) return;
+
         if (disposing)
         {
             Channel?.AbortAsync().Wait();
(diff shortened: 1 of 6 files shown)


iancooper:
I think some of the Claude feedback is worth looking at there @thomhurst, particularly around WaitForConfirms, as these will write Dispatched to the Outbox. Now, that just means we send a message more than once (recognized failure condition) but we should avoid it if we can.

iancooper (approved):

iancooper (approved):

iancooper:
Thanks @thomhurst I think this is good to merge, if you want to.

thomhurst:
I think the Claude points above are nits. Any you want addressing before a merge?

iancooper:
> I think the Claude points above are nits. Any you want addressing before a merge?

Agreed, go ahead and merge

PR #4114, opened 2026-05-03
fix: add non-generic UseOutboxArchiver overload that infers transaction type from DI

### LINKED ISSUE #3878: UseOutboxArchiver should not require DbTransaction generic type
While migrating to V10 i stumbled opon this issue:
```csharp
// Proper UseOutboxArchiver invocation looks like this:
brighterBuilder.UseOutboxArchiver<DbTransaction>(new NullOutboxArchiveProvider(), options => options.MinimumAge = TimeSpan.FromDays(4));
// But that would require knowing the transaction type at compile time. And we don't, since there will be different transaction types for mongo and relational dbs.
```

Since we support using Postgres or MongoDB as a outbox provider - we cannot set the DbTransaction generic type during setting up the outbox archiver. I've found this workaround to work:

```csharp
typeof(HostedServiceCollectionExtensions)
    .GetMethod(nameof(HostedServiceCollectionExtensions.UseOutboxArchiver))!
    .MakeGenericMethod(transactionType)
    .Invoke(null, [
        brighterBuilder,
        new NullOutboxArchiveProvider(),
        new Action<TimedOutboxArchiverOptions>(opt => opt.MinimumAge = TimeSpan.FromDays(4))
    ]);
```
The `transactionType` is passed down from microservice configuration and can be either mongo or postgres implementation. I think you are using similar logic when registering the outbox in `AddProducers`:
https://github.com/BrighterCommand/Brighter/blob/fc3933f36b1f071bdb039e5b0f7856f430da3c75/src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs#L166-L178

I think the same (or similar) logic can be performed in the `UseOutboxArchiver` helper (although i didn't analyze this to know for sure).

Ideally, `UseOutboxArchiver` should infer the transactionType automatically (i think it was the case in v9)

DeepakVermaaa:
Closes #3878

## Problem

`UseOutboxArchiver<TTransaction>()` requires the caller to supply the transaction type as a generic type argument at compile time. For users who don't have the transaction type readily available at the call site (e.g. when wiring is done in a shared composition root), this forces awkward workarounds using manual reflection - as reported in #3878.

## Solution

Add a non-generic `UseOutboxArchiver()` convenience overload that infers the transaction type automatically by inspecting the registered `IAmABoxTransactionProvider<>` in the `IServiceCollection` - the same pattern already used by `AddProducers()` in `ServiceCollectionExtensions.cs`.

The existing generic overload is kept unchanged for full backward compatibility. Callers who need explicit control (or who hit the multi-provider case) continue to use the generic form.

## Scope

This is an **ergonomic convenience wrapper** for the single-outbox case described in #3878. It is **not** an enabler for multiple outbox providers - that requires deeper architectural changes tracked in #4133. When multiple `IAmABoxTransactionProvider<>` registrations are found, this overload throws a `ConfigurationException` directing the caller to the generic overload.

## Changes

- `src/Paramore.Brighter.Outbox.Hosting/HostedServiceCollectionExtensions.cs` - new non-generic overload + private `ResolveTransactionTypes()` helper; XML docs added to both overloads
- - `tests/Paramore.Brighter.Extensions.Tests/` - 4 new tests covering the happy path, the fallback scan for non-generic provider registrations (deferred AddProducers compatibility), the no-provider error case, and the multiple-providers error case


iancooper:
1. Deferred AddProducers compatibility -- The non-generic UseOutboxArchiver() scans for IAmABoxTransactionProvider<T> in service descriptors, but the deferred AddProducers(Func<...>) overload deliberately skips registering the generic form. Users of that overload would get a  ConfigurationException.
2. First-wins behavior with multiple providers -- The break after finding the first IAmABoxTransactionProvider<> means multi-provider setups silently pick whichever was registered first.
3. NuGet.config with <clear/> -- A new root-level NuGet.config clears all package sources and pins only nuget.org, which could break builds relying on private/internal feeds. This seems unrelated to the PR's purpose. We should not commit this as part of this change.

iancooper:
Thanks for the PR though, let's review the comments and see if we can get this accepted

iancooper:
@DeepakVermaaa Just a nudge on the feedback. It's a good PR, let's get it fixed up.

DeepakVermaaa:
Hi @iancooper, thanks for the detailed review! I've addressed all three points:

NuGet.config removed - that was a local setup workaround, shouldn't have been committed.

Multiple providers - replaced the silent first-wins break with a HashSet that collects all found transaction types. Now throws ConfigurationException with a clear message listing the conflicting types and pointing to the generic overload.

Deferred AddProducers compatibility - added a fallback scan for non-generic IAmABoxTransactionProvider descriptors with a concrete ImplementationType. If the registration is factory-based (as the deferred overload uses), it correctly falls through to the ConfigurationException with a message explaining to use the generic overload instead.

Also added two new tests covering the fallback path and the multiple-providers error case (4 tests total).

iancooper:
Sorry for the delay, we had to wait for MongoDb to fix a dependency with a GHSA, now we are seeing if we can get a build

iancooper (commented):
First thanks for the PR. Second, I don't think that this PR actually enables what it says i.e. multiple outboxes for Brighter. That is actually a deeper requirement because we currently don't support that internally. 

Due to these constraints, UseOutboxArchiver can't really support multiple providers i.e. we are forced to refer folks back to the generic version.

So, what this fix offer us is a convenience wrapper for issues like: https://github.com/BrighterCommand/Brighter/issues/3878

We could reframe it as that, if you wanted?

DeepakVermaaa:
Thanks @iancooper - that framing is fair. You're right that this doesn't actually enable multi-outbox; it just removes the compile-time generic requirement for the single-outbox case in #3878. I've updated the PR description to reflect that scope. Happy to adjust further if you'd like the commit message reworded as well.

iancooper:
We are contemplating whether this adds value or actually makes it more confusing. I appreciate that it addresses a specific piece of user feedback, but given the constraints, I am not sure this requires a generic solution. I think we would need to better understand the use case, which implies that you don't know the transaction type when you create the archiver. If you have multiple outboxes (which we don't support yet), you would just use multiple archivers.

iancooper:
Going to close as not adding value

If you are looking to contribute, perhaps start with https://github.com/BrighterCommand/Brighter/issues/4175

PR #4116, opened 2026-05-05
[Feat] InMemoryBox Expiry and Outbox limits

### LINKED ISSUE #4115: InMemoryBox has an upper limit of 2048 items
The code for InMemoryBox, the base class for InMemoryInbox and InMemoryOutbox, limits the box to 2048 items before it begins reclaiming space by compacting it by the compaction percentage.

This value is required because otherwise, the inbox or outbox would grow for the entire lifetime of the program.

But, for production use cases, this value may be too low, as at scale, we may have more than 2048 items added before they are processed. 

This results in outbox errors when we try to dispatch an item that cannot be found. 

In addition, when OutboxProducerMediator cannot find an item in the Outbox, we throw a NullReferenceException.

We can set this limit, as it is a public property of InMemoryBox, but when the InMemoryOutbox is provided by ServiceCollectionExtensions.AddProducers (l.296 and l.430) because no Outbox was specified, we get the default settings and can't alter them. 

To fix this, we would need to add properties to ProducersConfiguration that allow us to configure the InMemoryOutbox if no outbox is explicitly provided (or use the default values if these are not set).

We should also allow an EntryLimit of -1, to not compact.

We can make two fixes to make this more robust:

(1) We should make RemoveExpiredMessages abstract protected virtual. We can allow the InMemoryOutbox to override and remove any messages that have a TimeFlushed in the past. We should run this even if -1 is set. We can allow InMemoryInbox to override with the existing algorithm, unless -1 is set.
(2) When OutboxProducerMediator cannot find a message, we should log an error rather than throw a null exception.

Do nothing and just push users to provide an InMemoryOutbox configured to their needs

iancooper:
## Description

The InMemoryOutbox (and InMemoryInbox) allow configuring their entry limits and compaction (designed to prevent the box from growing indefinitely during the program's lifetime), but they do not allow configuration when InMemoryOutbox is the default because no Outbox has been chosen.

The eviction of elements from the InMemoryOutbox should restrict itself to messages that have been dispatched

## Related Issues

closes #4115 

## Type of Change

- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update

## Checklist

- [X] I have read the [Contributing Guide](CONTRIBUTING.md)
- [X] I have checked the [documentation](https://brightercommand.gitbook.io/paramore-brighter-documentation/) for relevant guidance
- [X] I have added/updated XML documentation for any public API changes
- [X] I have added/updated tests as appropriate
- [X] My changes follow the existing code style and conventions


iancooper:
## Response to Code Review (Parts 1–3)

Thanks for the thorough review. Addressed items 1–4 in commit e04a2889f. Here's the full breakdown:

### Fixed

**1) CompactionPercentage default (0.0 vs 0.5)** — The intended default is 0.5. The ADR pseudocode and `InMemoryBoxConfiguration` record doc were documenting the old implicit `double` default (0.0) rather than the intentional value. Fixed: ADR now documents 0.5 as the default throughout.

**2) EnforceCapacityLimit arithmetic** — The ADR pseudocode was stale (`newSize = EntryCount * CompactionPercentage; entriesToRemove = EntryLimit - newSize`). The implementation (`newSize = EntryLimit * CompactionPercentage; entriesToRemove = EntryCount - newSize`) is the correct form — it targets a fraction of the *limit* and removes enough from the *current count* to reach that target. ADR pseudocode updated to match.

**3) EntryTimeToLive XML doc** — Updated the base class doc to be subclass-neutral: "The reference time depends on the subclass: the outbox measures from dispatch time (`TimeFlushed`), while the inbox measures from write time."

**4) Cooldown set before lock acquisition** — Moved `_lastCompactionAttemptAt` assignment inside `RunCompact` after `Monitor.TryEnter` succeeds. Now a failed lock acquisition doesn't stall compaction for a full `ExpirationScanInterval`. ADR pseudocode updated to match.

### Acknowledged — not addressing in this PR

**5) Unobserved background task exceptions** — Pre-existing on master. Agree it's worth adding `ContinueWith(OnlyOnFaulted)` logging, but out of scope for this PR.

**6) Missing log assertion** — The test verifies no-throw and correct dispatch behaviour. The log assertion would tighten the contract but FR-6 is implemented; this is a nice-to-have.

**7) Negative tests rely on fixed sleep** — Testing "nothing happened" is inherently timing-dependent. The 200ms delay is reasonable for unit tests and the time-check short-circuits make false passes unlikely.

**8) EnforceCapacityLimit in InMemoryInbox.Add** — Already documented in the ADR Negative Consequences section as a deliberate design choice.

**9) Non-atomic timestamp reads** — Pre-existing pattern. Worst case is an extra no-op background task spawned, which is harmless given the `Monitor.TryEnter` guard.

**10) Naming: InMemoryBoxConfiguration** — The XML doc already says "Configuration for the default InMemoryOutbox". The generic name leaves room for future inbox configuration without a rename.

**11) Spec workflow dotfiles** — These are committed across all 20+ specs in the repo. It's a repo-wide pattern, not specific to this PR.

**12) Minor polish** — The bulk Add calling per-message Add does N time-checks but they all short-circuit after the first. The `posts.ToArray().Length` is pre-existing code.

PR #4117, opened 2026-05-05
feat(sqs.v4): bypass string Body via raw bytes receive path

slang25:
## Description

The standard AWS SDK v4 `ReceiveMessageAsync` path materializes each message body as a `System.String`. Brighter then immediately re-encodes that string back to UTF-8 bytes inside `MessageBody`, so every received message pays for a string allocation plus an encoding round-trip — pure waste, since the bytes already came off the wire as UTF-8 and are about to be consumed as bytes by the rest of the pipeline (#4113).

This PR routes the SQS receive call through a custom unmarshaller that captures the body's UTF-8 bytes directly into a `ReadOnlyMemory<byte>` and feeds them straight to `MessageBody(in ReadOnlyMemory<byte>, …)`. The compressed-payload branch decodes Base64 in place over the same buffer (`Base64.DecodeFromUtf8InPlace` + slice), so a compressed message is allocation-free beyond the initial body buffer. The custom unmarshaller is reached via an internal `RawAmazonSQSClient` subclass (the only way to access the SDK's `protected InvokeAsync<T>`) constructed in parallel with `CreateSqsClient` from the same credentials and config — the public `AWSClientFactory.CreateSqsClient` contract stays as plain `AmazonSQSClient`, so a future user-supplied factory needs no awareness of the bypass.

Buffers are copied into fresh arrays rather than rented from `ArrayPool`: `MessageBody` has no `IDisposable` and `RequestContext.OriginatingMessage` exposes `Body.Memory` to handlers with no lifetime contract, so a pooled buffer would leak whenever a handler captured a reference past the pump iteration.

## Type of Change

## Checklist

- [ ] I have read the [Contributing Guide](CONTRIBUTING.md)
- [ ] I have added/updated tests as appropriate
- [ ] My changes follow the existing code style and conventions

## Additional Notes

V4 unit tests build clean across netstandard2.0/net8/net9/net10. Integration tests need Moto/LocalStack and weren't run.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -344,3 +344,5 @@ ralph.log
 
 # AI agent session state
 PROMPT.md
+
+*.lscache
--- a/src/Paramore.Brighter.MessagingGateway.AWSSQS.V4/AWSClientFactory.cs
+++ b/src/Paramore.Brighter.MessagingGateway.AWSSQS.V4/AWSClientFactory.cs
@@ -29,6 +29,7 @@ using Amazon.Runtime;
 using Amazon.SecurityToken;
 using Amazon.SimpleNotificationService;
 using Amazon.SQS;
+using Paramore.Brighter.MessagingGateway.AWSSQS.V4.Internal;
 
 namespace Paramore.Brighter.MessagingGateway.AWSSQS.V4;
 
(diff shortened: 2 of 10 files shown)


iancooper:
Thanks @slang25 I need to parse that, and check it's aligned with the recent changes here to use Span<T> in Message. I assume that might be the why if this PR - we can now pick up on improvements in AWS4.

I'll peek tomorrow

slang25:
@iancooper it's more of an experiment, I'm not sure it's worth the saving. Although it fits quite neatly here Vs JustSaying

slang25:
I'm looking at AWSSDK performance more broadly, and it turns out this is not the current bottleneck, so won't be revisiting this for a long while. There are good things coming via the SDK directly though.

PR #4120, opened 2026-05-06
Clarify DontAckAction transport semantics

### LINKED ISSUE #4097: DontAckAction triggers immediate redelivery via Nack, contradicting its name and test comments
## Summary

`DontAckAction` is documented as "don't acknowledge" the message — implying passive behavior where the broker's ack timeout / visibility timeout would later trigger redelivery. In practice, however, the Proactor (and likely Reactor) actively call `Channel.NackAsync(message)` when handling `DontAckAction`, which on `InMemoryMessageConsumer` re-enqueues the message **immediately** — bypassing the ack timeout entirely.

## Where

- `src/Paramore.Brighter.ServiceActivator/Proactor.cs:296` — `await Channel.NackAsync(message);` inside the `dontAck != null` branch
- `src/Paramore.Brighter/InMemoryMessageConsumer.cs:135-139` — `Nack` removes from `_lockedMessages` and `_bus.Enqueue(message)` instantly

```csharp
public void Nack(Message message)
{
    _lockedMessages.TryRemove(message.Id, out _);
    _bus.Enqueue(message);  // immediate re-enqueue
}
```

## Symptom

The test `When_A_Handler_Throws_DontAck_Action_Should_Not_Acknowledge_Async` (added in #4019) asserts `SendCount == 1` after the first delivery, then advances `FakeTimeProvider` past the ack timeout and asserts `SendCount >= 2`. The comment in the test reads:

> the ack timeout timer requeues it and the pump re-delivers it.

But the ack timeout never fires here — `Nack` already removed the message from `_lockedMessages` and put it back on the bus. The pump tight-loops and re-delivers instantly. With xUnit's slower scheduling the test thread usually wins the race and observes `SendCount == 1`. Under TUnit the pump can already be on the second delivery before the test thread resumes from the `WaitForHandle` semaphore, producing `SendCount == 2` and a failed assertion.

This was observed during the TUnit migration in #4065. The test had to be weakened from `== 1` to `>= 1` to be deterministic.

## Possible interpretations

1. **Code bug:** the `DontAckAction` branch should NOT call `Nack`. It should leave the message locked and rely on the ack-timeout requeue path (`CheckLockedMessages`). That matches the name "don't ack", matches the test comments, and matches typical broker semantics where unacknowledged messages reappear after a visibility timeout.
2. **Test/comment bug:** the immediate Nack is intentional; the test comments are misleading and should be rewritten. In this case, `DontAckAction` is effectively a synonym for "reject and immediately retry forever (until `UnacceptableMessageLimit`)", which seems redundant with `DeferMessageAction(delay: 0)`.

For real brokers (RabbitMQ, SQS) Nack semantics vary, so the in-memory implementation may be hiding the real intent.

## Suggested next steps

- Decide which interpretation is correct.
- If (1): change Proactor/Reactor to skip the Nack call on `DontAckAction`, document the contract, fix the test comments, and remove the race in the test.
- If (2): rewrite the test comments and consider whether `DontAckAction` and `DeferMessageAction(0)` should be unified.

## Repro

Run `tests/Paramore.Brighter.Core.Tests/MessageDispatch/Proactor/When_a_handler_throws_dont_ack_action_should_not_acknowledge_async.cs` under TUnit on a fast CPU. The first `SendCount == 1` assertion fails intermittently with "Expected to be 1 but found 2".

thomhurst:
## Summary
- Clarifies `DontAckAction` and `DontAckOnError` docs to describe transport-specific not-acknowledged behavior.
- Updates the Proactor DontAck test comments/assertion to reflect that `InMemoryMessageConsumer` nacks and re-enqueues immediately, rather than relying on the ack timeout.

Fixes #4097

## Test
- `dotnet test tests\Paramore.Brighter.Core.Tests\Paramore.Brighter.Core.Tests.csproj --filter "FullyQualifiedName~MessagePumpCommandDontAckActionTestsAsync"`

### DIFF
--- a/src/Paramore.Brighter/Actions/DontAckAction.cs
+++ b/src/Paramore.Brighter/Actions/DontAckAction.cs
@@ -27,16 +27,17 @@ using System;
 namespace Paramore.Brighter.Actions;
 
 /// <summary>
-/// Thrown to indicate that a message should not be acknowledged. The message remains on the channel and will be
-/// presented again on the next pump iteration. Call <c>throw new DontAckAction()</c> from within a
+/// Thrown to indicate that a message should not be acknowledged as successfully handled. A message pump translates
+/// this into the channel-specific not-acknowledged behavior. Call <c>throw new DontAckAction()</c> from within a
 /// <see cref="RequestHandler{TRequest}"/> or <see cref="RequestHandlerAsync{TRequest}"/> to prevent acknowledgment.
 /// </summary>
 /// <remarks>
-/// Unlike <see cref="DeferMessageAction"/> which requeues the message, and <see cref="RejectMessageAction"/> which
-/// moves the message to a dead letter queue, <see cref="DontAckAction"/> leaves the message completely unacknowledged
-/// on the channel. The message will be re-delivered by the transport after its visibility timeout expires.
-/// A configurable delay on the pump prevents tight-loop CPU burn. The unacceptable message count is incremented,
-/// allowing the pump to eventually stop if a limit is configured.
+/// Unlike <see cref="DeferMessageAction"/> which explicitly schedules a requeue, and <see cref="RejectMessageAction"/>
+/// which moves the message to a dead letter queue, <see cref="DontAckAction"/> asks the pump to use the transport's
+/// not-acknowledged disposition. For <see cref="InMemoryMessageConsumer"/> this nacks the message and immediately
+/// makes it available for redelivery; other transports may redeliver later or use their own offset or disposition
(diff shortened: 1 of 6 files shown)


iancooper (approved):

iancooper:
Thanks. Feel free to merge.

iancooper (approved):

iancooper:
Awaiting MongoDb fix for the build: https://github.com/mongodb/mongo-csharp-driver/pull/1994

PR #4128, opened 2026-05-11
fix(aws): handle null MessageAttributes/Attributes in V4 SQS message creators

### LINKED ISSUE #4126: AWSSQS.V4: SqsMessageCreator throws NullReferenceException when inbound SQS message has no MessageAttributes
## Summary

`Paramore.Brighter.MessagingGateway.AWSSQS.V4.SqsMessageCreator.CreateMessage`
throws `NullReferenceException` for any inbound SQS message that has zero
`MessageAttributes` set. The exception is caught inside `CreateMessage` and
swallowed into a `Message.FailureMessage(...)` with `MessageType =
MT_UNACCEPTABLE`, so the message pump rejects every such message and the
handler is never invoked.

The trigger is benign: an SQS message simply published without any
MessageAttributes (e.g. via `aws sqs send-message --message-body '...'`, an
EventBridge → SQS target with no attribute transform, an SNS → SQS
subscription where the publisher didn't set any, or anything that isn't a
Brighter producer). Under the v3 SDK these were consumable; under v4 they
aren't.

## Root cause

AWS SDK for .NET v4 changed the default for unset collection properties: they
are now `null` rather than empty collections. So
`Amazon.SQS.Model.Message.MessageAttributes` is `null` when the message
carries no attributes, instead of an empty `Dictionary<string,
MessageAttributeValue>`.

`SqsMessageCreator.ReadTopic` (and every sibling reader called from
`CreateMessage`) does:

```csharp
private static HeaderResult<RoutingKey> ReadTopic(Amazon.SQS.Model.Message sqsMessage)
{
    if (sqsMessage.MessageAttributes.TryGetValue(HeaderNames.Topic, out MessageAttributeValue? value))
    { ... }
}
```

`ReadTopic` is the first reader invoked from `CreateMessage`, so it explodes
first. The catch block at the bottom of `CreateMessage` turns the NRE into
`Message.FailureMessage(...)`. Result: silent message rejection.

[`src/Paramore.Brighter.MessagingGateway.AWSSQS.V4/SqsMessageCreator.cs`](https://github.com/BrighterCommand/Brighter/blob/master/src/Paramore.Brighter.MessagingGateway.AWSSQS.V4/SqsMessageCreator.cs)
(the `ReadTopic` method, and every other `Read*` method that touches
`sqsMessage.MessageAttributes`).

Observed in version **10.4.1**. The non-V4 (`AWSSQS`) gateway is not affected
because AWS SDK v3 returns an empty dictionary.

## Reproduction

1. Stand up a Brighter consumer using `Paramore.Brighter.MessagingGateway.AWSSQS.V4`
   on any SQS queue (`ChannelType.PointToPoint` or `ChannelType.PubSub`, either works).
2. Publish a message to that queue from outside Brighter with no
   MessageAttributes — e.g.:
   ```
   aws sqs send-message --queue-url <url> --message-body '{"hello":"world"}'
   ```
3. Observe in logs:
   ```
   warn: Paramore.Brighter.MessagingGateway.AWSSQS.V4.SqsMessageCreator
         Failed to create message from amqp message
         System.NullReferenceException: Object reference not set to an instance of an object.
         at Paramore.Brighter.MessagingGateway.AWSSQS.V4.SqsMessageCreator.ReadTopic(Message sqsMessage)
         at Paramore.Brighter.MessagingGateway.AWSSQS.V4.SqsMessageCreator.CreateMessage(Message sqsMessage)
   warn: Paramore.Brighter.ServiceActivator.MessagePump
         MessagePump: Failed to parse a message ...
         MessagePump: Rejecting message ...
   ```
   The handler is never invoked.

## Expected behaviour

A message with no SQS MessageAttributes should be parsed using the default
fallback values that the individual `Read*` methods already implement (random
`Id`, `MT_EVENT`, `text/plain`, empty topic, `UtcNow`, etc.) and the body
handed to the registered mapper. This is what the v3 gateway does today.

## Suggested fix

Treat a `null` `MessageAttributes` as equivalent to an empty dictionary at
the entry to each reader. Either:

- **Localised**: guard each `Read*` with
  `if (sqsMessage.MessageAttributes is null) return HeaderResult<...>.Empty();`
  (or equivalent fallback for that field).
- **Centralised** (smaller diff, less repetition): normalise once at the top
  of `CreateMessage`:
  ```csharp
  sqsMessage.MessageAttributes ??= new Dictionary<string, MessageAttributeValue>();
  ```
  before any reader is called.

Either approach makes the V4 gateway behave the same as V3 for
attribute-less messages.

## Workaround for consumers today

Until a fix lands, ensure every producer attaches at least one SNS/SQS
MessageAttribute (any key/value) so the dictionary is non-null on arrival.
Brighter's own producers already do this; only foreign producers need
adjustment.

slang25:
## Description

AWS SDK for .NET v4 changed the default for unset collection properties: they are now `null` rather than empty collections. As a result, `SqsMessageCreator.CreateMessage` in the `AWSSQS.V4` gateway throws `NullReferenceException` for any inbound SQS message that has no `MessageAttributes` (e.g. messages from `aws sqs send-message`, EventBridge → SQS targets, or any non-Brighter producer). The exception was swallowed into a `Message.FailureMessage(...)`, so every such message was silently rejected and the handler never invoked. This PR normalises `MessageAttributes` and `Attributes` to an empty dictionary at the top of `CreateMessage` in both `SqsMessageCreator` and `SqsInlineMessageCreator`, restoring v3-equivalent behaviour. The `??=` only allocates on the bug path (when the property is null); the happy path is unchanged.

## Related Issues

Fixes #4126

## Type of Change

- [ ] Documentation update

## Checklist

- [x] I have read the [Contributing Guide](CONTRIBUTING.md)
- [ ] I have added/updated tests as appropriate

## Additional Notes

There are no existing unit tests for `SqsMessageCreator`/`SqsInlineMessageCreator` (only AWS integration tests) — happy to add a regression test on request.

### DIFF
--- a/src/Paramore.Brighter.MessagingGateway.AWSSQS.V4/SqsInlineMessageCreator.cs
+++ b/src/Paramore.Brighter.MessagingGateway.AWSSQS.V4/SqsInlineMessageCreator.cs
@@ -49,9 +49,6 @@ internal sealed partial class SqsInlineMessageCreator : SqsMessageCreatorBase, I
 
         try
         {
-            // AWS SDK v4 returns null (not an empty dictionary) when Attributes was never set on the wire.
-            sqsMessage.Attributes ??= new Dictionary<string, string>();
-
             var jsonDocument = JsonDocument.Parse(sqsMessage.Body);
             _messageAttributes = ReadMessageAttributes(jsonDocument);
 
@@ -409,7 +406,8 @@ internal sealed partial class SqsInlineMessageCreator : SqsMessageCreatorBase, I
 
     private static HeaderResult<PartitionKey> ReadPartitionKey(Amazon.SQS.Model.Message sqsMessage)
     {
-        if (sqsMessage.Attributes.TryGetValue(MessageSystemAttributeName.MessageGroupId, out var value))
+        if (sqsMessage.Attributes is not null
+            && sqsMessage.Attributes.TryGetValue(MessageSystemAttributeName.MessageGroupId, out var value))
         {
             //we have an arn, and we want the topic
             return new HeaderResult<PartitionKey>(value, true);
@@ -420,7 +418,8 @@ internal sealed partial class SqsInlineMessageCreator : SqsMessageCreatorBase, I
 
     private static HeaderResult<string> ReadDeduplicationId(Amazon.SQS.Model.Message sqsMessage)
     {
-        if (sqsMessage.Attributes.TryGetValue(MessageSystemAttributeName.MessageDeduplicationId, out var value))
+        if (sqsMessage.Attributes is not null
+            && sqsMessage.Attributes.TryGetValue(MessageSystemAttributeName.MessageDeduplicationId, out var value))
         {
             //we have an arn, and we want the topic
             return new HeaderResult<string>(value, true);
--- a/src/Paramore.Brighter.MessagingGateway.AWSSQS.V4/SqsInlineMessageCreator.cs
+++ b/src/Paramore.Brighter.MessagingGateway.AWSSQS.V4/SqsInlineMessageCreator.cs
@@ -49,6 +49,9 @@ internal sealed partial class SqsInlineMessageCreator : SqsMessageCreatorBase, I
 
         try
         {
+            // AWS SDK v4 returns null (not an empty dictionary) when Attributes was never set on the wire.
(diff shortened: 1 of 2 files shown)


iancooper (approved):
Thanks @slang25

iancooper:
CI is failing for the MongoDB issue

PR #4130, opened 2026-05-11
fix(aws): surface unknown SNS/SQS message attributes into MessageHeader.Bag

### LINKED ISSUE #4129: Transformers.JustSaying: mapper doesn't handle JustSaying's gzip+base64 compressed messages
## Summary

`Paramore.Brighter.Transformers.JustSaying.JustSayingMessageMapper<T>.MapToRequest`
(and its async counterpart) calls `JsonSerializer.Deserialize<T>(message.Body...)`
unconditionally. JustSaying has a built-in publish-side compression feature; when
it kicks in, the SQS body is `base64(gzip(json))` rather than JSON, and the mapper
throws `JsonException` before the handler is ever invoked.

In other words: the transformer package covers JustSaying's uncompressed wire
format but is unaware of the compressed one. Any JustSaying producer with
compression enabled (or a body large enough to cross the default threshold) is
unconsumable by a Brighter subscriber using this mapper today.

## Failure mode

Consumer logs (v10.4.1):

```
warn: Paramore.Brighter.ServiceActivator.MessagePump
      MessagePump: Failed to map message <id> from <queue> with <topic> on thread # N
      Paramore.Brighter.ServiceActivator.MessageMappingException:
        Failed to map message <id> of <T> using transform pipeline
        ---> System.Text.Json.JsonException:
             'H' is an invalid start of a value.
             Path: $ | LineNumber: 0 | BytePositionInLine: 0.
        at Paramore.Brighter.Transformers.JustSaying.JustSayingMessageMapper`1
           .MapToRequest(Message message) line 222
        at Paramore.Brighter.Transformers.JustSaying.JustSayingMessageMapper`1
           .MapToRequestAsync(Message message, CancellationToken)        line 53
        at Paramore.Brighter.UnwrapPipelineAsync`1.UnwrapAsync(...)
        at Paramore.Brighter.ServiceActivator.Proactor.TranslateMessage(...)
```

The leading `'H'` is the first character of the base64 of a gzip stream
(`H4sIAAAA…` — gzip's magic bytes `1F 8B 08 00` encode to `H4sI`).

## JustSaying's compression wire format

Set per-publication via `SnsWriteConfiguration.CompressionOptions` /
`SqsWriteConfiguration.CompressionOptions`, or bus-wide via
`MessagingConfig.DefaultCompressionOptions`. Defaults:
`MessageLengthThreshold = 254 * 1024` bytes (computed across body + attribute
keys + values), `CompressionEncoding = "gzip,base64"`.

When the threshold is met, JustSaying:

1. Serializes the message to JSON.
2. Gzips the UTF-8 JSON bytes, then base64-encodes them.
3. Sends the base64 string as the SNS/SQS body (or as the inner `Message` of
   the SNS envelope for non-raw-delivery topics).
4. Adds an SNS MessageAttribute with key `"Content-Encoding"` (constant
   `JustSaying.AwsTools.MessageAttributeKeys.ContentEncoding`),
   `DataType = "String"`, `StringValue = "gzip,base64"` (constant
   `JustSaying.Messaging.Compression.ContentEncodings.GzipBase64`).

Key files in `justeattakeaway/JustSaying`:

- `src/JustSaying/AwsTools/MessageHandling/OutboundMessageConverter.cs`
  (`CompressMessageBody`, `CalculateTotalMessageSize`)
- `src/JustSaying/AwsTools/MessageHandling/InboundMessageConverter.cs`
  (`ApplyBodyDecompression` — the symmetric receive-side decode)
- `src/JustSaying/AwsTools/MessageHandling/PublishCompressionOptions.cs`
- `src/JustSaying/Messaging/Compression/GzipMessageBodyCompression.cs`
- `src/JustSaying/Messaging/Compression/ContentEncodings.cs`
- `src/JustSaying/AwsTools/MessageAttributeKeys.cs`

## What the Brighter adapter does today

`src/Paramore.Brighter.Transformers.JustSaying/JustSayingMessageMapper.cs`:

```csharp
// line 222 (sync), called from MapToRequestAsync at line 53
public T MapToRequest(Message message)
    => JsonSerializer.Deserialize<T>(message.Body.Memory.Span, ...);
```

No inspection of `Content-Encoding`, no decompression branch. Brighter's
generic `CompressPayloadTransformer` doesn't help here either — it signals
compression via `Header.ContentType = "application/gzip"` (not a
`Content-Encoding` MessageAttribute), so attaching `[Decompress]` to the
mapper would not match JustSaying's wire format.

## Reproduction

1. Configure any JustSaying 8.x publisher with compression and a small
   threshold:
   ```csharp
   x.WithTopic<MyEvent>(t => t.WithWriteConfiguration((SnsWriteConfiguration cfg) =>
   {
       cfg.CompressionOptions = new PublishCompressionOptions
       {
           MessageLengthThreshold = 1024,
           CompressionEncoding    = ContentEncodings.GzipBase64,
       };
   }));
   ```
2. Publish a message whose serialized size + attributes ≥ 1024 bytes.
3. Subscribe in Brighter using `Paramore.Brighter.Transformers.JustSaying`
   with `JustSayingMessageMapper<MyEvent>` as the mapper for `MyEvent`.
4. The message arrives in SQS, Brighter's pump hands the body to the mapper,
   and `MapToRequest` throws `JsonException`. The handler never runs.

## Expected behaviour

Symmetric with JustSaying's `InboundMessageConverter.ApplyBodyDecompression`:
when the inbound message carries `Content-Encoding == "gzip,base64"`, the
mapper should base64-decode → gunzip → UTF-8 the body before
`JsonSerializer.Deserialize`. Ideally the publish side (`MapToMessage`)
mirrors the same logic so Brighter producers can interop with JustSaying
consumers that expect compressed payloads.

## Suggested implementation

Two options, either acceptable:

1. **Inline in `JustSayingMessageMapper<T>`.** Add a small helper that peeks
   at the `Content-Encoding` header in the Brighter `MessageHeader.Bag` (or
   wherever the SQS gateway surfaces inbound attributes), and if equal to
   `"gzip,base64"`, decode before deserialising. For publishing, accept a
   threshold + encoding (or a `PublishCompressionOptions`-shaped record) on
   the mapper / publication, and when the serialised body crosses it,
   gzip+base64 the body and attach the `Content-Encoding` header so the SQS
   producer sets it as a MessageAttribute on the outgoing SNS publish.

2. **Dedicated transform.** A `JustSayingCompressionTransform` that runs
   before the mapper on the unwrap pipeline and after the mapper on the
   wrap pipeline. Cleaner separation, plays naturally with Brighter's
   existing transform attribute model.

Either way, the encoding string and attribute key should be matched
verbatim to JustSaying's constants
(`"Content-Encoding"`, `"gzip,base64"`).

## Related

A separate AWSSQS.V4 issue affects the path leading up to this mapper when
no MessageAttributes are set (#4126). When JustSaying compresses it
attaches `Content-Encoding`, so `MessageAttributes` isn't null and #4126
doesn't bite — but the JSON parse failure documented here is hit
immediately after.

slang25:
## Description

The AWS SQS gateway previously read a fixed set of known SNS/SQS `MessageAttribute` names into typed `MessageHeader` fields (and Brighter's own JSON-serialised `bag` attribute), silently discarding every other inbound attribute. That made Brighter consumers blind to interop metadata sent by foreign producers — the JustSaying gzip+base64 compression flag (`Content-Encoding: gzip,base64`) being the motivating example. Both V3 and V4 message creators (raw + inline paths) now copy any inbound `MessageAttribute` not in `HeaderNames.IsKnown` into `MessageHeader.Bag` under its raw name, with Brighter's own JSON bag attribute taking precedence on key collision. On top of that, `JustSayingMessageMapper.MapToRequest` now reads `Content-Encoding` from the bag and decompresses `base64(gzip(json))` bodies before deserialising.

## Related Issues

Fixes #4129

## Type of Change

- [ ] Documentation update

## Checklist

- [x] I have read the [Contributing Guide](CONTRIBUTING.md)

## Additional Notes

The publish-side of JustSaying compression is intentionally not in scope — the gateway publisher emits Brighter's bag as a single JSON attribute rather than per-key attributes, so emitting a native `Content-Encoding` SNS attribute would need a separate publisher change. Worth a follow-up issue.

Unit-test coverage for the gateway attribute-surfacing change lives transitively in the JustSaying mapper tests (which exercise the bag-lookup contract). `SqsMessageCreator` is `internal sealed` with no `InternalsVisibleTo`, and the existing gateway test suite is integration-only (LocalStack); happy to add a LocalStack integration test if reviewers want belt-and-braces.

### DIFF
--- a/src/Paramore.Brighter.Transformers.JustSaying/JustSayingMessageMapper.cs
+++ b/src/Paramore.Brighter.Transformers.JustSaying/JustSayingMessageMapper.cs
@@ -1,7 +1,5 @@
 using System;
 using System.Collections.Generic;
-using System.IO;
-using System.IO.Compression;
 using System.Net.Mime;
 using System.Text.Json;
 using System.Threading;
@@ -50,6 +48,7 @@ public class JustSayingMessageMapper<TMessage> : IAmAMessageMapper<TMessage>, IA
     }
 
     /// <inheritdoc />
+    [JustSayingDecompress(0)]
     public Task<TMessage> MapToRequestAsync(Message message, CancellationToken cancellationToken = default)
     {
         return Task.FromResult(MapToRequest(message));
@@ -220,48 +219,12 @@ public class JustSayingMessageMapper<TMessage> : IAmAMessageMapper<TMessage>, IA
     }
 
     /// <inheritdoc />
+    [JustSayingDecompress(0)]
     public TMessage MapToRequest(Message message)
-    {
-        var body = message.Body.Memory;
-        if (IsJustSayingGzipBase64(message))
-        {
-            return DeserializeCompressed(body);
-        }
-
 #if NETSTANDARD2_0
-        return JsonSerializer.Deserialize<TMessage>(body.ToArray(), JsonSerialisationOptions.Options)!;
+        => JsonSerializer.Deserialize<TMessage>(message.Body.Memory.ToArray(), JsonSerialisationOptions.Options)!;
 #else
-        return JsonSerializer.Deserialize<TMessage>(body.Span, JsonSerialisationOptions.Options)!;
+        => JsonSerializer.Deserialize<TMessage>(message.Body.Memory.Span, JsonSerialisationOptions.Options)!;
 #endif
-    }
-
-    private static bool IsJustSayingGzipBase64(Message message)
-    {
-        // JustSaying signals compression via a "Content-Encoding" SNS message attribute. The AWS
-        // SQS gateway surfaces foreign message attributes into MessageHeader.Bag under their raw
-        // name, so we look it up there.
-        return message.Header.Bag.TryGetValue(JustSayingAttributesName.ContentEncoding, out var encoding)
-               && encoding is string s
-               && string.Equals(s, JustSayingAttributesName.GzipBase64ContentEncoding, StringComparison.OrdinalIgnoreCase);
-    }
-
-    private static TMessage DeserializeCompressed(ReadOnlyMemory<byte> body)
-    {
-        // Body is base64(gzip(json)); base64-decode then gunzip before deserializing.
-        var base64 = System.Text.Encoding.ASCII.GetString(body.ToArray());
-        var gzipped = Convert.FromBase64String(base64);
-
-        using var input = new MemoryStream(gzipped, writable: false);
-        using var gz = new GZipStream(input, CompressionMode.Decompress);
-        using var output = new MemoryStream();
-        gz.CopyTo(output);
-
-        if (output.TryGetBuffer(out var buffer))
-        {
-            return JsonSerializer.Deserialize<TMessage>(buffer.AsSpan(), JsonSerialisationOptions.Options)!;
-        }
-
-        return JsonSerializer.Deserialize<TMessage>(output.ToArray(), JsonSerialisationOptions.Options)!;
-    }
 }
 
--- a/src/Paramore.Brighter.Transformers.JustSaying/JustSayingMessageMapper.cs
+++ b/src/Paramore.Brighter.Transformers.JustSaying/JustSayingMessageMapper.cs
@@ -1,5 +1,7 @@
 using System;
 using System.Collections.Generic;
+using System.IO;
+using System.IO.Compression;
 using System.Net.Mime;
 using System.Text.Json;
 using System.Threading;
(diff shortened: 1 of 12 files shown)


src/Paramore.Brighter.Transformers.JustSaying/JustSayingMessageMapper.cs
```diff
@@ -218,11 +220,48 @@ private static DateTimeOffset GetTimeStamp(DateTimeOffset currentValue)
...
 #if NETSTANDARD2_0
-        => JsonSerializer.Deserialize<TMessage>(message.Body.Memory.ToArray(), JsonSerialisationOptions.Options)!;
+        return JsonSerializer.Deserialize<TMessage>(body.ToArray(), JsonSerialisationOptions.Options)!;
 #else
-        => JsonSerializer.Deserialize<TMessage>(message.Body.Memory.Span, JsonSerialisationOptions.Options)!;
+        return JsonSerializer.Deserialize<TMessage>(body.Span, JsonSerialisationOptions.Options)!;
 #endif
+    }
+
+    private static bool IsJustSayingGzipBase64(Message message)
```
iancooper: We should handle this as a transformer, since that is how we do Brighter compression, so it's more aligned. 

You add an attribute to the mapper and a transformer that maps to it, thereby picking up the incoming (or outgoing) message. You can pick up properties from the MessageHeader (provided we don't lose them during mapping, which you should have fixed) and use them with the mapper.  Use them to decide whether you need to decompress the body before passing on to the mapper.

Take a look at: src/Paramore.Brighter/Transforms

iancooper (commented):
Thanks for picking this one up. For compression, though, we should use a transformer.

src/Paramore.Brighter.Transformers.JustSaying/JustSayingMessageMapper.cs
slang25: _Shakes fist at agent_
image

slang25: I mean, I didn't check it, so you could say that's on me.
But I wont hear it!

iancooper: At least the agent reads the docs :)

iancooper (approved):
Looks good now!

iancooper:
@slang25 getting a build error on the nestandard20 TFM

"Error: /home/runner/work/Brighter/Brighter/src/Paramore.Brighter.MessagingGateway.AWSSQS.V4/SqsMessageCreator.cs(264,35): error CS8602: Dereference of a possibly null reference. [/home/runner/work/Brighter/Brighter/src/Paramore.Brighter.MessagingGateway.AWSSQS.V4/Paramore.Brighter.MessagingGateway.AWSSQS.V4.csproj::TargetFramework=netstandard2.0]"

PR #4132, opened 2026-05-13
fix(cloud-events): preserve message header values in CloudEventsTransformer

iancooper:
## Summary

- **Fix**: `CloudEventsTransformer` was overwriting `Source`, `Type`, `DataSchema`, `Subject`, and `DataContentType` values already set by upstream message mappers (e.g. `JustSayingMessageMapper`). When neither the `[CloudEvents]` attribute nor the `Publication` provided explicit values, properties were nulled out or reset to defaults.
- **Impact**: This caused the SNS `PublishRequest.Subject` to be null, so AWS omitted the `"Subject"` field from the SNS notification envelope — breaking downstream consumers (e.g. Python Lambda) expecting `record.json_body["Subject"]`.
- **Fix approach**: Changed precedence to: attribute params > message header (if explicitly set) > publication value, matching the existing `SpecVersion` pattern on the same method.

### 1. Precedence flip (attribute > mapper > publication)

This changes the precedence of CloudEvent header resolution from `attribute > publication` to `attribute > message header (mapper) > publication`. Previously, `Publication.Source/Type/Subject/DataSchema` would always overwrite values that a custom message mapper had set. After this change, **mapper-set values take priority over Publication defaults**, and Publication values serve as a fallback.

**Who is affected**: Users with custom message mappers that set CloudEvent header values AND also set explicit (non-default) values on `Publication` for the same properties, expecting Publication to override the mapper. This is unlikely in practice since the mapper and transformer receive the same Publication object.

### 2. `CloudEventsAttribute.DataContentType` default removed

Previously `CloudEventsAttribute.DataContentType` defaulted to `"application/json"`, meaning `[CloudEvents(0)]` always stamped `application/json` as the content type — silently overriding any mapper-set content type (e.g. `text/xml`, `application/protobuf`). It is now nullable with no default, so the attribute only sets `datacontenttype` when explicitly provided. The mapper's content type is preserved by default.

**Wire format note**: Messages that previously had bare `application/json` as `datacontenttype` may now show `application/json; charset=utf-8` (the `MessageHeader.ContentType` default). Consumers using strict `==` comparison against `application/json` should update to a prefix/contains check.

### 3. `InitializeWrapFromAttributeParams` signature change

A new positional slot was inserted at index 3 (`DataContentType`), shifting `DataSchema` (→4), `Subject` (→5), and `Format` (→6). Internal callers all go through `attribute.InitializerParams()` so in-tree wiring is consistent. Third-party code calling `InitializeWrapFromAttributeParams` directly with a hand-rolled `object?[]` will need to update their call sites.

### Production code
- `CloudEventsTransformer.cs` — Renamed `WritePublicationHeaders` → `ApplyCloudEventsPrecedence`; extracted `ResolveWithSentinel<T>` helper for sentinel-based precedence; added `_dataContentType` field with explicit `if`-guard assignment
- `CloudEventsAttribute.cs` — Made `DataContentType` nullable (`string?`) with no default
- Updated class-level XML docs on both classes to document precedence consistently for all 5 settable attributes

### Tests
- Precedence matrix tests: no-attribute/no-publication preserves mapper; attribute wins over mapper+publication; mapper wins over publication; publication used as fallback
- DataContentType tests: default attribute doesn't override mapper (via real `InitializerParams()` pipeline); explicit attribute overrides mapper
- Sentinel collision test: documents that mapper Source == DefaultSource → publication wins (known trade-off)
- JSON format test: verifies `datacontenttype` in CloudEvent envelope body reflects mapper's original content type
- Relative URI source test (CloudEvents spec compliance)
- V9→V10 Bag camelCase migration tests (PR #4132 context)
- Aligned test class names to match file names per project convention

## Test plan

- [x] All 39 CloudEvents tests pass on both net9.0 and net10.0
- [x] Precedence matrix fully covered (attribute > mapper > publication > defaults)
- [x] DataContentType nullable change tested via real attribute pipeline
- [x] Sentinel collision behavior documented and tested
- [x] JSON envelope body content type verified

### DIFF
--- a/src/Paramore.Brighter/Transforms/Transformers/CloudEventsTransformer.cs
+++ b/src/Paramore.Brighter/Transforms/Transformers/CloudEventsTransformer.cs
@@ -26,7 +26,7 @@ namespace Paramore.Brighter.Transforms.Transformers;
 /// REQUIRED
 ///     id => the message id from <see cref="MessageHeader"/>; you don't set this here, as we use the id from the <see cref="Request"/>
 ///     source => attribute > mapper > publication
-///     specversion => attribute > mapper; defaults to 1.0
+///     specversion => attribute > mapper; defaults to 1.0 (no publication fallback — <see cref="Publication"/> has no SpecVersion property)
 ///     type => attribute > mapper > publication
 /// OPTIONAL
 ///     datacontenttype => attribute > mapper (no publication fallback)
@@ -132,7 +132,7 @@ public partial class CloudEventsTransformer : IAmAMessageTransform, IAmAMessageT
     /// <inheritdoc />
     public Message Wrap(Message message, Publication publication)
     {
-        var msg =  WritePublicationHeaders(message,  publication);
+        var msg =  ApplyCloudEventsPrecedence(message,  publication);
         return _format == CloudEventFormat.Binary ? msg : WriteJsonMessage(msg, publication);
     }
 
@@ -219,7 +219,7 @@ public partial class CloudEventsTransformer : IAmAMessageTransform, IAmAMessageT
         }
     }
 
-    private Message WritePublicationHeaders(Message message, Publication publication)
+    private Message ApplyCloudEventsPrecedence(Message message, Publication publication)
     {
         // Precedence for all attributes: attribute params > message header (set by mapper) > publication (fallback)
         // Source and Type use sentinel checks (s_defaultSource / CloudEventsType.Empty) because their defaults
@@ -230,7 +230,8 @@ public partial class CloudEventsTransformer : IAmAMessageTransform, IAmAMessageT
         message.Header.Type = _type is not null
             ? new CloudEventsType(_type)
             : (message.Header.Type != CloudEventsType.Empty ? message.Header.Type : publication.Type);
-        message.Header.ContentType = _dataContentType ?? message.Header.ContentType;
+        if (_dataContentType is not null)
+            message.Header.ContentType = _dataContentType;
         message.Header.DataSchema = _dataSchema ?? message.Header.DataSchema ?? publication.DataSchema;
         message.Header.Subject = _subject ?? message.Header.Subject ?? publication.Subject;
         message.Header.SpecVersion = _specVersion ?? message.Header.SpecVersion;
--- a/src/Paramore.Brighter/Transforms/Transformers/CloudEventsTransformer.cs
+++ b/src/Paramore.Brighter/Transforms/Transformers/CloudEventsTransformer.cs
@@ -24,7 +24,7 @@ namespace Paramore.Brighter.Transforms.Transformers;
 ///     type => uses the type <see cref="MessageHeader"/>; as we used type based routing, we recommend using the hostname
 ///         scoped name of the request class you are sending
 /// OPTIONAL
-///      datacontenttype => sets the content type for <see cref="MessageBody"/> and <see cref="MessageHeader"/>
+///      datacontenttype => uses the content type from the <see cref="CloudEventsAttribute"/>, or preserves the existing <see cref="MessageHeader"/> value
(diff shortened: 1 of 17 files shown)


src/Paramore.Brighter/Transforms/Transformers/CloudEventsTransformer.cs
```diff
@@ -208,10 +208,13 @@ private static Message ReadCloudEventJsonMessage(Message message)
 
     private Message WritePublicationHeaders(Message message, Publication publication)
     {
-        message.Header.Source = _source ?? publication.Source;
-        message.Header.Type = _type is not null ? new CloudEventsType(_type) : publication.Type;
-        message.Header.DataSchema = _dataSchema ?? publication.DataSchema;
-        message.Header.Subject = _subject ?? publication.Subject;
+        message.Header.Source = _source
+            ?? (message.Header.Source.AbsoluteUri != MessageHeader.DefaultSource ? message.Header.Source : publication.Source);
```
lillo42:
```suggestion
            ?? (message.Header.Source.ToString() != MessageHeader.DefaultSource ? message.Header.Source : publication.Source);
```
`AbsoluteUri` can throw if the `Source` isn't absolute

lillo42 (changes requested):

lillo42 (approved):

slang25:
The AbsoluteUri vs DefaultSource bug Claude spotted does look like it needs fixing

iancooper:
## Investigation: V9 → V10 Subject migration & additional fixes

### Root cause analysis

We've confirmed the root cause of the SNS Subject issue has two contributing factors:

**1. CloudEventsTransformer overwriting mapper-set headers (fixed in this PR)**

The original bug: `WritePublicationHeaders` used precedence `attribute > publication`, so a `Publication` with no Subject set would overwrite a mapper-set `message.Header.Subject` with `null`. This is now fixed with the correct precedence: `attribute > message header > publication`.

**2. V9 → V10 Bag camelCase serialization (user-side migration needed)**

For users who set Subject via the V9 pattern `header.Bag.Add("Subject", value)`: in V10, `JsonSerialisationOptions` uses `PropertyNamingPolicy = JsonNamingPolicy.CamelCase`, and the custom `DictionaryStringObjectJsonConverter` explicitly applies this policy to dictionary keys during serialization ([source](https://github.com/BrighterCommand/Brighter/blob/master/src/Paramore.Brighter/Serialization/DictionaryStringObjectJsonConverter.cs#L57)). This means `"Subject"` in the Bag becomes `"subject"` in the serialized JSON — breaking downstream consumers (e.g. Python Lambdas) that look for PascalCase `"Subject"`.

**Migration path:** Replace `header.Bag.Add("Subject", Subject)` with `header.Subject = Subject` in message mappers. This feeds the native SNS `Subject` field via the `PublishRequest` constructor, which AWS serializes as PascalCase `"Subject"` in the SNS notification envelope — exactly what downstream consumers expect.

Tests proving this are in commit 36a86cfe8.

### Additional fix: InitializerParams index mismatch

Commit 05d230102 fixes a pre-existing bug where `CloudEventsAttribute.InitializerParams()` returns 7 elements `[Source, Type, SpecVersion, DataContentType, DataSchema, Subject, Format]` but `InitializeWrapFromAttributeParams` only expected 6 (skipping DataContentType). This caused indices 3–6 to be misaligned when wired through `TransformerFactory`:

- `[3]` DataContentType (e.g. `"application/json"`) was silently parsed as a URI for DataSchema
- `[4]` DataSchema URI string was assigned as Subject  
- `[5]` Subject string failed the `is CloudEventFormat` check, so Format stayed at default
- `[6]` Format enum was never read

The fix adds `_dataContentType` support to the transformer (matching the attribute > message header precedence of the other fields) and aligns all indices correctly.

slang25 (approved):

Issue #4144, opened 2026-05-19
Box provisioning: support per-tenant isolation of __BrighterMigrationHistory


iancooper:
## Context

`__BrighterMigrationHistory` (or `BrighterMigrationHistory` on Spanner) is currently created in the backend's default schema (`dbo` on MSSQL, `public` on PG, the connection-bound `DATABASE()` on MySQL), regardless of the configured `SchemaName`. Multi-tenant rows remain unambiguous because the PK is `(SchemaName, BoxTableName, MigrationVersion)`, but **physically a single history table is shared across all tenants** in a deployment.

For operators running a separate schema per tenant — typically for isolation, backup-restore boundaries, or per-tenant retention policies — there is no way today to keep the migration history table out of the shared schema.

## Suggested options (none committed yet)

- Honour `IAmARelationalDatabaseConfiguration.SchemaName` when creating `__BrighterMigrationHistory` (per-tenant history table; runner becomes schema-aware end-to-end).
- Or: expose a separate `BoxProvisioningOptions.MigrationHistoryScope` enum (`Global` (today's default) | `PerSchema`).

Either option needs a migration story for existing deployments that already have a `dbo.__BrighterMigrationHistory` populated.

## References

- Raised in code review of PR #4039 (reviewer item F2-5).
- Existing schema-routing fixes for fresh-install DDL: PR #4039 commits `df0e5a847`, `2c719254e`, `c9ee00311` (F1a/F1b/F1c).

iancooper:
Requirements documented in `specs/0029-multi-tenancy-migrations/requirements.md`.

Captures the per-tenant isolation need for `__BrighterMigrationHistory`, the end-to-end schema-awareness requirement, and a backward-compatible default. The two candidate mechanisms (honour `SchemaName` vs. a new `MigrationHistoryScope` opt-in) and the migration story for existing default-schema deployments are deferred to the ADR (will be `docs/adr/0060`). SQLite and Spanner are out of scope (no schema concept).

PR #4155, opened 2026-05-27
Multi-tenancy database migration 

iancooper:
## Summary

Specification and accepted **ADR 0060** for per-tenant placement of the box migration-history table — issue #4144 (PR #4039 review item F2-5). No production code yet; this PR lands the approved requirements and design for early feedback.

- Introduces an opt-in `MigrationHistoryScope { Global (default) | PerSchema }` on `BoxProvisioningOptions`.
- Under `PerSchema`, `__BrighterMigrationHistory` is created in the configured `SchemaName` on **MSSQL and PostgreSQL**, schema-aware end-to-end (CREATE/INSERT + detection).
- `Global` (default) preserves today's placement for **all** operators — including those who already set `SchemaName` — so upgrades move nothing and re-run nothing.
- One-time, idempotent seed copies prior applied-migration rows on the first `Global → PerSchema` flip so no migration re-runs (FR5).
- **Out of scope**: MySQL (`PerSchema` no-op — a MySQL schema is a database), SQLite/Spanner (no schema concept), reverse flip.

## Process

- Requirements: PASS after 4 adversarial review rounds (`specs/0029-multi-tenancy-migrations/review-requirements.md`).
- Design (ADR 0060): Accepted, PASS after 2 adversarial review rounds (`specs/0029-multi-tenancy-migrations/review-design.md`).
- Builds on the role-interface + template-method box-provisioning architecture from ADRs 0057–0059.

## Next

`/spec:tasks` → TDD implementation (tidy-first: structural plumbing, then D3 guard → D4 wiring → D5 seed → D6 logging), integration-tested on real containers.


iancooper:
Thanks for the careful read — agreed on both Medium concerns and addressed in [`3214b95a4`](https://github.com/BrighterCommand/Brighter/pull/4155/commits/3214b95a4). Walk-through of every numbered concern below.

### Medium — addressed

**1. Narrow the seed catch to the actual permission codes.** Both runners now filter via a `when` clause:

- `MsSqlBoxMigrationRunner.IsLegacyHistoryReadPermissionDenied(SqlException) => ex.Number is 229 or 230` — 229 is the canonical "permission denied on object", 230 is the column-level variant (DENY-ed on a specific column even when SELECT is allowed at the table level). 297/916 deliberately excluded as you suggested.
- `PostgreSqlBoxMigrationRunner.IsLegacyHistoryReadPermissionDenied(PostgresException) => ex.SqlState == "42501"` — `insufficient_privilege`, the only canonical code for missing SELECT/INSERT on a relation.

Unmatched `SqlException` / `PostgresException` (deadlock victim 1205 / 40P01, connection drop 64/10054 / 08xxx, statement timeout, login failure 18456, future-refactor syntax errors 42601, CHECK/FK violations) now propagate untouched. The outer UoW transaction still rolls back so partial per-schema state isn't left behind. Existing permission-denied integration tests (which set up real least-privilege CREATE LOGIN / CREATE ROLE credentials) still pass — confirming SQL Server actually raises 229 and PG actually raises 42501 in practice against the documented grant set.

**2. Symmetric wrap on the probe.** `DoesHistoryTableExistAsync` is now called inside a `try { … } catch (… ex) when (IsLegacyHistoryReadPermissionDenied(ex))` block in both runners, with the same operator-facing message factored into a shared private `BuildLegacyHistoryReadDeniedException` helper. As you noted, `sys.tables` / `information_schema.tables` are row-filtered (a missing grant returns zero rows rather than 229/42501), so this catch is mostly theoretical — but the wrap keeps the documented "any legacy-history-read failure → `ConfigurationException` with the documented message" contract honest against future backend behaviour changes or new probe shapes.

Regression-clean: MSSQL BoxProvisioning 85/85 net9 · PG BoxProvisioning 76/76 net9, including the two least-privilege integration tests that exercise the narrowed catch end-to-end.

### Low — acknowledged, deferred

**3. Steady-state PerSchema cost (extra round-trip).** Agreed and honestly documented at `MsSqlBoxMigrationRunner.cs:147-150`. The "seed completed" sentinel-row idea is a nice optimization — out of scope for this PR but I'll file it as a follow-up for operators with very high restart rates or hundreds of provisioned tables.

**4. PR description is stale.** Yes — will update before merge to describe what actually landed (full D3/D4/D5/D6 production implementation across MSSQL/PG/MySQL/SQLite/Spanner, the `IDetectionHelper` signature change, registrations updates, and ~50 integration tests including the permission-denied harness with real least-privilege CREATE LOGIN setup).

**5. MSSQL `OrdinalIgnoreCase` vs PG `Ordinal`-after-`Normalize` asymmetry.** The adjacent comments already explain the per-backend folding-rule justification, but adding a one-line shared invariant ("we compare the resolved schema to the backend default in the case-form the backend stores") makes the pair self-documenting at a glance. I'll add that the next time these helpers get touched rather than churning the file just for the nit.

**6. `BoxProvisioningOptions.Registrations` not thread-safe.** Correct — the type is `List<Action<IServiceCollection>>`. Configure-time mutation followed by build-time replay is the documented usage and isn't concurrent, but you're right that the pattern is now copy-pasted across four backend extensions, so the failure mode would also propagate if someone ever does call `Add(…)` concurrently with iteration. Adding a thread-safety note to the property doc is cheap and worth doing; converting to a synchronized collection is more invasive and not warranted by current usage. Will follow up.

### Spec/ADR notes
Appreciated re: ADR 0060 wording, the Errata block, the `MigrationHistoryScope` operator-facing remark, and explicitly scoping out the `PerSchema → Global` reverse flip. Those framings were all carved out under spec review and the validation is helpful to have on the implementation PR record.

PR #4165, opened 2026-06-01
[Feat] Message Mapping Errors should use the Reject path and an invalid message queue if available

### LINKED ISSUE #4160: Message pump silently Acks on MessageMappingException instead of using RejectMessage flow
**Describe the bug**
When the message pump catches `MessageMappingException` (or any unhandled `Exception`) during message translation/dispatch, it logs the failure, increments the unacceptable-message counter, and then falls through to the unconditional `Acknowledge(message)` at the bottom of the loop. On AWS SQS V4 (and other transports with the same Ack semantics) this performs a straight `DeleteMessage` — the Invalid Message Queue and DLQ are never consulted, even when they are configured on the consumer.

Every other meaningful exception path in the pump already routes through `RejectMessage(...)`:

| Exception                | Action                                                              |
|--------------------------|---------------------------------------------------------------------|
| `ConfigurationException` | `RejectMessage(DeliveryError)` → dispose channel, stop pump         |
| `RejectMessageAction`    | `RejectMessage(DeliveryError)` → `continue`                         |
| `InvalidMessageAction`   | `RejectMessage(Unacceptable)` → `continue`                          |
| `DeferMessageAction`     | `RequeueMessage(delay)` → `continue`                                |
| `DontAckAction`          | `Channel.NackAsync` → `continue` (no delete)                        |
| **`MessageMappingException`** | **logs only → falls through to `Acknowledge` (silent delete)** |
| **`Exception` (catch-all)**   | **logs only → falls through to `Acknowledge` (silent delete)** |

`SqsMessageConsumer.RejectAsync` (V4) already implements the desired routing: `Unacceptable` → Invalid Message Queue, falling back to DLQ; `DeliveryError`/`None` → DLQ; in both cases the original is then deleted. So the plumbing exists — the message pump just isn't using it for these two exception types.

**To reproduce**
This was originally discovered via #4159 (a missing transformer registration causing `MessageMappingException` from the unwrap pipeline). With an SQS V4 consumer configured with an Invalid Message Queue and a DLQ, messages that fail mapping are deleted from the source queue and never appear on either of the rejection channels.

**Expected behaviour**
On `MessageMappingException`, route the message through `RejectMessage` with `RejectionReason.Unacceptable` so the consumer can move it to the Invalid Message Queue (or fall back to the DLQ), then delete the source. The same treatment for the bottom `catch (Exception)` is also worth considering — today it has the same silent-delete behaviour.

**Affected files**
- `src/Paramore.Brighter.ServiceActivator/Proactor.cs:374-385` (`catch (MessageMappingException ...)` and `catch (Exception ...)` blocks)
- `src/Paramore.Brighter.ServiceActivator/Reactor.cs:339-350` (sync equivalents)

The existing `IncrementUnacceptableMessageCount()` call should stay so the `UnacceptableMessageLimit` / `UnacceptableMessageLimitWindow` guardrail still terminates the pump on repeated mapping failures. Note that the default `UnacceptableMessageLimit` is `0` (no limit), so today messages that fail mapping are silently deleted forever unless an operator has explicitly set the limit.

No changes needed in `SqsMessageConsumer` — its `RejectAsync` already implements the correct routing semantics.

**Related**
Discovered while investigating #4159. That issue covers the root cause (assembly load timing in auto-registration); this issue covers the downstream symptom (failed-mapping messages being silently deleted instead of routed to IMQ/DLQ).

iancooper:
## Description

MessageMappingException should trigger RejectMessage instead of falling through to acknowledge, which results in the message being deleted rather than re-routed. We want malformed messages to be routed to a configured IMQ/DLQ rather than being silently deleted. 

Note that the scope is MessageMappingException alone: the catch-all Exception (dispatch) path is deliberately left unchanged per Brighter's handler-failure policy [see Error Handling](https://brightercommand.gitbook.io/paramore-brighter-documentation/using-an-external-bus/handlerfailure).

## Related Issues

Closes #4160.

## Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

## Checklist

- [X] I have read the [Contributing Guide](../blob/master/CONTRIBUTING.md)
- [X] I have checked the [documentation](https://brightercommand.gitbook.io/paramore-brighter-documentation/) for relevant guidance
- [] I have added/updated XML documentation for any public API changes
- [] I have added/updated tests as appropriate
- [] My changes follow the existing code style and conventions


iancooper:
## Review follow-up

Addressed the two actionable test gaps from the code review:

**Fix 1 — catch-all tests now exercise `catch (Exception e)`, not `catch (AggregateException)`**

Both `When_a_dispatch_exception_is_thrown_the_catch_all_acknowledges(_async).cs` previously used `SpyExceptionCommandProcessor` with `MyEvent`, which causes the pump to call `PublishAsync`/`Publish`. That throws `AggregateException`, caught by the earlier `catch (AggregateException)` block — never reaching the generic catch-all. Changed to `MyCommand` so the pump calls `SendAsync`/`Send` instead. `SpyExceptionCommandProcessor.Send[Async]` throws bare `new Exception()`, which falls through to `catch (Exception e)` at `Proactor.cs:383` / `Reactor.cs:348`. The assertion (`Assert.Empty(invalidMessageKey)`) was already correct; now the mechanism genuinely guards the block it names.

**Fix 2 — no-IMQ test files renamed to reflect actual setup**

Both `When_a_message_fails_to_be_mapped_with_no_imq_the_pump_still_delegates_reject(_async).cs` renamed to `When_no_imq_configured_reject_falls_back_to_dlq(_async).cs`. Class names updated to `MessagePumpMappingFailureNoImqFallsToDlq(Async)Tests`; fact names updated to `When_No_Imq_Configured_Reject_Falls_Back_To_Dlq`. The test setup (IMQ absent, DLQ present) and assertions are unchanged — only the naming now matches.

95/95 MessageDispatch tests green on net9.0 and net10.0.

---

**Remaining review observations — noted, no action taken:**

- **`Thread.CurrentThread.ManagedThreadId` vs `Environment.CurrentManagedThreadId` inconsistency** — explicitly scoped out by ADR 0061 §C-1 ("do NOT normalise"). The shared `description` local uses the established pattern in that block verbatim. Noted as a follow-up if anyone normalises the surrounding code.

- **`RejectAsync`-throws risk** — documented in ADR 0061 as deliberate parity with the adjacent `InvalidMessageAction` and `RejectMessageAction` blocks. All three blocks share the same exposure: an exception from `RejectAsync` escapes the catch and propagates out of `EventLoop()`. Extending the guard here would be scope creep relative to the stated fix. A separate issue can be raised if this warrants a broader guard across all three reject paths.

PR #4166, opened 2026-06-05
Fix consumer OTel trace isolation

thomhurst:
Fixes consumer spans for messages that arrive without a propagated `TraceParent`.

Previously those spans inherited the long-lived pump activity, so unrelated consumed messages could share one trace. This starts those consumer spans as roots while preserving propagated message context when it exists.

Validated with the new regression test and the core observability test filter.


iancooper:
Let me know when you need a review?

src/Paramore.Brighter/Observability/BrighterTracer.cs
```diff
@@ -958,6 +965,57 @@ private static void Propogate(Activity? parentActivity, Activity? activity)
...
+                kind: kind,
+                parentId: parentId,
+                tags: tags,
+                startTime: startTime);
+        }
+
+        return StartRootActivity(spanName, kind, tags, startTime);
+    }
+
+    private Activity? StartRootActivity(
```
iancooper, line 992: Should we link a Root Activity? In the context of the message pump and the broker consumer, we want to associate the two without making one a child, so we can measure duration separately. I think this would imply a link should be used.

thomhurst:
Confirming the 🔴 "producer-parented spans don't restore `Activity.Current`" concern is a **false positive** — the suite is green, and the test that targets exactly this scenario passes.

`When_propagated_messages_are_consumed_should_keep_linking_the_message_pump_span` (added in e74e0e2) passes on both target frameworks:

```
Passed!  - Failed: 0, Passed: 1, Skipped: 0, Total: 1 - Paramore.Brighter.Core.Tests.dll (net9.0)
Passed!  - Failed: 0, Passed: 1, Skipped: 0, Total: 1 - Paramore.Brighter.Core.Tests.dll (net10.0)
```

The reasoning gap is the model of `Activity.Stop()`. It does **not** restore `Activity.Current` to `span.Parent`. It restores `Activity.Current` to whatever was current **when the span started**. The producer-parented branch never suppresses `Activity.Current` (only `StartRootActivity` does, via `Activity.Current = null`), so the pump span is the captured restore target and `Activity.Current` returns to the pump when the span ends — no `PreviousActivity` property is needed on that path.

Probing the exact two-propagated-message loop (`Begin` pump span, then receive + process per message, both messages carrying a `TraceParent`):

```
m1.traceparent / m2.traceparent : both set (producer-parented branch taken)
afterProcess1 == pump?           : True   (not null)
receive2.linksPump?              : True
process2.linksPump?              : True
```

So pump correlation is retained on every iteration, not just the first. The `PreviousActivity` restore mechanism is required only by `StartRootActivity`, precisely because that is the one path that nulls `Activity.Current` during creation.

(The minor note about test source registration is correct but pre-existing; the two-source registration is required here so the synthetic producer activity samples and writes a real `TraceParent`.)

iancooper (approved):
Thanks @thomhurst

PR #4180, opened 2026-06-10
Spec 0034: failed-delivery observability for confirmation-based producers (#4179)

### LINKED ISSUE #4179: Provide more context for failed delivery using a callback
Some of our gateways notify us asynchronously as to the success/failure of publication of a message (RMQ/Kafka). Eventually, this routes to a method on `OutboxProducerMediator`

```
       private bool ConfigurePublisherCallbackMaybe(IAmAMessageProducerSync producer, RequestContext requestContext)
        {
            if (producer is ISupportPublishConfirmation producerSync)
            {
                producerSync.OnMessagePublished += delegate(bool success, string id)
                {
                    if (success)
                    {
                        Log.SentMessage(s_logger, id);

                        if (_outBox != null)
                            ExecuteWithResiliencePipeline(
                                () => _outBox.MarkDispatched(id, requestContext, _timeProvider.GetUtcNow()),
                                requestContext);
                    }
                };
                return true;
            }

            return false;
        }
```

This method logs that a message was sent, and then tries to mark it as dispatched in the outbox.

However, this method fails to handle the error circumstance:

- The publish failed, and we are being called with success==false. 

Note: ExecuteWithResiliencePipeline will log if the message cannot be published:

```
           catch (Exception ex)
            {
                Log.ExceptionWhilstTryingToPublishMessage(s_logger, ex);
                CheckOutstandingMessages(requestContext);
                return false;
            }
```

In the event that we cannot publish a message, we should log that we were unable to publish. In principle, since the message is not marked as dispatched, we will retry. However, there is no information provided to the user that the publish has failed, even if there are issues with the broker. We should treat this as a warning when logging and not an error.

iancooper:
## Summary

Implements **Spec 0034** — making failed message delivery from confirmation-based producers (Kafka, RabbitMQ) observable, addressing issue #4179. This PR carries the **full workflow: requirements → design (ADR 0063) → tasks → implementation** (the production code is included; see "Production changes" below).

Today, when `OnMessagePublished(success: false, ...)` fires, the failure is silently dropped: nothing logged, no telemetry, the circuit breaker is never tripped for confirmation-based producers, and Kafka discards the failed message id and the `ProduceException` reason/code. This PR closes those gaps **without** changing the deliberate "don't bubble, let the Sweeper retry" design.

## Production changes (~511 LOC across 11 `src/` files)

- **`ISupportPublishConfirmation` / `PublishConfirmationResult`** — `OnMessagePublished` is enriched to `Action<PublishConfirmationResult>` (success, message id, wire topic, publish `ActivityContext`). The only interface contract that changes, affecting just the two `ISupportPublishConfirmation` producers.
- **`OutboxProducerMediator`** — the confirmation callback emits a standalone settle span first (linked to the publish span), logs a Warning on failure, and trips the circuit breaker on the per-message **wire topic** (parity with the non-confirmation `!sent` path). Observability is isolated in `try/catch` (NFR-4); the span is disposed in `finally` (NFR-2).
- **`BrighterTracer.CreateConfirmationSpan`** (+ `IAmABrighterTracer`) — creates the short-lived settle span carrying an `ActivityLink` to the original publish span; degrades to no-link when the context is absent.
- **Kafka** (`KafkaMessageProducer` / `KafkaMessagePublisher`) — logs the previously-swallowed `ProduceException` (reason + code) at Warning and propagates the failed message id through `PublishResults`.
- **RabbitMQ** (`RmqMessageProducer` async + sync) — confirmation carries the id, wire topic and publish link on ack/nack.
- **`InMemoryMessageProducer`** — opt-in async confirmation pump (used in tests and early/local development).
- **`InMemoryOutboxCircuitBreaker`** — `ConcurrentDictionary` + `TryRemove` to address the documented concurrency race.

## Requirements highlights

- **FR-1** Warning-level log (not Error — message stays in Outbox for the Sweeper) with message id + wire topic.
- **FR-2** Standalone short-lived OTel `Activity` carrying an `ActivityLink` to the original publish span (graceful degradation to no-Link when the context isn't reachable).
- **FR-3** Trip the circuit breaker for the per-message **wire topic** (`message.Header.Topic`) — exact parity with the non-confirmation `!sent` path, including reply/rewritten topics.
- **FR-6/FR-8** Kafka: log the currently-swallowed `ProduceException` (reason + code); propagate the failed message id through `PublishResults`.
- **FR-9** RabbitMQ already carries the id on nack — verify-only for the id path.
- **FR-10 (superseded)** Originally proposed an optional `RequestContext?` on the producer `Send` methods; ADR 0063 supersedes this — see below.

## ADR 0063 — the design decision

The open C-7/C-8 question (how the wire topic + original trace context reach the `OnMessagePublished` callback) is resolved as **mechanism (a): extend the contract**, refined:

- **Enrich `OnMessagePublished`** to `Action<PublishConfirmationResult>` (success, message id, wire topic, publish `ActivityContext`) — the only interface contract that changes, affecting just the two `ISupportPublishConfirmation` producers (Kafka, RMQ).
- **No `Send`-signature change.** The producer self-sources the publish span from `Activity.Current` (set to S1 by `CreateProducerSpan`, `AsyncLocal` so race-free), captured synchronously inside `Send`/`SendAsync`. This meets FR-10's intent with a strictly smaller blast radius (the ~10 non-confirmation gateways are untouched) — so **FR-10/AC-14 are marked superseded** in `requirements.md`.
- Rejected alternatives (b) correlation map and (c) header propagation are documented with rationale.

## Review trajectory

- **Requirements:** 3 adversarial rounds, 7 → 4 → 2 findings ≥ threshold, all resolved before approval.
- **Design (ADR 0063):** 4 adversarial rounds — Critical (AC-2 capture linchpin) → High (concurrency race on shared `producer.Span`, fixed by sourcing `Activity.Current`) → PASS → PASS. Findings in `review-design.md`.

Closes #4179.


iancooper:
Thanks for the thorough review — all four points are now addressed on this branch.

## #2 — confirmation span "link, not nest" (the correctness gap)

Confirmed and fixed in `f4db0b3c7`, with a RED-first test (`When_creating_a_confirmation_span_under_a_live_ambient_span_should_be_a_root`).

One correction worth recording: **the suggested fix — `StartActivity(..., parentContext: default(ActivityContext))` — does not work.** I verified empirically with a live `ActivityListener` + a non-null `Activity.Current`:

| call | resulting parent |
|---|---|
| `StartActivity(parentId: null)` (original code) | nests under ambient ❌ |
| `StartActivity(parentContext: default(ActivityContext))` (suggested fix) | **still nests under ambient** ❌ |
| clear `Activity.Current`, then `StartActivity(parentId: null)` | true root ✅ |

`default(ActivityContext)` trips .NET's `useCurrentActivityContext = (context == default && Activity.Current != null)` branch in `ActivitySource.CreateActivity`, so it falls back to the ambient activity just like the string overload. The only reliable way to force a root is to clear `Activity.Current` for the duration of the call — which the method then overwrites with the confirmation span anyway (so the success-branch `MarkDispatched` span still nests under S2). Fix + comment do exactly that.

Also slightly worse than described: on the InMemory async path the worker is started by the *first* enqueue, so without the fix every confirmation nested under whichever publish span happened to start the pump — not even its own message's span.

## #3 — Kafka persisted-but-id-less treated as a failure

Fixed in `87ab7c90f`. `PersistenceStatus.Persisted` now always raises **success** (id degrades to `Id.Empty` on the no-id case) rather than falling through to the failure path, so a persisted report can never trip the breaker. As you noted it's effectively unreachable (Brighter always stamps `MESSAGE_ID`), so no behavioural test was added — the existing tests are all live-broker integration tests and the state can't be produced through the broker contract.

## #4 — InMemory worker volatility / drain robustness

Fixed in `9c2542e09`. `_worker` is now `volatile` (so the CAS winner's publish is visible cross-thread), and `DisposeAsync` keys off `_pumpStarted` rather than a snapshot of `_worker`, spinning briefly for worker visibility — closing the CAS-vs-assign window so a dispose racing the very first enqueue still drains. Treated as production-robust since `InMemoryMessageProducer` is used beyond tests (early/local development), not just as a test double. The sync-over-async `Dispose()` is left as-is (acceptable for the in-memory case).

## #1 — PR description scope

Reconciled — the description now states the full requirements → design → tasks → implementation scope and lists the production changes, rather than "no production code yet".

iancooper:
Thanks for the thorough review — addressed in 2233fa536.

**#3 — bulk-confirming guard (the gating item).** Added `tests/Paramore.Brighter.Core.Tests/Confirmation/When_bulk_dispatching_a_confirming_producer.cs`. `InMemoryMessageProducer` is both `IAmABulkMessageProducerAsync` and `ISupportPublishConfirmation`, so it drives the real `OutboxProducerMediator.BulkDispatchAsync` path with no broker. Two facts, both deterministic via the `DisposeAsync` drain + `InMemoryOutbox` state:
- **Success:** bulk-clearing a confirming producer dispatches via the confirmation callback. Before the guard change this threw `InvalidOperationException("No async bulk message producer defined.")`, leaving the batch outstanding — so this locks the change in.
- **Failure (the inline-skip safety property):** with a nacking confirmation the message is **not** dispatched (despite `sent == true`, the inline `MarkDispatched` is correctly skipped) and the breaker trips on the wire topic. The trip also proves the bulk branch was entered — the pre-change throw would have sent and tripped nothing.

**#1 — InMemory success path drops the topic.** Fixed — success confirmations now carry `message.Header.Topic` on both the sync (`Enqueue`) and async (`DrainAsync`) paths, mirroring Kafka/RMQ which carry the wire topic on ack and nack.

**#2 — `Dispose()` no longer a no-op.** Rather than just document it, `Dispose()` now short-circuits when `_pumpStarted == 0`, keeping the default sync-confirmation path a true no-op instead of spinning up a `BrighterAsyncContext` on every dispose.

**#4 / #5** are notes only and already covered by the existing XML remark on `CreateConfirmationSpan` and the `DrainAsync` flush comment — no change.

Full core suite green: 775 passed, 0 failed, 7 skipped (net9.0).

Issue #4175, opened 2026-06-09
Add a Validation Handler to Brighter


iancooper:
**Is your feature request related to a problem? Please describe.**

Provide a validation handler and an attribute that validates command inputs before the handler executes.

Commands typically need validation; bad input can cause exceptions, SQL injection risks, or unnecessary database load. MediatR has a well-established pattern of validation via pipeline behaviors and FluentValidation.

## Proposal

- Create a `[ValidateQuery]` attribute and corresponding `ValidateRequestHandler`.
- The decorator resolves `IValidator<TRequest>` from the DI container and validates before calling `next`
- If validation fails, throw a `RequestValidationException` with structured error details
- Consider Supporting Brighter's Specification pattern as a validation framework
- Consider supporting FluentValidation as validation framework (separate package: `Paramore.Brighter.Validation.FluentValidation`)
- Consider supporting `System.ComponentModel.DataAnnotations` as a lightweight alternative

If you are less familiar with Brighter, FluentValidation might be a good starting PR

iancooper:
Note that this requirement also exists for Darker V5

xbizzybone:
I've opened a PR implementing this for **FluentValidation** as the starting point you suggested. It adds a provider-agnostic `Paramore.Brighter.Validation` package (one `[ValidateQuery]` / `[ValidateQueryAsync]` attribute, abstract `ValidateRequestHandler` base handlers, and `RequestValidationException` with structured per-field errors) plus `Paramore.Brighter.Validation.FluentValidation` that plugs FluentValidation in via `UseFluentValidation()`. The single attribute is designed so DataAnnotations and the Specification pattern drop in later as additive packages — no change to what's shipped here. The core is untouched; an ADR (0063) is included.

I used "Contributes to #4175" rather than "Closes" so you can keep this open to track the other two frameworks. A couple of design questions in the PR (attribute naming `[ValidateQuery]` vs `[ValidateRequest]`, and whether the abstractions should live in core vs their own package), and I flagged the Darker V5 need as a follow-up.

PR: #4183

PR #4183, opened 2026-06-15
feat: add request validation handler with FluentValidation provider (#4175)

xbizzybone:
## What

Adds request validation to Brighter as an optional, additive concern, with FluentValidation as the first provider — the starting point suggested in the issue.

Contributes to #4175

## Design

Two assemblies so a **single `[ValidateQuery]` attribute serves every framework**, while users only take the dependency they need:

**`Paramore.Brighter.Validation`** (provider-agnostic, depends only on the core)
- `[ValidateQuery]` / `[ValidateQueryAsync]` — extend `RequestHandlerAttribute`, mirroring `RequestLoggingAttribute`; they target the abstract base handlers.
- `ValidateRequestHandler<TRequest>` / `ValidateRequestHandlerAsync<TRequest>` — **abstract** base handlers that own the shared behaviour: null-guard the request, ask the provider for failures, throw before calling `next`, otherwise call `base.Handle`/`base.HandleAsync`.
- `RequestValidationException` carrying a framework-agnostic `IReadOnlyCollection<ValidationError>` (property, message, attempted value, error code), so a caller catches **one** exception type regardless of provider.

**`Paramore.Brighter.Validation.FluentValidation`** (this PR's provider)
- `FluentValidationRequestHandler<TRequest>` / `...Async` derive from the base handlers and resolve a FluentValidation `IValidator<TRequest>` from the container, mapping its failures onto `ValidationError`.
- `UseFluentValidation()` maps the abstract handlers to these implementations. **The core is not modified.**

Because `[ValidateQuery]` targets the abstract handler and the provider package maps it to a concrete one, adding `Paramore.Brighter.Validation.DataAnnotations` or `...Specification` later is purely additive — a new concrete handler plus a `UseX()` registration, with no change to the abstractions or to this package.

A request marked `[ValidateQuery]` with no registered validator throws `ConfigurationException` (Brighter's existing misconfiguration type), fail-fast, rather than silently skipping validation. An ADR (0063) is included as the first commit, per CONTRIBUTING.

## Scope / additive guarantees

- Two new assemblies + a new test project. The two edits to existing files are unavoidable wiring with no logic change:
  - `Brighter.slnx` — register the new projects.
  - `Directory.Packages.props` — central version for `FluentValidation` (11.12.0, the last 11.x line that still supports `netstandard2.0`, which Brighter targets).
- Targets `netstandard2.0;net8.0;net9.0;net10.0`; strong-named with `Brighter.snk`.

## Tests

xUnit, one behavior per file. Adversarial coverage through the FluentValidation provider: valid request (sync + async), single and multiple validation errors, missing validator (`ConfigurationException`), a validator that throws (propagated, not swallowed), null request (`ArgumentNullException`), cancellation-token propagation in the async handler, the exact shape of `RequestValidationException`, and concurrent reuse of a single handler instance. All green on net10.0 locally; CI covers every TFM.

## Open questions for the maintainer

1. **Naming.** The issue names the attribute `[ValidateQuery]`, but the concern validates commands as well as queries. Keep it, or rename to `[ValidateRequest]` (optionally keeping `[ValidateQuery]` as an alias)? Implemented as specified.
2. **Placement.** The abstractions live in a new `Paramore.Brighter.Validation` package; they depend only on the core, so they could equally live in the core assembly. Which do you prefer?
3. **Issue.** This delivers FluentValidation only; I used "Contributes to" rather than "Closes" so you can keep #4175 open to track DataAnnotations / Specification.
4. **Darker.** The same requirement exists for Darker V5; out of scope here, happy to follow up.

## Contributor License Agreement

Signed via CLA assistant — the `license/cla` check is green.

Minor docs note: CONTRIBUTING.md still links to clahub.com for the CLA, but ClaHub has shut down; the CLA assistant GitHub App is what actually runs on PRs. Might be worth updating that link in a follow-up.

## Type of Change

## Checklist

## Additional Notes

User-facing documentation for the BrighterCommand docs site can follow as a separate change once the API here is confirmed.


iancooper:
Thanks for this contribution! It helps a lot. It looks good and we should try to get it out asap.


1. Let's fold the abstractions into Paramore. Brighter, within a sub-folder. They depend only on the core and carry no third-party reference. Since RequestLoggingAttribute and friends already live in core, this follows the same pattern for built-ins. We should still ship providers separately. 
2. Let's name these [ValidateRequest] and not [ValidateQuery]; they query name is a  holdover from writing up the issue for Darker. Brighter has requests, not queries.
3. <Authors>Miguel Ramirez</Authors> in both src csproj files overrides the repo-wide Authors=Ian Cooper / Company=Brighter Command from src/Directory.Build.props. I'm not trying to steal your work, it just helps us attribute official packages.  Your attribution belongs in the file header's copyright notice, not in the package author field. You can also add yourself to the contributors file at the root (saves me the job)
4. There is no friendly error from ValidatePipelines() when UseFluentValidation() is omitted, but we have attributed handlers. We can put that in a separate issue/PR, though
5. UseFluentValidation() itself is never exercised.

iancooper:
Let's leave Darker for now, as it has a lot of structural changes for V5, but I can tag you there when it is ready to go. You are welcome to push implementations of DataAnnotations and Specifications here, too.

xbizzybone:
Thanks for the thorough review, @iancooper — really appreciate it. I've pushed the changes addressing all your points:

1. **Folded into the core.** The abstractions now live in `Paramore.Brighter` under `RequestValidation/` (with `Attributes/` + `Handlers/` sub-namespaces, mirroring `Logging`/`Inbox`); the standalone `Paramore.Brighter.Validation` package is gone and the providers stay separate. One thing to flag: I renamed the error type to `RequestValidationError` to avoid clashing with the existing `Paramore.Brighter.ValidationError` from the startup pipeline-validation (ADR 0053). I also used file-scoped namespaces to match the current core style.
2. **`[ValidateRequest]`** — renamed (and `[ValidateRequestAsync]`); the "query" name is gone.
3. **Authors** — dropped the `<Authors>` override so the repo-wide `Authors`/`Company` from `src/Directory.Build.props` apply, and added myself to `CONTRIBUTORS.md`.
4. **Friendly `ValidatePipelines()` error** — left as you suggested; happy to follow up in a separate issue/PR.
5. **`UseFluentValidation()` coverage** — added a test that drives the real `AddBrighter().UseFluentValidation()` through the pipeline (previously the handlers were only wired by hand in the test harness).

I also took you up on the offer to add the other two providers in this PR:
- **`Paramore.Brighter.Validation.DataAnnotations`** — validates the request's own `System.ComponentModel.DataAnnotations` attributes (no per-request validator, no extra dependency — the types ship with the framework).
- **`Paramore.Brighter.Validation.Specification`** — built on your Specification pattern (ADR 0040): resolves an `ISpecification<TRequest>` and collects failures via the public `ValidationResultCollector`. I documented that the specification should be registered with a per-request lifetime, since `Specification<T>` carries per-evaluation state.

The registration methods are `UseFluentValidation()` / `UseDataAnnotations()` / `UseSpecification()`. Each provider has the same adversarial test coverage plus sync and async `UseX()` wiring tests. **Darker** — left out for now, tag me when V5 is ready.

tests/Paramore.Brighter.Validation.FluentValidation.Tests/TestDoubles/ValidationPipeline.cs
```diff
@@ -0,0 +1,99 @@
...
+namespace Paramore.Brighter.Validation.FluentValidation.Tests.TestDoubles
+{
+    /// <summary>
+    /// Test Data Builder that wires a real Brighter pipeline for <see cref="GreetingCommand"/> with the
+    /// FluentValidation handlers attached, so a test can focus on the evident data (the validator and the
+    /// request) rather than the container plumbing.
+    /// </summary>
+    internal sealed class ValidationPipeline
+    {
+        private ValidationPipeline(CommandProcessor commandProcessor, HandlerReceipt receipt)
```
iancooper: Can we rename this? I found it confusing on first review, as it's a CommandProcessorBuilder, not the pipeline doing validation, which Brighter itself provides.

tests/Paramore.Brighter.Validation.FluentValidation.Tests/TestDoubles/CommandProcessorHarness.cs
```diff
@@ -0,0 +1,99 @@
...
+    {
+        private ValidationPipeline(CommandProcessor commandProcessor, HandlerReceipt receipt)
+        {
+            CommandProcessor = commandProcessor;
+            Receipt = receipt;
+        }
+
+        public CommandProcessor CommandProcessor { get; }
+
+        public HandlerReceipt Receipt { get; }
```
iancooper, line 49: I appreciate that we are creating a receipt here, as well as the CommandProcessor, so we might be able to come up with a name that reflects both those concerns

tests/Paramore.Brighter.Validation.Specification.Tests/When_a_valid_request_satisfies_its_specification_should_return_the_request.cs
```diff
@@ -0,0 +1,48 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Paramore.Brighter.Validation.Specification.Tests.TestDoubles;
+using Xunit;
+
+namespace Paramore.Brighter.Validation.Specification.Tests;
+
+public class ValidSpecificationValidationTests
+{
+    [Fact]
+    public void When_a_valid_request_satisfies_its_specification_should_return_the_request()
```
iancooper: Really, what you are testing here is that we don't throw an exception. The fact that the test returns the same request is just an artifact of the path having completed; it's not what you are trying to confirm. Assert.True() would work here as well, because if an exception were thrown, you would throw an exception from the test.

iancooper (commented):
This looks really good, and the comments here are mostly about creating clarity rather than providing functional fixes. As this is the first PR for this, it would be nice to get the baseline right.

It would be useful to have a simple sample for these. It can be something as simple as /Users/ian_hammond_cooper/CSharpProjects/github/xbizzybone/Brighter/samples/CommandProcessor but in a separate folder Validation and include some samples that support the different validation types that we have.

The samples will also help us create the docs

tests/Paramore.Brighter.Validation.FluentValidation.Tests/TestDoubles/ValidationPipeline.cs
xbizzybone: Done — renamed `ValidationPipeline` → `CommandProcessorHarness`. That drops the misleading "pipeline" (Brighter owns the validation pipeline) and names it for what it actually is: a builder whose `With(...)`/`WithAsync(...)` factories assemble a real `CommandProcessor`. Updated the XML doc and all call sites (the local `pipeline` is now `harness`).

tests/Paramore.Brighter.Validation.FluentValidation.Tests/TestDoubles/CommandProcessorHarness.cs
xbizzybone: Good point — it builds the `CommandProcessor` *and* exposes the `HandlerReceipt` the handler writes to. `CommandProcessorHarness` names that pairing, and the doc comment now spells out both concerns explicitly.

tests/Paramore.Brighter.Validation.Specification.Tests/When_a_valid_request_satisfies_its_specification_should_return_the_request.cs
xbizzybone: Agreed, the request-equality was incidental. Switched to `Assert.Null(Record.Exception(() => handler.Handle(validRequest)))` — the repo's existing "does not throw" idiom — so the test confirms exactly that. I applied the same fix to the four sibling happy-path tests (DataAnnotations + Specification, sync/async) that shared the `Assert.Same` pattern, and renamed them `..._should_return_the_request` → `..._should_not_throw` to match the intent.

xbizzybone:
Thanks @iancooper — addressed the review:

- **Renamed the `ValidationPipeline` test double → `CommandProcessorHarness`.** It's a builder for the `CommandProcessor` (Brighter owns the actual validation pipeline), and the name now reflects that it bundles the processor with the `HandlerReceipt`. Doc comment and call sites updated.
- **Valid-request tests now assert "does not throw"** via `Assert.Null(Record.Exception(...))` instead of returning the request; renamed the four affected happy-path tests to `..._should_not_throw`.
- **Added a sample** as suggested: `samples/Validation/` (modeled on `samples/CommandProcessor`) with three console apps — `FluentValidationSample`, `DataAnnotationsSample`, `SpecificationSample` — each sending one valid request (reaches the handler) and one invalid request (rejected with the collected `RequestValidationException.Errors`), plus a README on when to use each provider and the async variant. Verified each runs end-to-end.

iancooper:
Thanks for this contribution @xbizzybone. It is much appreciated. Feel free to ping us about other issues that you might want to pick up, and we will let you know. Typically, look for ones that are unassigned. Avoid the gateway for a couple of weeks, as we want to get the generic test framework in

PR #4185, opened 2026-06-15
Make Brighter logging instance-scoped and explicit

### LINKED ISSUE #4184: Static ApplicationLogging.LoggerFactory is shared across Brighter instances in a process
## Problem

Brighter routes all logging through a process-wide mutable static, `ApplicationLogging.LoggerFactory`. The dependency-injection extensions copy the container's `ILoggerFactory` into this static when building the `CommandProcessor` and `Dispatcher`.

Because it is a single static, it is **shared between two or more Brighter instances running in the same process** — a common situation in parallel test suites (each test class spins up its own `ServiceProvider`).

This leads to two failures:

1. **Wrong logger from the wrong service provider.** With two or more service providers, whichever Brighter initializes last overwrites the static. Code belonging to one instance then logs through the `ILoggerFactory` owned by a *different* service provider.

2. **Disposal breaks other running instances.** When a `ServiceProvider` is disposed it disposes its `ILoggerFactory`. The static still holds a reference to that now-disposed factory, so other still-running Brighter instances can throw `ObjectDisposedException` (or silently log nowhere) the next time they log.

The net effect is that two or more Brighters cannot run cleanly side by side in the same process.

## Expected

Logging should be instance-scoped: each Brighter instance logs through the `ILoggerFactory` of the service provider that created it, and disposing one instance must not affect another.

thomhurst:
Closes #4184

## Problem

Logging was routed through the process-wide static `ApplicationLogging.LoggerFactory`. Multiple Brighter instances could log through the wrong service provider and break each other when one provider was disposed.

Implicit `NullLogger` fallbacks also hid incomplete logging configuration.

## Change

Logging is now instance-scoped and explicit:

- Remove the `ApplicationLogging` compatibility shim.
- Require `ILoggerFactory`/`ILogger` dependencies throughout the runtime object graph instead of defaulting to `NullLogger`.
- Require `ConfigureLogging(ILoggerFactory)` before `CommandProcessorBuilder.Build()` and `DispatchBuilder.Build()`.
- Resolve `ILoggerFactory` with `GetRequiredService` in DI paths so missing logging configuration fails fast.
- Thread each owning container/builder's logger factory through command processors, dispatchers, message pumps, transports, stores, schedulers, provisioners, mediators, and pipelines.
- Update all repository callers. Tests and samples that intentionally suppress logging now opt into `NullLoggerFactory.Instance` explicitly.

## Behaviour change

Missing logger configuration now throws instead of silently disabling logs. Applications that intentionally want no logging must explicitly supply or register `NullLoggerFactory.Instance`.

## Tests

- Full solution build: 0 errors.
- `Paramore.Brighter.Core.Tests`: 978 passed, 7 skipped on both net9.0 and net10.0.
- `Paramore.Brighter.Extensions.Tests`: 142 passed on both target frameworks.
- `Paramore.Brighter.InMemory.Tests`: 153 passed on both target frameworks.
- `Paramore.Brighter.BoxProvisioning.Tests`: 111 passed on both target frameworks.


iancooper:
Alright, we will need to look over this one carefully, as it was previously a choice to avoid passing loggers throughout the application. It may not be a breaking change, but it may change behavior for V10 users who don't transition, potentially resulting in a silent failure. That would mean, IMO, it would be V11, and it would be preferable to force success rather than allow silent failure. In addition, it may change log messages.

Not necessarily adverse, but didn't really want to trigger V11 work until SEP/OCT

thomhurst:
Agreed it's a breaking change for anyone manually constructing these classes.  We can hold off until V11 if you were happy with this change and park for now if needed. 

For MEDI users I don't believe they'd have to tweak any code.

But also happy to discuss the design further. The current behaviour is a problem though for any two or more brighter's running under the same process. We noticed it earlier in a test suite.

What do you mean by force success btw? Remove the NullLogger fallback?

iancooper:
> What do you mean by force success btw? Remove the NullLogger fallback?

Yes, because I get that stops users having compilation issues, but it feels like it could silently fail on folks who where not expecting it.

iancooper:
>But also happy to discuss the design further. The current behaviour is a problem though for any two or more brighter's running under the same process. We noticed it earlier in a test suite.

I think it may help with some other aspects like testing too. We currently have to use a test capture context and I am not sure that is the best route

PR #4189, opened 2026-06-17
feat(validation): ValidatePipelines warns on unresolvable transforms and missing validation providers (#4159)

### LINKED ISSUE #4159: ValidatePipeline should check all required assemblies will be scanned
When assembly scanning takes place to look for handlers, it only scans loaded assemblies. However, assemblies may be loaded dynamically, meaning not all assemblies are loaded during scanning. This means that we may fail to register the handlers that we need.

Brighter has a solution to this problem: explicitly declare the assemblies you need to scan in the  load  from assemblies call. 

```
 services.AddBrighter(options =>
            {
                //we want to use scoped, so make sure everything understands that which needs to
                options.HandlerLifetime = ServiceLifetime.Scoped;
                options.MapperLifetime = ServiceLifetime.Singleton;
                options.PolicyRegistry = new GreetingsPolicy();
            })
            .AddProducers(configure =>
            {
                configure.ProducerRegistry = ConfigureTransport.MakeProducerRegistry<GreetingMade>(messagingTransport);
                configure.Outbox = makeOutbox.outbox;
                configure.TransactionProvider = makeOutbox.transactionProvider;
                configure.ConnectionProvider = makeOutbox.connectionProvider;
                configure.MaxOutStandingMessages = 5;
                configure.MaxOutStandingCheckInterval = TimeSpan.FromMilliseconds(500);
            })
            .AutoFromAssemblies([typeof(AddPersonHandlerAsync).Assembly])
            .ValidatePipelines()
            .DescribePipelines();
```

But it has no validation that you have added all the required assemblies. If you use a feature for Brighter, such as our ` JustSayingMessageMapper` there is no guarantee that it will be correctly registered, or its dependencies such as `JustSayingCompressionTransform`

We should extend `ValidatePipelines` to verify that the user has added all required assemblies to the `AutoFromAssemblies` call, ensuring pipelines are correctly registered.

Obviously, ValidatePipelines needs to be able to locate the pipelines from the Subscription and Publication information to force the assemblies to be loaded. It already runs the PipelineBuilder, and as such, we may have a contradiction: ValidatePipelines will force the loading of the very assemblies in question, i.e., it solves the problem because it runs. However, our expectation is that ValidatePipelines may not run in Production environments, which still makes its warning valuable.

There is work looking at removing reflection in favor of source generation, but that is out of scope here. However, we may need to adjust the warning if assembly scanning is replaced

xbizzybone:
## What

Extends `ValidatePipelines()` with two non-blocking (**Warning**) startup checks, closing #4159:

- **(A) Missing-transform check.** For each `Publication` (wrap) / `Subscription` (unwrap), if the resolved message mapper declares a transform whose transformer type (the attribute's `GetHandlerType()`) is not registered, emit a Warning naming the request type, transformer, entity (`Topic`/subscription `Name`), and prompting an `AutoFromAssemblies()` check — the strong signal that the transformer's assembly was not scanned (the issue's `JustSayingCompressionTransform` example).
- **(B) Missing validation-provider check** (your add-on in the issue comment). For each handler whose pipeline declares a request-validation step (a step targeting the open generic `ValidateRequestHandler<>` / `ValidateRequestHandlerAsync<>`, as `[ValidateRequest]`/`[ValidateRequestAsync]` do) while no provider is registered, emit a Warning naming the request/handler and `UseFluentValidation()` / `UseDataAnnotations()` / `UseSpecification()`.

Both checks are **Warning severity and never block startup**, even under `ValidatePipelines(throwOnError: true)` — matching your steer in the issue (ValidatePipelines force-loads the assemblies and is commonly off in production, so this is a dev-time heads-up).

## Scope (honest narrowing)

A discovery pass against the codebase narrowed (A) to **transforms** and added (B):

- **Mapper-missing is not reliably detectable** on the DI path: `MessageMapperRegistry` always falls back to the default `JsonMessageMapper<>`, so "no mapper" never occurs and a silent fall-back to the default is indistinguishable from intentionally using it. Excluded as too noisy (OOS-7 / C-9).
- **Subscription handler-missing is already detected** by the existing `ConsumerValidationRules.HandlerRegistered` (Error). Not re-implemented (OOS-8).
- The **default mapper** (`JsonMessageMapper`, which declares a built-in `[CloudEvents]` transform) is **out of scope** for (A) and skipped — its transforms are Brighter built-ins.

So (A) targets the reliably-detectable case — *the mapper is scanned but a transformer it declares is not* — which is exactly the `JustSayingMessageMapper` → `JustSayingCompressionTransform` example.

## Design

- New rules plug into the existing `ISpecification<T>` + `ValidationResultCollector` machinery evaluated by `PipelineValidator` — no new hosted service, no change to `throwOnError`.
- (A)-producer lives in core `ProducerValidationRules`; (A)-consumer in `Paramore.Brighter.ServiceActivator` `ConsumerValidationRules` (consumer concerns stay out of core); (B) in core `HandlerPipelineValidationRules`.
- A small role interface `IAmATransformerResolvabilityProbe` (core) answers "is this transformer registered?" via a **non-instantiating** membership lookup against the `IServiceCollection` (default impl in the DI assembly, mirroring `IAmASubscriberRegistryInspector`) — so the check never builds the transform pipeline (which throws on a missing transformer) and never instantiates a transformer.
- An async-aware `DescribeTransforms` overload unions the sync- and async-resolved mappers' transforms (de-duplicated), so a transform declared on an async-only mapper is still evaluated.
- Full rationale in **ADR 0064** (`docs/adr/0064-validate-pipeline-assembly-and-provider-registration.md`) with requirements/tasks under `specs/0035-validate-pipeline-assembly-and-provider-registration/`. Built through the spec workflow (requirements → ADR → adversarial review → tasks → TDD).

## Notes for review

- **Severity is Warning, deliberately.** The condition is deterministic (the build path throws a `ConfigurationException` when that request type is actually produced/consumed), which by the ADR 0053 precedent would be an Error. I kept it a Warning per your framing; it is flagged **revisitable** in the ADR (§3a / C-2) if you'd prefer Error or a configurable severity.
- Consistent with every existing rule, the new rules do not catch exceptions — an inspection exception surfaces as the framework's Error (ADR §3a).
- `ValidatePipelines()` should be the last call in the builder chain (the provider flags + transformer probe are snapshotted at that point) — documented on the method.

## Testing

Developer tests across `tests/Paramore.Brighter.Core.Tests/Validation/` and `tests/Paramore.Brighter.Extensions.Tests/` cover each rule's happy/negative/boundary paths (missing / all-resolvable / none / null `RequestType` / no-mapper / default-mapper / two-transforms / async-only / sync∪async union + de-dup), (B) both modalities + transposition + after-steps, the non-instantiating probe, end-to-end DI wiring, non-blocking behaviour under `throwOnError`, per-entity determinism, and the inert fallback when validation isn't wired.

Closes #4159.


iancooper (approved):
Looks good. Thanks @xbizzybone

PR #4195, opened 2026-06-19
fix: stamp scheduler fire-envelopes as MT_COMMAND

DevJonny:
## Problem

  Consumers running scheduled messages log a spurious error on every fire:

  > Message "…" mismatch. Message type is 'MT_EVENT' yet mapper produced message of type ICommand

  `MessagePump.ValidateMessageType` raises this when the wire header says `MT_EVENT`
  but the resolved request implements `ICommand`. Because the header is `MT_EVENT`,
  the pump also dispatches the envelope via `PublishAsync` rather than `SendAsync`.

  ## Root cause

  The scheduler fire-envelopes — `FireAzureScheduler` and `FireAwsScheduler` — are
  `Command`s, but their wire `MessageType` was hardcoded to `MT_EVENT`:

  - `AzureServiceBusScheduler` — all four schedule overloads (sync/async × message/request)
  - `AzureSchedulerFiredMapper.MapToMessage`
  - `AwsSchedulerFiredMapper.MapToMessage` (V3 and V4)

  On fire, the scheduler-topic consumer resolves the envelope to `FireAzureScheduler`/
  `FireAwsScheduler` (both `ICommand`) while the header says `MT_EVENT` → mismatch logged,
  and dispatch goes through publish (0..n handlers) instead of send (exactly one).

  Note: `AwsScheduler` itself already builds its envelopes with `MT_COMMAND`; only its
  mapper was inconsistent. This change aligns everything with that existing precedent.

  ## Fix

  Stamp the fire-envelope `MessageType` as `MT_COMMAND` to match the request type. The
  consumer then dispatches via `SendAsync` (correct single-handler command semantics — the
  scheduler-fired handler is the one registered handler) and the validator passes. No
  behavioural change beyond removing the spurious error and using send instead of publish.

  ## Tests

  - `Paramore.Brighter.Azure.Tests` (NUnit): `AzureSchedulerFiredMapper` + all four
    `AzureServiceBusScheduler` overloads, asserted on the scheduled `ServiceBusMessage`
    via a hand-written `FakeServiceBusSender` (no mocking libraries, no live infra).
  - `Paramore.Brighter.AWSScheduler.Tests` / `.V4.Tests` (xUnit): `AwsSchedulerFiredMapper`.

  All new tests RED before the change, GREEN after, on net9.0 and net10.0. Scheduler
  source projects build clean in Release (warnings-as-errors). Live integration suites
  (AWS SQS/SNS round-trips, Azure Blob/Service Bus) are unaffected — envelope `MessageType`
  literal only — and run in CI.

  Co-Authored-By: Claude (claude-opus-4-8) <noreply@anthropic.com>


iancooper (approved):
Thanks @DevJonny

PR #4201, opened 2026-06-22
fix(#4196): forward dead-letter reason and description to Azure Service Bus native fields

### LINKED ISSUE #4196: ASB: rejected messages dead-lettered with blank reason/description
### Symptom
Handler-rejected Azure Service Bus messages are dead-lettered with **blank** `DeadLetterReason` and `DeadLetterErrorDescription`. The real reason exists only in logs, so DLQ triage means pivoting to log search and correlating by sequence number / message id.

### Root cause
`AzureServiceBusConsumer.RejectAsync` computes the reason and description, logs them, then settles with the **no-reason** overload:

```csharp
var reasonString = reason is null ? nameof(RejectionReason.DeliveryError) : reason.RejectionReason.ToString();
var description  = reason is null ? "unknown" : reason.Description ?? "unknown";
Log.DeadLetteringMessage(Logger, message.Id.Value, lockToken, reasonString, description); // logged…
await ServiceBusReceiver!.DeadLetterAsync(lockToken);                                      // …then dropped
```

`git blame`: these fields were introduced in #3918 (Universal DLQ) and wired only into the log line; the no-reason settle predates it and was never updated. Sync `Reject` delegates here via `BrighterAsyncContext.Run`, so it shares the behaviour. This is the **only** Brighter-initiated dead-letter call site — ASB-native auto-deadlettering (TTL expiry, max-delivery-count) already populates the columns itself.

Present on `10.4.1` and `master`. The SDK (`Azure.Messaging.ServiceBus` 7.20.1) already ships `DeadLetterMessageAsync(message, deadLetterReason, deadLetterErrorDescription, ct)` (4096-char cap, requires PeekLock — which Brighter uses); the gateway just never calls it. Consistent with ADR `0045-provide-dlq-where-missing` ("defer to native") — this makes the deferral non-lossy rather than pulling ASB into the produced-message Universal DLQ path.

### Proposed fix
Add a forwarding overload to `IServiceBusReceiverWrapper`, implement it via the existing `CreateMessageShiv` pattern (as `Complete`/`DeadLetter`/`Abandon` do) with 4096-char truncation, and change the single settle call to pass `reasonString`/`description`. Result: `DeadLetterReason` ← enum name, `DeadLetterErrorDescription` ← description text. Test-first, covering async + sync paths.

### Questions
@iancooper:
1. Confirm the unforwarded reason in #3918 is an oversight and forwarding to the native overload is acceptable?
2. The overload must go on the **public** `IServiceBusReceiverWrapper` (the consumer calls through it) — a breaking change for external implementers. A default interface method isn't viable (`netstandard2.0` is in the target matrix). Accept the interface addition, or prefer a different shape?

DevJonny:
Fixes #4196.

When a handler rejects a message consumed from Azure Service Bus, AzureServiceBusConsumer.RejectAsync computed the rejection reason and description, logged them, then settled via the no-reason DeadLetterAsync overload — leaving the broker's native DeadLetterReason / DeadLetterErrorDescription fields blank. Operators triaging the DLQ saw empty columns and had to pivot to logs and correlate by sequence number.

git blame shows the reason/description were introduced in #3918 (Universal DLQ) and wired only into the log line; the no-reason settle predates it and wasn't updated. As confirmed on #4196, this is an oversight, and forwarding to the native overload is the intended fix (consistent with ADR 0045's "defer to native").

Change
- New DeadLetterAsync(lockToken, reason, description) overload on the public IServiceBusReceiverWrapper (per maintainer steer on #4196 — interface addition accepted; DIM not viable under netstandard2.0).
- Implemented via the existing message shiv against the SDK's native DeadLetterMessageAsync, truncating to ASB's 4096-char limit.
- RejectAsync forwards the values it already computes; sync Reject inherits via BrighterAsyncContext.Run.

  Testing
  - Developer tests (no mocks, fake wrapper) for both Proactor (async) and Reactor (sync) consumers.
  - Builds clean across netstandard2.0;net8.0;net9.0;net10.0; full ASB unit suite green (live-infra tests excluded).
  - No ADR (bug fix, honours existing ADR 0045); release note added.


iancooper:
@DevJonny Can you peer at this feedback for me:

"Test coverage 🔶

Per the repo's strict TDD stance, two introduced behaviors are unverified:

    Truncation — Truncate() and the 4096 boundary are entirely untested. A test with a >4096-char description asserting the forwarded value is clamped (and exactly-4096 is untouched) would lock in the limit and document why it exists.
    Null/default reason path — the reason is null → "DeliveryError"/"unknown" branch in RejectAsync and the wrapper's description is null ? null branch aren't exercised. The new tests only cover the populated-reason case.

The two happy-path tests are clear and mirror each other well across Proactor/Reactor."

Jonny-Freemarket:
Thanks for the nudge @iancooper — facepalm, I'd already forgotten about these. 🤦

Done in the latest commit:

- **Null/default reason path** — added Proactor + Reactor tests asserting that rejecting with no reason forwards the default `DeliveryError`/`unknown` to the DLQ. Good catch, that gap was real.
- **Constant naming** — `MAX_DEAD_LETTER_FIELD_LENGTH` → `MaxDeadLetterFieldLength`.

On **truncation** — leaving it untested deliberately. `Truncate` is a private guard in `ServiceBusReceiverWrapper` and per our "test public surface, not internals" line the only ways to reach it are reaching into internals or a live-broker integration test, neither of which is worth it for a defensive clamp on diagnostic text. Happy to revisit if you'd rather see it covered.

Also leaving the single-arg `DeadLetterAsync(string)` overload in place rather than `[Obsolete]`-ing it for now — didn't want to touch the public surface beyond what this fix needs.

iancooper (approved):

PR #4205, opened 2026-06-25
fix: case-insensitive SessionId and reserved-header handling for ASB (#4054)

iancooper:
## What

Fixes #4054 — an ASB `SessionId` set from a Brighter Outbox was silently dropped, and camelCased reserved keys leaked into `ApplicationProperties`.

## Root cause

Brighter's Outbox serializes the header `Bag` with `JsonSerialisationOptions.Options`, whose `PropertyNamingPolicy` **rewrites bag keys** (see `DictionaryStringObjectJsonConverter.Write`). Under the default `JsonNamingPolicy.CamelCase`, a key written as `"SessionId"` comes back as `"sessionId"` after a round-trip.

In `AzureServiceBusMessagePublisher`:

- `Bag.TryGetValue("SessionId", …)` is an exact match, so it missed the camelCased `"sessionId"` key and the session was never set on the outgoing message.
- The reserved-header filter `ReservedHeaders.Contains(h.Key)` was also case-sensitive, so the camelCased reserved keys (`sessionId`, `messageType`, `handledCount`, `replyTo`) leaked into `ApplicationProperties`.

## Fix

Resolve bag keys using the **same naming policy that serialized them**, rather than a fixed case comparison.

A first pass used `OrdinalIgnoreCase`, but that only works for the default camelCase policy (camelCase differs from PascalCase by leading case only). `JsonSerialisationOptions.Options` has a public setter, so a user can override `PropertyNamingPolicy` — under e.g. `SnakeCaseLower`, `"SessionId"` round-trips as `"session_id"` and a case-insensitive match fails again, reintroducing the bug.

`ASBConstants.IsBagKey` / `IsReservedHeader` now match a bag key against **either** the raw reserved key **or** the key as the active `PropertyNamingPolicy` renders it (read at call time, so a later `Options` override is honoured), case-insensitive on top. The publisher uses these for both the `SessionId` lookup and the reserved-header filter. This is correct under camelCase, snake_case, a custom policy, or none.

## Tests

`When_Converting_A_Message_With_A_Session_Id.cs` is now a `[Theory]` over `CamelCase`, `SnakeCaseLower`, and no policy. Each case **round-trips the `Bag` through `JsonSerialisationOptions.Options`** exactly as the Outbox does (serialize → deserialize), so the key is mangled by the real policy rather than a hand-written string, then asserts the `SessionId` is set and the reserved key is kept out of `ApplicationProperties`. The tests live in a parallelism-disabled collection that snapshots and restores `Options` in a `finally`, so mutating the global policy can't leak into other tests. The `SnakeCaseLower` case fails the earlier `OrdinalIgnoreCase` approach and passes the policy-aware one.

## Scope / follow-up

Kept to the ASB publisher. A survey found the same exact-match pattern elsewhere; those are tracked in #4214 (RocketMQ `Keys`/`Tag` broken under camelCase; AWS `messageDeduplicationId` and core `ProducerTopic` latent under a non-casing policy; `AzureServiceBusScheduler` same smell but not reproducible — it converts an empty freshly-built bag). #4214 also proposes lifting a shared policy-aware lookup into `Paramore.Brighter` core.


iancooper:
@Jonny-Freemarket Any thoughts?

iancooper:
A note to self here, I wonder if the issue is more that we need to look at the options, which the user could override, and use those when we look into the bag. Otherwise, if someone changes the JsonNamingPolicy, we can't be sure that we won't have a different bug

PR #4207, opened 2026-06-26
fix: message-pump observability fixes (#4086, #4089, #4090)

### LINKED ISSUE #4086: Pump-lifetime span (MessagePumpSpanOperation.Begin) recorded into messaging.client.operation.duration histogram
## Summary

`Reactor.Run` and `Proactor.EventLoop` open a single `pumpSpan` via `Tracer?.CreateMessagePumpSpan(MessagePumpSpanOperation.Begin, ...)` at startup and close it only when the pump exits. That span's lifetime equals the consumer worker's lifetime — potentially hours or days.

`BrighterMetricsFromTracesProcessor.OnEnd` then records that lifetime into the `messaging.client.operation.duration` histogram, because the `switch` on `messaging.operation.type` falls through `default` for `"begin"`.

## Affected sites

- `src/Paramore.Brighter.ServiceActivator/Reactor.cs:97` (open) and `:358` (close).
- `src/Paramore.Brighter.ServiceActivator/Proactor.cs:138` (open) and the matching close at the end of `EventLoop`.
- `src/Paramore.Brighter/Observability/BrighterMetricsFromTracesProcessor.cs:60-76`:

```csharp
switch (operation)
{
    case "publish": ...
    case "receive": ...
    case "process": ...
    default:
        messagingMeter.RecordClientOperation(activity);
        break;
}
```

The `"begin"` operation (from `MessagePumpSpanOperation.Begin.ToSpanName()` at `src/Paramore.Brighter/Observability/BrighterSpanExtensions.cs:69`) is the only operation that hits this `default` arm.

## Symptom

- A massive outlier appears in the `messaging.client.operation.duration` histogram every time a consumer pump shuts down — value equal to the pump's wall-clock lifetime (minutes to hours).
- Histogram percentiles (p50/p95/p99) for `messaging.client.operation.duration` are skewed and unusable for SLOs.

## Proposed fix

Either:

1. **Drop the `default` arm** so unknown operations are not counted as client operations.
2. **Special-case `"begin"`** to no-op (or record into a different, optional pump-lifecycle gauge if one is wanted).

Option 1 is the safer minimum: it preserves the explicit `publish`/`receive`/`process` recording paths and stops unrelated operations from polluting the histogram. The `MessagePumpSpanOperation.Begin` span is still emitted as a trace span — it just shouldn't feed the client-operation duration metric.

## Acceptance

- [ ] Pump shutdown does not emit a multi-hour value into `messaging.client.operation.duration`.
- [ ] Existing `publish` / `receive` / `process` recording behaviour is unchanged.
- [ ] A test asserts that ending a `MessagePumpSpanOperation.Begin` activity does not invoke `IAmABrighterMessagingMeter.RecordClientOperation`.

## Context

Surfaced while reviewing PR #4065 — pre-existing, unrelated to that PR.

### LINKED ISSUE #4089: Perf: message header serialized twice per message in pump observability hot path

### LINKED ISSUE #4090: Pump observability: pumpSpan and processSpan can leak on exception paths

iancooper:
## Summary

Rolls up three independent message-pump observability fixes, each developed diagnosis-first via the `/bugfix` workflow (triage → confirm → test-first → fix → verify). All three touch the consumer pump tracing/metrics path and each ships with regression tests.

| Issue | Fix | Area |
|-------|-----|------|
| #4086 | Exclude the pump `begin` span from the client-operation duration metric | `BrighterMetricsFromTracesProcessor` |
| #4089 | Serialize the message header once per message, not twice | `Message` / `BrighterTracer` |
| #4090 | End the pump and process spans on exception paths | `Reactor` / `Proactor` / `BrighterTracer` |

## #4086 — Pump `begin` span skews `messaging.client.operation.duration`

**Symptom:** every time a consumer pump (`Reactor.Run` / `Proactor.EventLoop`) shuts down, an outlier equal to the pump's entire wall-clock lifetime (minutes → days) is recorded into the `messaging.client.operation.duration` histogram, making p50/p95/p99 unusable for SLOs.

**Confirmed root cause:** `BrighterMetricsFromTracesProcessor.OnEnd` recorded the long-lived pump `begin` span into the client-operation-duration histogram via the `default` arm of its inner `switch (operation)`. The span opens at pump start and closes at shutdown, so its `Activity.Duration` is the whole pump lifetime.

**Fix:** added `case "begin": break;` immediately before the `default` arm — short-circuits only the pump-lifetime span; every other operation still records. (Confirm proved the issue's preferred "drop the default arm" would have regressed `create/deposit/send/clear/archive/scheduler/settle`.)

## #4089 — Message header serialized twice per message

**Symptom:** for every serviceable message, `JsonSerializer.Serialize(message.Header, …)` runs **twice** — once enriching the receive span, once creating the process span — a per-message hot-path CPU/allocation regression introduced when #4085 split the consumer span into receive + process.

**Confirmed root cause:** the receive span and process span each independently full-serialize the header (reflection over `Bag` + `Baggage`) under the default `InstrumentationOptions.Messaging` flag. 2× per serviceable message, 1× for `MT_UNACCEPTABLE`. No correctness defect — purely performance.

**Fix:** added an `internal` lazy-cached `Message.HeaderJson` (snapshot on first access) and pointed both pump-path sites (`EnrichReceiveSpan`, `CreateSpan(Process)`) at it. Header now serializes once per message; both spans share the same string. No public API change; `MessageBody`/`RequestBody` left untouched (plain string, no reflection).

## #4090 — Pump and process spans leak on exception paths

**Symptom:** two `Activity` spans can be started but never ended on exception paths — leaking activities, polluting `Activity.Current`, and skewing/dropping span durations.

**Confirmed root cause:**
- `pumpSpan` — `EndSpan(pumpSpan)` sat after the receive loop with no `try/finally`, so a throw out of the loop (e.g. the deterministic message-is-null path) skipped it and leaked the `begin` activity. Deterministic; symmetric in `Reactor` and `Proactor`.
- `processSpan` — `CreateSpan`'s post-start enrichment (`TraceStateString`, baggage, `Activity.Current`) ran outside the caller's `try`, so a throw there orphaned an already-started activity. (Confirm corrected the issue here: mirroring the receive-span pattern in the caller would have been a no-op — the fix had to live inside `CreateSpan`.)

**Fix:** wrapped the receive loop in `try/finally { EndSpan(pumpSpan) }` in both pumps (exception still propagates, preserving shutdown semantics); and inside `CreateSpan` (Process overload) a post-start throw now ends the activity (status Error) and rethrows instead of leaking it.

## Testing

- Per-fix regression tests added and green on net9.0 + net10.0.
- Full `Paramore.Brighter.Core.Tests` suite: **843 passed, 0 failed, 7 pre-existing skips** (net10.0).

## Also included

- `fix(bugfix): remove illegal command substitution from gate preflights` — a small fix to the `/bugfix` skill's preflight scripts, unrelated to runtime behaviour.

Fixes #4086
Fixes #4089
Fixes #4090


iancooper:
Thanks for the thorough review. Points **1** and **3** made me reconsider whether caching the serialized header on `Message` was the right approach at all, so I reworked #4089 rather than patching the cache.

**The reframing:** `CorrelationId` is already a top-level serialized property of `MessageHeader`, so the serialized header *always* carries it. The only thing the process span lost (vs. pre-#4089) was a *redundant duplicate* of that value nested inside the serialized `Baggage` sub-object — the same value that's also on the dedicated `ConversationId` tag of both spans. That turns point 1 from "lost value" into "redundant duplicate," which made a cleaner fix possible.

**What changed:**

1. **Removed `Message.HeaderJson`** entirely — `Message` is observability-agnostic again *(point 3: no coupling, no stale snapshot, no thread-safety question)*.

2. **Serialize the as-received header once, share it.** `EnrichReceiveSpan` now serializes the header once and returns the string; the pump threads it into `CreateSpan(Process)`, which reuses it (falling back to serializing only when there's no receive span). Both spans now carry the **identical** header tag — the existing `Assert.Same` invariant is now *intentional* rather than an accident of caching *(point 1)*.

3. **Separated baggage propagation from serialization.** The `correlationId → Baggage` + `SetBaggage` step is propagation setup, not "the headers," so it moved out of both span methods into a single `PropagateConsumerContext(message)` call made once per message by the pump. As a side effect, a malformed `correlationId` now surfaces there and is caught **non-fatally** by the pump — so a poison message no longer tears the consumer down *(eases point 5)*.

**Tests:**
- Added a non-empty-`correlationId` case asserting both spans share one header and both carry the id (the gap flagged in point 1).
- Reworked the now-moot `When_Create_Span_Throws_After_Starting…` test (its baggage-throw-inside-`CreateSpan` scenario no longer exists by construction) into one asserting the decoupled contract; the validation now surfaces from `PropagateConsumerContext`.
- Added a pump-level test proving a malformed `correlationId` doesn't stop the pump.

Full `Paramore.Brighter.Core.Tests`: **845 passed, 0 failed, 7 pre-existing skips** on net9.0 + net10.0.

On **point 4** (committed `/bugfix` scratch state) — separate concern, happy to `.gitignore` the marker files in a follow-up.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

lillo42 (approved):

PR #4215, opened 2026-07-03
fix: preserve MessageHeader.Bag keys verbatim through serialization (#4151)

### LINKED ISSUE #4151: JsonSerialisationOptions.Options applies DictionaryKeyPolicy to MessageHeader.Bag, breaking session-keyed delivery
### Describe the bug

`Paramore.Brighter.JsonConverters.JsonSerialisationOptions.Options` sets `PropertyNamingPolicy = JsonNamingPolicy.CamelCase`. `DictionaryStringObjectJsonConverter` applies that policy to dictionary **keys** as well as property names.

As a result, `MessageHeader.Bag["SessionId"]` is serialised as `"sessionId"`. On the consume side, the ASB publisher does a case-sensitive `Bag.TryGetValue("SessionId", out _)` to decide whether to set `ServiceBusMessage.SessionId`. The lookup misses, the message is sent without a session, and the ASB broker silently drops it when the destination subscription has session-required = true.

Bag keys are arbitrary string identifiers, not C# property names — applying a property naming policy to them is semantically wrong, and the failure mode is fully silent (no exception, no log line, the message just never arrives).

### To Reproduce

1. Configure a session-required ASB subscription.
2. Send a Brighter message whose `MessageHeader.Bag["SessionId"]` is set.
3. Round-trip the message through Brighter's STJ serialisation (e.g. via the scheduler envelope path, or any path using `JsonSerialisationOptions.Options`).
4. Observe the bag at the publisher: the key is now `"sessionId"`.
5. Publisher's `Bag.TryGetValue("SessionId", ...)` misses → session-less message → silent drop.

### Suggested fix

Set `DictionaryKeyPolicy = null` on `JsonSerialisationOptions.Options` (and any sibling options contributing to the envelope serialiser). Bag keys are user-defined identifiers and must round-trip verbatim.

(Alternative — case-insensitive lookup at fire time — papers over symptoms; the root issue is the bag's keys being mutated by a naming policy.)

### Further technical details

- Brighter version: **10.4.1**
- `dotnet --info`:
  ```
  .NET SDK 10.0.108 (commit 94ea82652c, MSBuild 18.0.11+94ea82652)
  Host 10.0.8 (x64)
  RID: win-x64
  ```
- OS: Windows 11 (build 10.0.26200)

### LINKED ISSUE #4214: Header-bag key lookups are not naming-policy-aware (latent SessionId/#4054-class bugs)
## Background

#4054 / #4205 fixed the ASB `AzureServiceBusMessagePublisher`, where a `SessionId` bag key was silently dropped after an Outbox round-trip. Root cause: the Outbox serializes the header `Bag` with `JsonSerialisationOptions.Options`, whose `PropertyNamingPolicy` (default `JsonNamingPolicy.CamelCase`) **rewrites bag keys** (see `DictionaryStringObjectJsonConverter.Write`). A key written as `"SessionId"` therefore comes back as `"sessionId"` — and the exact-match `TryGetValue`/`Contains` missed it.

The fix in the publisher now resolves keys via the **active naming policy** (`ASBConstants.IsBagKey` / `IsReservedHeader`), which is correct under any policy, including a user-overridden one (`JsonSerialisationOptions.Options` has a public setter).

This issue tracks the **other sites with the same pattern**. They were left out of the #4054 PR to keep its scope tight, and are categorised below by whether they reproduce today.

## Reproducibly broken under the default (camelCase) policy

- **`RocketMqMessageProducer`** — `Bag.TryGetValue(HeaderNames.Keys, …)` (`RocketMqMessageProducer.cs:156`) and `HeaderNames.Tag` (`:172`). The constants are `"Brighter-RocketMQ-Keys"` / `"Brighter-RocketMQ-Tag"`; camelCase lowercases the leading char (`brighter-RocketMQ-Keys`), so an exact-match lookup misses the key after an Outbox round-trip. Producer-side, app-supplied keys → genuinely affected.

## Latent — correct only by accident under camelCase, would break under another policy (e.g. `SnakeCaseLower`)

- **AWS `messageDeduplicationId`** — `SnsMessagePublisher.cs:84`, `SqsMessageSender.cs:100` (plus the `.V4` equivalents and `AwsScheduler`). The constant is already camelCase, so it survives the default policy — but `ConvertName` under snake_case yields `message_deduplication_id` and the exact match fails.
- **Core `ProducerTopic`** — `OutboxProducerMediator.cs:946` reads `Message.ProducerTopicHeaderName` (`"paramore.brighter.ProducerTopic"`), set in `WrapPipeline`/`WrapPipelineAsync` before the Outbox store. Leading char is already lowercase, so it survives camelCase; would break under snake_case.

## Same code smell but NOT reproducibly broken

- **`AzureServiceBusScheduler`** — `AzureServiceBusScheduler.cs:135` (`ReservedHeaders.Contains(h.Key)`, no comparer) and `:150` (`TryGetValue(ASBConstants.SessionIdKey, …)`). These look like the #4054 twin, **but** `ConvertToServiceBusMessage` runs against a freshly-constructed header whose `Bag` is empty (the original message is serialized into the *body* as a `FireAzureScheduler` payload, not the header bag). So the lookups never mismatch today. Worth hardening for consistency, but no live bug. Note the scheduler has its **own** duplicate `ASBConstants` (separate assembly), so it can't share the gateway's helper.

## Suggested direction

Provide a shared, naming-policy-aware bag-key lookup (mirroring `ASBConstants.IsBagKey`) that reads `JsonSerialisationOptions.Options.PropertyNamingPolicy` at call time and matches raw-or-`ConvertName`, case-insensitive. Prime candidate for lifting into `Paramore.Brighter` core so every gateway/transform can reuse it rather than each re-implementing exact-match lookups.

Deeper question worth a design note: reserved/system bag keys (`SessionId`, etc.) arguably shouldn't be subject to a cosmetic naming policy during persistence at all — but changing the converter is a breaking wire-format change, so the read-side policy-aware lookup is the pragmatic fix.

_Filed as follow-up to #4205._

iancooper:
## What

Fixes #4151 at the root: `MessageHeader.Bag` keys now round-trip **verbatim** through Brighter's serialization instead of being rewritten by the camelCase naming policy.

Closes #4151.
Closes #4214.

## Root cause

`DictionaryStringObjectJsonConverter.Write` applied `JsonSerialisationOptions.Options.PropertyNamingPolicy` (default `JsonNamingPolicy.CamelCase`) to dictionary **keys** as well as property names. A bag key written as `"SessionId"` was therefore serialized as `"sessionId"`. Bag keys are arbitrary user-defined identifiers, not C# property names, so any consumer that looks the key up by its original name missed it — silently.

Note (from review): `DictionaryStringObjectJsonConverter.Read` never applied the policy — it reads keys via `reader.GetString()` verbatim. So `Write` and `Read` were **asymmetric**, and even a same-process serialize→deserialize already mangled PascalCase keys. This change makes `Write` symmetric with the existing `Read`, rather than introducing new behaviour.

This is the shared root cause behind a cluster of issues/bugs:
- #4054 / #4205 — ASB `SessionId` dropped after an Outbox round-trip.
- #4132 — `Subject` put in the Bag arrived camelCased, breaking downstream consumers (e.g. Python Lambdas reading SNS notifications by PascalCase key); that PR worked around it by moving `Subject` to `MessageHeader.Subject`.
- #4214 — the survey of other exact-match bag lookups (RocketMQ `Keys`/`Tag`, AWS `messageDeduplicationId`, core `ProducerTopic`, ASB scheduler).

### Origin of the mutation

Traced to the converter's introduction in #1531 — the key transformation was copied from the reference implementation the converter is *based on* (Josef Ottoson's), not added to fix a specific Brighter problem. So there is no known behaviour that intentionally relies on it.

## Fix

Stop applying the naming policy to dictionary keys — write them verbatim. `PropertyNamingPolicy` still governs real `MessageHeader` property names, so the property wire format is unchanged; **only Bag keys** are affected.

### Scope

`DictionaryStringObjectJsonConverter` is registered globally in `JsonSerialisationOptions` and applies to **every** `Dictionary<string, object?>` Brighter serializes (`MessageHeader.Bag`, `RequestContext` bags, outbox/scheduler payloads, any internal dictionary) — not just `MessageHeader.Bag`. This is intended: dictionary keys are data, not property names, and should not be case-mangled. Because `Read` never applied the policy, no in-process consumer could have depended on the old `Write` casing.

## Why this closes #4214

#4214 catalogued the other exact-match `Bag.TryGetValue(constant)` lookups that the camelCase policy could break (RocketMQ `Keys`/`Tag`, AWS `messageDeduplicationId`, core `ProducerTopic`; plus the ASB scheduler, which reads an empty bag and was never live). Now that keys round-trip verbatim under **any** policy, every one of those lookups is correct as written — the "make each read site policy-aware" direction #4214 proposed is no longer needed. The #4205 policy-aware reader (`ASBConstants.IsBagKey`) is retained for ASB only, as a belt-and-braces compat shim for messages persisted under the old camelCasing during an upgrade drain.

## Tests

- `BagKeyRoundTripTests` — pins that PascalCase (`SessionId`, `PartitionKey`) and lowercase (`customer.header`) bag keys survive verbatim, both through the converter directly and through a full `MessageHeader` round-trip.
- `When_migrating_subject_from_bag_to_header` — previously a **characterisation test** that asserted the bug (`converts_to_camelCase`); now asserts the key is preserved. The `Header.Subject` guidance from #4132 remains valid and those tests are unchanged.
- Full `Paramore.Brighter.Core.Tests`: 849 passed, 0 failed, 7 skipped. The only test that changed behaviour was the characterisation test above.

## Risk / reviewer notes

- **In-flight wire compatibility (drain window):** messages already written to a persistent Outbox/queue under the old camelCasing hold keys like `"sessionId"`; after this change they are read back verbatim. Code looking for `"SessionId"` won't match during the drain. ASB is covered by the #4205 policy-aware reader; the other gateways folded in from #4214 (RocketMQ, AWS, core `ProducerTopic`) have **no** such shim, so their pre-upgrade in-flight messages would miss the lookup during the drain — transient and niche, but worth a release-notes callout.
- **Not exercised here:** the persistent-outbox suites (SQL / Mongo / DynamoDB) need live infra. They store the same JSON and read keys back with the same literal, so no key-level regression is expected, but they should be run in CI before merge.

### DIFF
--- a/src/Paramore.Brighter/JsonConverters/JsonSerialisationOptions.cs
+++ b/src/Paramore.Brighter/JsonConverters/JsonSerialisationOptions.cs
@@ -8,8 +8,10 @@ namespace Paramore.Brighter.JsonConverters
     /// <summary>
     /// Global Configuration for the Json Serializer
     /// We provide custom type converters, particularly around serializing objects so that they appear as types not JsonElement
-    /// The camelCase property will convert a key, such as a header bag key, to camelCase if you use UpperCase and so if you expect
-    /// an UpperCase property you will find that it is now a camelCase one. It is best to use camelCase names for this reason
+    /// The <see cref="JsonSerializerOptions.PropertyNamingPolicy"/> (camelCase) applies to C# property names only.
+    /// Dictionary keys — such as <see cref="MessageHeader.Bag"/> keys — are serialized verbatim by
+    /// <see cref="Serialization.DictionaryStringObjectJsonConverter"/>, so a key written as "SessionId" is read back
+    /// as "SessionId".
     /// </summary>
     public static class JsonSerialisationOptions
     {
--- a/src/Paramore.Brighter/MessageHeader.cs
+++ b/src/Paramore.Brighter/MessageHeader.cs
@@ -154,8 +154,9 @@ namespace Paramore.Brighter
 
         /// <summary>
         /// A property bag that can be used for extended header attributes.
-        /// Use camelCase for the key names if you intend to read it yourself, as when converted to and from Json serializers will tend convert the property
-        /// name from UpperCase to camelCase
+        /// Key names round-trip verbatim through Brighter's serialization, so you can read a key back
+        /// using the exact name you wrote (for example "SessionId"); Brighter no longer applies a naming
+        /// policy to bag keys.
         /// </summary>
         /// <value>The bag.</value>
         public Dictionary<string, object> Bag { get; set; } = new();
--- a/src/Paramore.Brighter/Serialization/DictionaryStringObjectJsonConverter.cs
+++ b/src/Paramore.Brighter/Serialization/DictionaryStringObjectJsonConverter.cs
@@ -53,18 +53,21 @@ namespace Paramore.Brighter.Serialization
 
             foreach (var entry in dictionary)
             {
+                //Bag keys are arbitrary user-defined identifiers, not C# property names, so they must
+                //round-trip verbatim. We deliberately do NOT apply PropertyNamingPolicy here: doing so
+                //rewrote keys such as "SessionId" to "sessionId", silently breaking consumers that look
+                //the key up by its original name (see #4151 / #4054).
                 var propertyName = entry.Key;
-                var convertedPropName = options.PropertyNamingPolicy?.ConvertName(propertyName) ?? propertyName;
 
                 if (entry.Value != null)
                 {
-                    writer.WritePropertyName(convertedPropName);
+                    writer.WritePropertyName(propertyName);
                     var valueConverter = new ObjectToInferredTypesConverter();
                     valueConverter.Write(writer, entry.Value, options);
                 }
                 else
                 {
-                    writer.WriteNull(convertedPropName);
+                    writer.WriteNull(propertyName);
                 }
             }
 
--- a/tests/Paramore.Brighter.Core.Tests/CloudEvents/When_migrating_subject_from_bag_to_header.cs
+++ b/tests/Paramore.Brighter.Core.Tests/CloudEvents/When_migrating_subject_from_bag_to_header.cs
@@ -8,31 +8,33 @@ using Xunit;
 namespace Paramore.Brighter.Core.Tests.CloudEvents;
 
 /// <summary>
-/// Demonstrates the V9 → V10 migration issue for Subject (PR #4132).
-/// In V9, users put Subject in the Bag and the Bag preserved PascalCase keys.
-/// In V10, the Bag serialization applies camelCase to dictionary keys, so "Subject" becomes "subject",
-/// causing downstream consumers (e.g. Python Lambdas reading SNS notifications) to fail with KeyError.
-/// The fix is to use MessageHeader.Subject instead of the Bag, which feeds the native SNS Subject field.
+/// Covers the V9 → V10 migration for Subject (PR #4132), and the root-cause fix for #4151 / #4054.
+/// In V9, users put Subject in the Bag and the Bag preserved PascalCase keys. V10 briefly applied the
+/// camelCase naming policy to dictionary keys, so "Subject" became "subject" and downstream consumers
+/// (e.g. Python Lambdas reading SNS notifications) failed with KeyError. Bag keys are arbitrary user
+/// identifiers, not C# property names, so the naming policy is no longer applied to them and the key
+/// now round-trips verbatim. Using MessageHeader.Subject (which feeds the native SNS Subject field)
+/// remains the recommended pattern.
 /// </summary>
 public class When_migrating_subject_from_bag_to_header
 {
     [Fact]
-    public void When_subject_is_in_bag_serialization_converts_to_camelCase()
+    public void When_subject_is_in_bag_serialization_preserves_the_key_verbatim()
     {
-        //Arrange - V9 pattern: user puts Subject in the Bag with PascalCase key
+        //Arrange - V9 pattern: user puts Subject in the Bag with a PascalCase key
         var message = new Message(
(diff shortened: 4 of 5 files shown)


iancooper:
Thanks — good review, and the `Read`-was-never-policy-aware observation is a sharper framing of the risk than the PR body had. Addressed in `70f0942`:

**Required (stale docs) — done:**
- `MessageHeader.Bag` XML doc no longer advises camelCasing keys; it now states keys round-trip verbatim.
- `JsonSerialisationOptions` class summary now clarifies that `PropertyNamingPolicy` applies to C# property names only, while dictionary keys are serialized verbatim by `DictionaryStringObjectJsonConverter`.

**Optional test — done:**
- Added a `MessageHeader`-level round-trip test next to the bare-dictionary pin, so the real user-facing path (Bag wired into `MessageHeader` serialization) is guarded, not just the converter in isolation.

**Global-scope note (making it explicit, as requested):**
`DictionaryStringObjectJsonConverter` is registered globally and applies to **every** `Dictionary<string, object?>` Brighter serializes — `RequestContext` bags, outbox/scheduler payloads, any internal dictionary — not just `MessageHeader.Bag`. This is intended: dictionary keys are data, not property names, and shouldn't be case-mangled. And because `Read` never applied the policy, no in-process consumer could have depended on the old asymmetric `Write` casing, so there's no hidden regression — the change makes `Write` symmetric with the `Read` that was already there.

Left as-is:
- The `Assert.Equal("a-value", roundTripped[key])` string-vs-object nit — agree it's not worth changing.

Still open before merge (flagged, not code): run the persistent-outbox suites (SQL/Mongo/Dynamo) in CI, and a release-notes callout for the drain-window compat (covered by keeping #4205's policy-aware reader).

Issue #4224, opened 2026-07-06
Spanner Outbox/Inbox emulator tests collide on concurrent DDL (need serialized collection)


iancooper:
## Summary

After #4162 (missing `EmulatorDetection`) is fixed, ~28 Spanner Outbox/Inbox tests fail against the emulator with:

```
Google.Cloud.Spanner.Data.SpannerException : Operation was rejected because the system is not in a state required for the operation's execution.
---- Grpc.Core.RpcException : Status(StatusCode="FailedPrecondition", Detail="Schema change operation rejected because a concurrent schema change operation or read-write transaction is already in progress.")
```

The Spanner **emulator serialises DDL**, but the Outbox (`SpannerBinary`/`SpannerText` × `Sync`/`Async`) and Inbox test classes run in parallel, and their fixture `CREATE`/`DROP TABLE` calls collide on the shared `brightertests` database. The passing `BoxProvisioning` tests avoid this via `[Collection("SpannerBoxProvisioning")]`; the Outbox/Inbox test classes have no such collection.

This was **previously masked** by the ADC fixture-init failure from #4162 — once `EmulatorDetection` lets the tests reach the DDL stage, the concurrency limitation surfaces. Different root cause (test parallelism vs emulator DDL serialisation).

## Evidence

- Each affected test **passes in isolation**.
- Full `~Spanner` filter with collection parallelisation **disabled** is green:
  ```bash
  SPANNER_EMULATOR_HOST=localhost:9010 GOOGLE_CLOUD_PROJECT=brighter-tests \
    dotnet test tests/Paramore.Brighter.Gcp.Tests/Paramore.Brighter.Gcp.Tests.csproj -f net9.0 \
    --filter "FullyQualifiedName~Spanner" -- xUnit.ParallelizeTestCollections=false
  # 94 passed / 0 failed
  ```
  Under the default parallel run: ~66 passed / ~28 failed (which tests fail is nondeterministic — it's a race).

## Candidate fix (test-only)

Serialise the Spanner Outbox/Inbox test classes:
- Place them in a shared xUnit collection (mirroring `BoxProvisioning`'s `[Collection("SpannerBoxProvisioning")]`), **or**
- Disable parallelisation for that set.

## Context

[...]

PR #4225, opened 2026-07-06
fix: honour Spanner emulator detection in outbox/inbox connections (#4162)

### LINKED ISSUE #4162: Spanner Outbox/Inbox tests fail against the emulator (missing EmulatorDetection on test SpannerConnection)
## Summary

When run against the Spanner emulator (per `setup-spanner-emulator.sh`), 60 Spanner Outbox and Inbox tests fail at fixture setup with `System.InvalidOperationException : Your default credentials were not found. To set up Application Default Credentials, see https://cloud.google.com/docs/authentication/external/set-up-adc.` — the test helpers construct `SpannerConnection` directly from the connection string, which does not opt into `EmulatorDetection`, so the gRPC client tries to load ADC instead of routing to `SPANNER_EMULATOR_HOST`.

The fix is small and contained: route Outbox/Inbox test helpers through `SpannerConnectionStringBuilder { EmulatorDetection = EmulatorDetection.EmulatorOrProduction }` the way the BoxProvisioning Spanner tests already do.

## Repro

```bash
docker compose -f docker-compose-spanner.yaml up -d
bash ./setup-spanner-emulator.sh
SPANNER_EMULATOR_HOST=localhost:9010 GOOGLE_CLOUD_PROJECT=brighter-tests \
  dotnet test tests/Paramore.Brighter.Gcp.Tests/Paramore.Brighter.Gcp.Tests.csproj -f net9.0 \
  --filter "FullyQualifiedName~Spanner"
```

Result on `multi-tenancy@c6335eb00`: `Failed: 60, Passed: 33, Total: 93`. The 33 passing are exactly the `Spanner/BoxProvisioning/` tests.

## Failure breakdown

| Group | Count |
| --- | ---: |
| `Paramore.Brighter.Gcp.Tests.Outbox.SpannerBinary.{Sync,Async}` | 24 |
| `Paramore.Brighter.Gcp.Tests.Outbox.SpannerText.{Sync,Async}` | 24 |
| `Paramore.Brighter.Gcp.Tests.Spanner.Inbox.SpannerInbox{,Async}Test` | 12 |
| **Total** | **60** |

All fail with the identical "credentials were not found" message at fixture-init time.

## Root cause (located)

The failing test helpers construct the connection without enabling emulator detection. Example — `tests/Paramore.Brighter.Gcp.Tests/Spanner/Inbox/SpannerInboxAsyncTest.cs:23`:

```csharp
protected override async Task CreateInboxTableAsync(RelationalDatabaseConfiguration configuration)
{
    await using var connection = new SpannerConnection(configuration.ConnectionString);  // <-- no EmulatorDetection
    await connection.OpenAsync();
    ...
}
```

The Spanner BoxProvisioning tests (which all pass) use the pattern that works. Example — `tests/Paramore.Brighter.Gcp.Tests/Spanner/BoxProvisioning/When_spanner_outbox_provisioner_runs_on_fresh_database_it_should_create_outbox_table.cs`:

```csharp
private SpannerConnection CreateConnection()
{
    var builder = new SpannerConnectionStringBuilder(_connectionString)
    {
        EmulatorDetection = EmulatorDetection.EmulatorOrProduction
    };
    return new SpannerConnection(builder);
}
```

`EmulatorDetection.EmulatorOrProduction` makes the SDK check `SPANNER_EMULATOR_HOST` and bypass ADC when set; without it, the SDK falls back to ADC and fails when no creds are configured.

## Proposed fix

Apply the `EmulatorDetection.EmulatorOrProduction` connection-builder pattern to the Outbox/Inbox test setup helpers (test-only change). Candidate files:

- `tests/Paramore.Brighter.Gcp.Tests/Spanner/Inbox/SpannerInboxTest.cs` (sync helpers)
- `tests/Paramore.Brighter.Gcp.Tests/Spanner/Inbox/SpannerInboxAsyncTest.cs` (async helpers)
- The Outbox test base/builder helpers used by `Outbox/Spanner{Binary,Text}/{Sync,Async}/` — likely `tests/Paramore.Brighter.Gcp.Tests/Outbox/...` shared fixture (worth a `grep -rn 'new SpannerConnection' tests/Paramore.Brighter.Gcp.Tests`).

Consider extracting the builder pattern into a `tests/Paramore.Brighter.Gcp.Tests/Spanner/SpannerTestConnectionFactory.cs` helper used by every test that constructs a `SpannerConnection`.

Out-of-scope question for a future ADR: should the production-side `SpannerInbox` / `SpannerOutbox` honour `SPANNER_EMULATOR_HOST` by default, or is that strictly a test concern? Today the test-only fix is the right scope.

## Setup-script gap (not the root cause but worth noting)

`setup-spanner-emulator.sh` exports env vars in its own subshell. Running `bash ./setup-spanner-emulator.sh` does NOT export to the caller; you must either `source ./setup-spanner-emulator.sh` or prefix env inline. This is documented at the end of the script, but a reader may miss it. (The emulator detection fix above subsumes this — once `EmulatorDetection` is on the connection, the env var is the SDK's contract, not the test runner's.)

## Out-of-scope context

Uncovered during the VERIFY phase of spec 0029 (multi-tenancy migration-history scope) on branch `multi-tenancy` (PR #4155). That spec touches only `src/Paramore.Brighter.BoxProvisioning*/` — zero Spanner Outbox/Inbox code — so the failures are pre-existing and orthogonal. The `Spanner/BoxProvisioning/` subset (`--filter "FullyQualifiedName~Spanner.BoxProvisioning"`) is 33/33 on both TFMs on the same branch.

iancooper:
## Symptom

Run against the Spanner emulator, 60 Spanner Outbox/Inbox tests failed at fixture setup with:

```
System.InvalidOperationException : Your default credentials were not found.
```

The gRPC client tried to load Application Default Credentials instead of routing to `SPANNER_EMULATOR_HOST`. The 33 `Spanner/BoxProvisioning/` tests passed because they already opt into emulator detection.

## Confirmed root cause

`SpannerConnectionProvider` and `SpannerUnitOfWork` built `new SpannerConnection(configuration.ConnectionString)` **without opting into `EmulatorDetection`**. Because `EmulatorDetection` is opt-in and absent, the client falls back to ADC and throws. The passing `BoxProvisioning` tests work precisely because their production `SpannerConnectionHelper` sets `EmulatorDetection.EmulatorOrProduction`.

Evidence (code-trace, verified against the emulator):
- The internal providers are exercised by the outbox/inbox test bodies (`SpannerConnectionProvider.cs:22/30`, `SpannerUnitOfWork.cs:28/41`), rebuilding from `configuration.ConnectionString` — so a test-object-only fix is insufficient; the fix must reach the connection string that round-trips through `RelationalDatabaseConfiguration`.
- `SpannerConnectionStringBuilder.EmulatorDetection` serialises into `.ConnectionString` (Google.Cloud.Spanner.Data 5.12.0), so embedding it once propagates to every derived connection.
- `EmulatorDetection.EmulatorOrProduction` is a **no-op in real production** — it only routes to the emulator when `SPANNER_EMULATOR_HOST` is set.

## Fix

- **Production parity** — `SpannerConnectionProvider` and `SpannerUnitOfWork` now build their connection string via `SpannerConnectionStringBuilder { EmulatorDetection = EmulatorDetection.EmulatorOrProduction }`, matching the existing `SpannerConnectionHelper`.
- **Test-side** — the Gcp test connection string (`Const.ConnectionString`) embeds `EmulatorDetection` so the pre-existing Spanner outbox/inbox tests reach the emulator.
- **Regression test** — `SpannerConnectionProviderEmulatorTests` builds the production provider from a bare connection string and asserts it routes to the emulator (was red with the ADC error, now green).

## Verification (emulator)

- New regression test: green.
- `--filter "FullyQualifiedName~Spanner"` with collection parallelisation disabled: **94 passed / 0 failed** (33 BoxProvisioning kept green — no regression; all 60 ADC failures resolved).

## Out of scope — tracked separately (#4224)

Under a **parallel** `dotnet test` run, ~28 Spanner Outbox/Inbox tests still fail with `FailedPrecondition: Schema change operation rejected because a concurrent schema change operation ... is already in progress`. That is a **different** root cause — the emulator serialises DDL while those test classes run in parallel (unlike `BoxProvisioning`, which uses a shared `[Collection]`). It was previously masked by the ADC failure and only surfaced once emulator detection let the tests reach the DDL stage. Filed as #4224; each test passes in isolation / when serialised.

Fixes #4162


src/Paramore.Brighter.Spanner/SpannerConnectionProvider.cs
```diff
@@ -14,7 +15,13 @@ namespace Paramore.Brighter.Spanner;
...
     : RelationalDbConnectionProvider
 {
-    private readonly string _connectionString = configuration.ConnectionString;
+    // Enable emulator detection so the connection routes to SPANNER_EMULATOR_HOST when it is set,
+    // and behaves as normal production otherwise. The keyword serialises into the connection string.
+    private readonly string _connectionString =
+        new SpannerConnectionStringBuilder(configuration.ConnectionString)
+        {
+            EmulatorDetection = EmulatorDetection.EmulatorOrProduction
+        }.ConnectionString;
```
lillo42, line 25: Should it be done in the `IAmARelationalDatabaseConfiguration`?

Like 
```csharp
new RelationalDatabaseConfiguration(new SpannerConnectionStringBuilder("MyConnection")
        {
            EmulatorDetection = EmulatorDetection.EmulatorOrProduction
        }.ConnectionString)

```

iancooper: The RelationalDatabaseConfiguration is the constructor parameter. This extracts the connection string from that, since it is required to implement the underlying properties. This hasn't changed here (it always worked this way). The only change is that we insert the emulator string if needed. If RelationalDatabaseConfiguration performed any calculations on the property, we might have wanted to call the constructor again, with the modified string, but since it's just a data structure, it doesn't do anything, and would be an extra allocation. I think we might as well keep the calculation local here as a result

Issue #4190, opened 2026-06-17
Confirm that Kafka Gateway supports KIP-848


iancooper:
**Is your feature request related to a problem? Please describe.**
KIP-848 introduces a new rebalancing protocol in which the broker coordinates rebalancing rather than the consumer. It is efficient, similar to cooperative rebalance.

It was added to Confluent.NET

https://github.com/confluentinc/confluent-kafka-dotnet/pull/2529/changes

**Describe the solution you'd like**

We need to validate that you can use this protocol with Brighter; this may involve exposing properties that we do not have in the config

PR #4233, opened 2026-07-14
GH-4190 Add support to KIP-848

lillo42:
## Description

Add support for [KIP-848](https://cwiki.apache.org/confluence/display/KAFKA/KIP-848%3A+The+Next+Generation+of+the+Consumer+Rebalance+Protocol) — Kafka's new consumer group protocol — alongside the existing classic protocol.

- Introduces an `IGroupProtocol` strategy applied to the `ConsumerConfig` at consumer creation:
  - `ClassicGroupProtocol` (default): carries `SessionTimeout`, `HeartbeatInterval`, and `PartitionAssignmentStrategy`. When no protocol is supplied, the consumer back-fills any `null` properties from its constructor parameters (session timeout default 10 s, assignment strategy default `RoundRobin`), preserving current behavior. The back-fill mutates the instance in place, so an instance must not be shared across subscriptions (documented on the type and on `KafkaSubscription.GroupProtocol`).
  - `ConsumerGroupProtocol` (KIP-848): carries `GroupRemoteAssignor` and `GroupInstanceId`, and deliberately does **not** set `session.timeout.ms`, `heartbeat.interval.ms`, or `partition.assignment.strategy`, which librdkafka rejects when `group.protocol=consumer`.
- `KafkaSubscription.GroupProtocol` selects the protocol; the legacy `SessionTimeout` and `PartitionAssignmentStrategy` properties on `KafkaSubscription` are obsoleted (with `#pragma` guards at call sites) in favor of `ClassicGroupProtocol`.
- XML documentation covers the back-fill precedence and no-sharing contract, the static-membership caveat (`GroupInstanceId` must be unique per consumer instance — only set it with a single performer, or assign unique ids via `configHook`), and the weaker infrastructure-validation guarantees of the consumer protocol (broker-driven membership completes asynchronously, so `OnMissingChannel.Validate` cannot reliably detect missing topics; prefer `Create` or provision topics out-of-band).

### Test infrastructure

- Migrated the Kafka test broker to KRaft (`apache/kafka`, no ZooKeeper) in `docker-compose-kafka.yaml` and the CI workflow; images pinned to explicit versions (`apache/kafka:4.0.2`, `cp-schema-registry:8.0.6`, `cp-enterprise-control-center:7.9.8`).
- Added a generated "Consumer" messaging-gateway test suite (KIP-848 protocol) alongside the renamed "Classic" suite; infrastructure-validation tests are excluded for the consumer protocol for the reason above.
- Added broker-free unit tests over `ClassicGroupProtocol.Apply` / `ConsumerGroupProtocol.Apply` that lock in the KIP-848 config contract (consumer protocol never sets the classic-only keys).
- `KafkaMessageAssertion` now compares timestamps normalized to UTC seconds (the previous string comparison was timezone-sensitive).

## Related Issues

https://github.com/BrighterCommand/Brighter/issues/4190

## Type of Change

- [ ] Documentation update

## Checklist

## Additional Notes

Addresses the automated review feedback: documented the `IGroupProtocol` no-sharing contract and the `GroupInstanceId` static-membership caveat, added unit tests over the `Apply` mappings, pinned the Docker images, and fixed the cosmetic issues (trailing whitespace/newline, double space).


.github/workflows/ci.yml
```diff
@@ -283,27 +283,29 @@ jobs:
...
-         ALLOW_ANONYMOUS_LOGIN: yes
       kafka:
-        image: confluentinc/cp-enterprise-kafka:latest
+        image: apache/kafka:latest
         ports:
           - 9092:9092
         env:
-          KAFKA_BROKER_ID: 1
-          KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
+          # 1. KRaft Specific Configurations
```
iancooper, line 291: Let's see if this stable. I have had a lot of trouble with Zookeeper free Kafka on courses.

src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageConsumer.cs
```diff
@@ -129,8 +132,10 @@ public KafkaMessageConsumer(
             RoutingKey? deadLetterRoutingKey = null,
             RoutingKey? invalidMessageRoutingKey = null,
             TimeProvider? timeProvider = null,
-            IAmAMessageScheduler? scheduler = null
-            )
+            IAmAMessageScheduler? scheduler = null,
+            GroupProtocol groupProtocol = GroupProtocol.Classic,
```
iancooper: I wonder if we should not group these up, as they operate together, perhaps in a record, rather than passing them separately? Let's imagine we call it RebalancePolicy. That would allow us to treat the absence of this record as an indicator to use classic, and the presence of a RebalancePolicy as an indicator to use the Consumer Group Protocol. RebalancePolicy would then hold the RemotesAssignor and GroupInstanceId

src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageConsumer.cs
```diff
@@ -190,17 +195,27 @@ public KafkaMessageConsumer(
...
                 AllowAutoCreateTopics = false, //We will do this explicit always so as to allow us to set parameters for the topic
                 IsolationLevel = isolationLevel,
                 //We commit the last offset for acknowledged requests when a batch of records has been processed. 
                 EnableAutoOffsetStore = false,
                 EnableAutoCommit = false,
-                // https://www.confluent.io/blog/cooperative-rebalancing-in-kafka-streams-consumer-ksqldb/
-                PartitionAssignmentStrategy = partitionAssignmentStrategy,
             };
+
+            if (groupProtocol == GroupProtocol.Classic)
```
iancooper: An if on an enum doesn't allow for other values. Consider a switch with a ConfigurationException on default? Otherwise, what happens on fall through?

src/Paramore.Brighter.MessagingGateway.Kafka/KafkaSubscription.cs
```diff
@@ -117,6 +117,26 @@ public class KafkaSubscription : Subscription, IUseBrighterDeadLetterSupport, IU
         /// How long to wait when asking for topic metadata
         /// </summary>
         public TimeSpan TopicFindTimeout { get; set; } = TimeSpan.FromMilliseconds(5000);
+
+        /// <summary>
+        /// Which Kafka group coordination protocol to use.
+        /// Use <see cref="Confluent.Kafka.GroupProtocol.Classic"/> for the classic consumer group protocol,
+        /// or <see cref="Confluent.Kafka.GroupProtocol.Consumer"/> for the newer consumer protocol.
+        /// </summary>
+        public GroupProtocol GroupProtocol { get; set; } = GroupProtocol.Classic;
```
iancooper: Possibly group here too

iancooper (approved):
Nothing wrong with the way it is, so approved, but I would be interested if we could collate those parameters better.

PR #4242, opened 2026-07-18
Fix outbox clear isolation between mediators

### LINKED ISSUE #4241: OutboxProducerMediator can skip clears across independent instances
## Problem

`OutboxProducerMediator<T>` uses a static `SemaphoreSlim` for nonblocking background clears. Static state is shared by every mediator instance for the same closed generic `T`.

When two independent mediators—potentially backed by different outboxes—call the clear path concurrently, the first acquires the semaphore and the second fails `WaitAsync(TimeSpan.Zero)`. The second clear then returns without processing its own outbox.

## Location

- `src/Paramore.Brighter/OutboxProducerMediator.cs`: static `s_backgroundClearSemaphoreToken`
- The clear path calls `WaitAsync(TimeSpan.Zero, cancellationToken)` and only clears after acquisition.

## Impact

An unrelated mediator can suppress another mediator's requested outbox clear. Enabling test parallelism in PR #4065 exposed this as alternating confirmation-test failures rather than a deterministic test-isolation failure.

## Proposed direction

Scope exclusion to the mediator/outbox instance instead of the closed generic type. Add a deterministic concurrent regression test using two independent mediators/outboxes and prove both requested clears execute.

Related: #4065

thomhurst:
## Summary

- scope the nonblocking background-clear semaphore to each `OutboxProducerMediator` instance
- preserve suppression of overlapping clears on the same mediator
- add a deterministic concurrent regression test using two independent mediators and outboxes

## Root cause

The semaphore was static for each closed generic mediator type. A clear running on one mediator therefore caused another mediator with a different outbox to fail `WaitAsync(TimeSpan.Zero)` and silently skip its own clear.

## Validation

- regression test failed before the fix on net9.0 and net10.0 (`Expected: 1`, `Actual: 0`)
- regression test passes after the fix on net9.0 and net10.0
- Core command-processor clear suite: 4 passed, 2 existing skips per target framework
- InMemory sweeper suite: 10 passed per target framework

Closes #4241

### DIFF
--- a/src/Paramore.Brighter/OutboxProducerMediator.cs
+++ b/src/Paramore.Brighter/OutboxProducerMediator.cs
@@ -63,7 +63,7 @@ namespace Paramore.Brighter
         private readonly IAmAnOutboxCircuitBreaker? _outboxCircuitBreaker;
         private readonly ConcurrentDictionary<string, List<TMessage>> _outboxBatches = new();
 
-        private static readonly SemaphoreSlim s_backgroundClearSemaphoreToken = new(1, 1);
+        private readonly SemaphoreSlim _backgroundClearSemaphore = new(1, 1);
 
         //Used to checking the limit on outstanding messages for an Outbox. We throw at that point. Writes to the static
         //bool should be made thread-safe by locking the object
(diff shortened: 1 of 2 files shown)


iancooper (approved):

thomhurst:
@iancooper Happy for me to merge?

PR #4245, opened 2026-07-19
Await in-memory publish confirmation handling

### LINKED ISSUE #4243: InMemoryMessageProducer.DisposeAsync does not await async publish-confirmation handling
### Describe the bug

`InMemoryMessageProducer.DisposeAsync` returns before asynchronous publish-confirmation handlers have completed.

Its documented contract says disposal drains the worker and waits until all confirmation callbacks have returned, allowing callers to rely on callbacks having fired before disposal completes. However, `OnMessagePublished` is an `Action<PublishConfirmationResult>`, while `OutboxProducerMediator.ConfigureAsyncPublisherCallbackMaybe` subscribes with an `async` delegate. That delegate is therefore `async void`.

The producer stores this invocation:

```csharp
_raiseTasks.Add(Task.Run(() => OnMessagePublished?.Invoke(successResult)));
```

Awaiting that task waits only until the `async void` handler reaches its first incomplete `await`, not until confirmation processing finishes. In the success path, `MarkDispatchedAsync` can therefore still be running after `DisposeAsync` returns.

The source comment in `InMemoryMessageProducer.DrainAsync` already notes this limitation, but it contradicts the public disposal contract and causes observable outbox state races.

### To Reproduce

PR #4065 exposes the race under TUnit parallel scheduling:

1. Create an `InMemoryMessageProducer` with `UseAsyncPublishConfirmation = true`.
2. Register it with an `OutboxProducerMediator` backed by an `InMemoryOutbox`.
3. Add outstanding messages.
4. Call `ClearOutstandingFromOutboxAsync(..., useBulk: true, ...)`.
5. Await `producer.DisposeAsync()`.
6. Immediately query `DispatchedMessages`.

Expected: every successful confirmation has completed and both messages are marked dispatched.

Actual: disposal can return first, leaving a successfully published message temporarily undispatched.

Failing CI job:
https://github.com/BrighterCommand/Brighter/actions/runs/29664202068/job/88131793631

Failure:

```text
When_bulk_dispatching_a_confirming_producer_dispatches_via_confirmation
Expected to be true but found False
at Assert.That((dispatched).Any(m => m.Id == first.Id)).IsTrue()
```

The same run passed the assembly on `net9.0` and failed on `net10.0`; an earlier run passed both. Parallel load changes scheduling and widens the race, but does not create it.

Relevant locations:

- `src/Paramore.Brighter/InMemoryMessageProducer.cs`: `OnMessagePublished`, `DisposeAsync`, and `DrainAsync`
- `src/Paramore.Brighter/OutboxProducerMediator.cs`: `ConfigureAsyncPublisherCallbackMaybe`
- `tests/Paramore.Brighter.Core.Tests/Confirmation/When_bulk_dispatching_a_confirming_producer.cs`

### Exceptions (if any)

No production exception is required. Failure is premature completion and stale outbox state.

### Expected behaviour

Awaiting `DisposeAsync` should satisfy its documented drain contract: all queued publish confirmations and their asynchronous handlers must be complete before it returns.

### Proposed direction

Make asynchronous confirmation handling awaitable instead of using an `Action` with an `async void` subscriber. Possible approaches:

- expose/register `Func<PublishConfirmationResult, Task>` handlers and await them;
- provide a separate asynchronous confirmation event/registration path while preserving compatibility for synchronous producers;
- introduce an explicit completion mechanism owned by the confirmation pipeline.

Polling or delaying the test would hide the lifetime bug and should not be the primary fix.

### Further technical details

- Present on current `master`-based code; PR #4065 changes test runner/scheduling, not this production implementation.
- Observed on GitHub Actions Ubuntu runner.
- Observed with `net10.0`; timing-dependent rather than framework-specific.

thomhurst:
## Summary

- add a public `Task`-returning publish-confirmation callback (`ISupportPublishConfirmationAsync`) alongside the existing fire-and-forget event
- make the in-memory confirmation pump await the mediator's full async dispatch update
- roll the awaited callback out to the broker producers (per review): RMQ.Async awaits it on the ack loop; RMQ.Sync and Kafka track the callbacks with a shared `InFlightCallbackTracker` and drain them at dispose
- consolidate shared plumbing: `InvokeAllAsync` for the awaited multicast raise, `StartConfirmationSpan`/`EndConfirmationSpan` in the mediator, and a discard-lambda instead of `async void` for the sync-event fallback

## Behavioural note (for release notes)

RMQ.Async ack processing is now paced by the confirmation callback: the outbox `MarkDispatchedAsync` write is awaited on the channel's ack-dispatch loop (fanned out concurrently within a multiple-ack batch), so a slow outbox slows ack draining for that producer's channel. This is intentional backpressure — previously the write raced dispose and could be lost to a sweeper re-delivery.

The wait is not bounded by Brighter itself: the default `CommandProcessor.OutboxProducer` resilience pipeline is retry-only, so the effective bound is the outbox client's own timeout (e.g. DB command timeout). Operators who want a hard ceiling can add a Polly timeout strategy to the `CommandProcessor.OutboxProducer` pipeline. Nothing is lost either way — an unmarked message is re-swept.

A multiple-ack batch settles its `MarkDispatchedAsync` writes concurrently (`Task.WhenAll`), so under RMQ.Async the configured outbox must tolerate concurrent `MarkDispatchedAsync` calls; standard connection-per-call DB outboxes do.

Also fixed in passing: `RMQ.Sync.RmqMessageProducer.DisposeAsync` previously returned a `TaskCompletionSource` task that was never completed, so awaiting it hung forever; it now completes after the synchronous dispose (which this PR routes the confirmation drain through).

## Root cause

The mediator subscribed an `async void` handler to `ISupportPublishConfirmation.OnMessagePublished`. The in-memory producer's drain task therefore completed at the handler's first incomplete await, allowing `DisposeAsync()` to return before `MarkDispatchedAsync` finished. Increased test parallelism exposed that race. The same gap existed for the broker producers: their dispose paths waited for broker acks but not for the outbox `MarkDispatchedAsync` those acks trigger.

## Validation

- `Paramore.Brighter.Core.Tests`: 855 passed, 0 failed (7 pre-existing skips)
- `Paramore.Brighter.InMemory.Tests`: 144 passed, 0 failed
- New tests: dispose-awaits-callback (gated outbox), default-path stays non-blocking, span-teardown isolation, broker-style producer awaits `MarkDispatchedAsync`, mediator registers exactly one callback
- RMQ.Async/RMQ.Sync/Kafka producer and test projects build clean; their integration suites need broker infrastructure and were not run here

Closes #4243

### DIFF
--- a/src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageProducer.cs
+++ b/src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageProducer.cs
@@ -73,12 +73,10 @@ namespace Paramore.Brighter.MessagingGateway.Kafka
         private bool _hasFatalProducerError;
         private readonly InstrumentationOptions _instrumentation;
         private event Func<PublishConfirmationResult, Task>? _onMessagePublishedAsync;
-        // Confirmation raises run on worker tasks (never on Confluent's poll thread); this counter and
-        // TCS let Dispose wait for those callbacks — including the awaited Outbox mark-dispatched —
-        // after Flush() has drained the delivery reports themselves. Guarded by _confirmationStateLock.
-        private readonly object _confirmationStateLock = new();
-        private int _inFlightConfirmationCallbacks;
-        private TaskCompletionSource<bool>? _confirmationCallbacksCompleted;
+        // Confirmation raises run on worker tasks (never on Confluent's poll thread); the tracker
+        // lets Dispose wait for those callbacks — including the awaited Outbox mark-dispatched —
+        // after Flush() has drained the delivery reports themselves.
+        private readonly InFlightCallbackTracker _confirmationCallbacks = new();
 
         public KafkaMessageProducer(
             KafkaMessagingGatewayConfiguration configuration, 
@@ -450,20 +448,14 @@ namespace Paramore.Brighter.MessagingGateway.Kafka
             // invoke so a faulting subscriber is logged rather than left as an unobserved Task exception
             // (which can escalate via TaskScheduler.UnobservedTaskException). Brighter's own mediator
             // callback is already self-contained; this guards any other subscriber and the broker thread.
-            // The in-flight counter must be released on every path or dispose would block on its timeout.
-            BeginConfirmationCallback();
+            // The in-flight tracker must be released on every path or dispose would block on its timeout.
+            _confirmationCallbacks.Begin();
             Task.Run(async () =>
             {
                 try
                 {
                     OnMessagePublished?.Invoke(result);
-
-                    var handlers = _onMessagePublishedAsync;
-                    if (handlers is not null)
-                    {
-                        foreach (Func<PublishConfirmationResult, Task> handler in handlers.GetInvocationList())
-                            await handler(result);
-                    }
+                    await _onMessagePublishedAsync.InvokeAllAsync(result);
                 }
                 catch (Exception ex)
                 {
@@ -471,57 +463,15 @@ namespace Paramore.Brighter.MessagingGateway.Kafka
                 }
                 finally
                 {
-                    EndConfirmationCallback();
+                    _confirmationCallbacks.End();
                 }
             });
         }
 
-        private void BeginConfirmationCallback()
-        {
-            lock (_confirmationStateLock)
-            {
-                if (_inFlightConfirmationCallbacks == 0)
-                    _confirmationCallbacksCompleted = new TaskCompletionSource<bool>(TaskCreationOptions.RunContinuationsAsynchronously);
-
-                _inFlightConfirmationCallbacks++;
-            }
-        }
-
-        private void EndConfirmationCallback()
-        {
-            lock (_confirmationStateLock)
-            {
-                _inFlightConfirmationCallbacks--;
-
-                if (_inFlightConfirmationCallbacks == 0)
-                    _confirmationCallbacksCompleted?.TrySetResult(true);
-            }
-        }
-
         private void WaitForConfirmationCallbacks()
         {
-            Task? callbacksCompleted;
-            lock (_confirmationStateLock)
-            {
-                if (_inFlightConfirmationCallbacks == 0)
-                    return;
-
-                callbacksCompleted = _confirmationCallbacksCompleted?.Task;
-            }
-
-            if (callbacksCompleted is null)
-                return;
-
-            if (!callbacksCompleted.Wait(TimeSpan.FromMilliseconds(ConfirmationCallbacksShutdownTimeoutMs)))
-            {
-                int inFlightCallbacks;
-                lock (_confirmationStateLock)
-                {
-                    inFlightCallbacks = _inFlightConfirmationCallbacks;
-                }
-
-                Log.FailedToAwaitConfirmationCallbacks(s_logger, inFlightCallbacks, ConfirmationCallbacksShutdownTimeoutMs);
-            }
+            if (!_confirmationCallbacks.TryWait(TimeSpan.FromMilliseconds(ConfirmationCallbacksShutdownTimeoutMs), out int stillInFlight))
+                Log.FailedToAwaitConfirmationCallbacks(s_logger, stillInFlight, ConfirmationCallbacksShutdownTimeoutMs);
         }
 
         private static partial class Log
--- a/src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageProducer.cs
+++ b/src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageProducer.cs
@@ -32,13 +32,26 @@ using Paramore.Brighter.Tasks;
 
 namespace Paramore.Brighter.MessagingGateway.Kafka
 {
-    public partial class KafkaMessageProducer : KafkaMessagingGateway, IAmAMessageProducerSync, IAmAMessageProducerAsync, ISupportPublishConfirmation
+    public partial class KafkaMessageProducer : KafkaMessagingGateway, IAmAMessageProducerSync, IAmAMessageProducerAsync, ISupportPublishConfirmation, ISupportPublishConfirmationAsync
     {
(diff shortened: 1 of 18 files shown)


iancooper:
@thomhurst 

>keep RabbitMQ, Kafka, and other broker producers on the existing confirmation event

Given that this event is only ever consumed internally, is this the right choice? If we have fixed a bug with async void terminating early, I wonder if it makes sense to apply that fix broadly, to avoid divergence between the behavior of in-memory and broker-based solutions

iancooper (commented):
See my comment about rolling out more broadly

thomhurst:
Done — rolled out broadly in this PR. `ISupportPublishConfirmationAsync` is now public and all three confirming broker producers implement it, so the mediator's confirmation callback (span + `MarkDispatchedAsync` / `TripTopic`) is no longer async void anywhere:

- **RMQ.Async** — the ack/nack handlers now await the async subscribers directly (RabbitMQ.Client awaits `BasicAcksAsync` handlers, so this is a natural fit). Disposal previously waited only for the broker's acks; the drain TCS now also covers callbacks still in flight, so `DisposeAsync` waits for the full `MarkDispatchedAsync` before tearing down the channel. Subscriber faults are isolated per confirmation so one bad callback can't abort the rest of a multiple-ack batch.
- **Kafka** — callbacks must not block Confluent's poll thread, so they still run on worker tasks, but they're now counted in flight and `Dispose` waits for them (bounded at 5s) after `Flush()` drains the delivery reports.
- **RMQ.Sync** — same pattern as Kafka: the awaited subscribers run on a worker task (blocking the connection loop would deadlock `WaitForConfirms`) and `Dispose` waits for them, bounded by `WaitForConfirmsTimeOutInMilliseconds`. One caveat: the sync client doesn't guarantee the ack event handlers have run by the time `WaitForConfirms` returns, so this one is best-effort — a callback the client hasn't dispatched yet isn't counted. The async client has no such gap.

The existing `Action`-based `OnMessagePublished` event stays untouched for external subscribers; the mediator registers exactly one callback — the awaited event when `UseAsyncPublishConfirmation` is true, which the broker producers hard-code since their confirmations are inherently asynchronous (it's only a real toggle on the in-memory producer, where the awaited path also defers sends to the pump).

Verified: full `Paramore.Brighter.Core.Tests` (855 passed) and `Paramore.Brighter.InMemory.Tests` (144 passed) suites are green, including new tests that a broker-style producer's confirmation raise awaits `MarkDispatchedAsync` and that the mediator doesn't double-subscribe. The RMQ/Kafka test projects compile; their integration suites need broker infrastructure and weren't run here.

thomhurst:
Addressing the automated review's asks:

**(a) Test suites were run** — the "not executed" note was stale PR text from before the broker rollout; the description is now updated. Current state of this branch: `Paramore.Brighter.Core.Tests` 855 passed / 0 failed (7 pre-existing skips) and `Paramore.Brighter.InMemory.Tests` 144 passed / 0 failed, both including the new concurrency tests (gated-outbox dispose drain, broker-style awaited raise, span-teardown isolation). The RMQ/Kafka integration suites need broker infrastructure and weren't run here.

**(b) Kafka shutdown timeout** — kept fixed at 5s, now documented in code: `KafkaPublication` has no confirm-wait setting (unlike `RmqPublication.WaitForConfirmsTimeOutInMilliseconds`) and none of its existing timeouts describe this wait, so there was nothing natural to derive it from; 5s mirrors the RMQ.Async producer's active-sends shutdown bound. Happy to add a publication property if you'd rather have it tunable.

**(c) RMQ.Async ack-loop backpressure** — confirmed intentional. Callbacks for a multiple-ack batch run concurrently (`Task.WhenAll`), so the loop is bounded by the slowest write in a batch rather than the sum; the blast radius is that producer's own channel; and a producer using publisher confirms has already opted into confirmation-paced semantics (`WaitForConfirms*`). If the Outbox stalls badly the ack loop slows, but nothing is lost — un-marked messages are re-swept.

On sequential `InvokeAllAsync` (point 3): deliberate and documented on the helper — the mediator is the only subscriber today, and sequential is the conservative default for ordering.

thomhurst:
Both blockers addressed in bf4189ef1:

**1. In-memory drain fault isolation** — `DrainAsync` now routes the awaited raise through `RaiseConfirmationCallbacksAsync`, which wraps `InvokeAllAsync` in try/catch and logs `ConfirmationCallbackFault`, matching the broker producers. A throwing subscriber can no longer fault the drain worker, strand queued items, or re-surface from `DisposeAsync`. Regression test added: `When_an_async_confirmation_subscriber_throws_dispose_still_drains` (first callback throws; second message still confirmed; dispose completes).

**2. `InFlightCallbackTracker` unit tests** — added `InFlightCallbackTrackerTests` covering the four edges listed: immediate return when nothing in flight, blocking until the last `End()`, timeout returning false with an accurate remaining count, and reuse across drain cycles.

**3. Ack-loop backpressure** — added a "Behavioural note (for release notes)" section to the PR description calling out that RMQ.Async ack processing is now paced by the outbox write.

**4. Tracker shutdown race** — documented on `TryWait`: a `Begin` racing in after the wait observed zero is untracked; callers are expected to have stopped producing callbacks before waiting.

Suites re-run after the changes: `Paramore.Brighter.Core.Tests` 860 passed / 0 failed (7 pre-existing skips), `Paramore.Brighter.InMemory.Tests` 144 passed / 0 failed.

iancooper:
@thomhurst This looks good; I just need to review, but it shouldn't take too long

iancooper:
@thomhurst Not a blocker, but you can change this to be you, as our model is that you retain the copyright but license it back to us (in perpetuity under our existing MIT license), which means we can't rug-pull your work commercially.

>Copyright © 2026 Ian Cooper <ian_hammond_cooper@yahoo.co.uk>

thomhurst:
Thanks @iancooper — updated the copyright on the new files in 111581272.

iancooper (approved):
Sorry @thomhurst my clear day got busy. But this is great

iancooper:
Feel free to merge @thomhurst

PR #4251, opened 2026-07-20
fix(#4250): match ASB scheduler SessionId/reserved-header bag lookups case-insensitively

### LINKED ISSUE #4250: Session-affine messages lose their SessionId when sent through a relational outbox to Azure Service Bus (header-bag keys camelCased on persist, looked up PascalCase)
> 📖 **Before filing a bug**, please check the [documentation](https://brightercommand.gitbook.io/paramore-brighter-documentation/) to see if your issue is covered.

### Describe the bug
When a message carrying a `SessionId` header-bag entry is sent via the **relational outbox** (confirmed with the MsSql outbox) to **Azure Service Bus**, the native `ServiceBusMessage.SessionId` is never set, and the message is rejected/dead-lettered by session-required (or partitioned) entities with *"Session enabled entity doesn't allow a message whose session identifier is null."*

Root cause is a casing mismatch between three internal Brighter components that all handle `MessageHeader.Bag` differently:

1. A producer/mapper writes a bag entry with a **PascalCase** key: `Bag["SessionId"] = "<key>"`.
2. `RelationDatabaseOutbox` serialises `message.Header.Bag` with `JsonSerialisationOptions.Options`, whose `PropertyNamingPolicy = JsonNamingPolicy.CamelCase`. The custom converter, `DictionaryStringObjectJsonConverter.Write`, applies that policy to the dictionary **keys** (not just POCO property names), so `"SessionId"` is persisted to the `HeaderBag` column as `"sessionId"`.
3. On read, `DictionaryStringObjectJsonConverter.Read` restores the key **verbatim** (no reverse mapping) into a **case-sensitive** `Dictionary<string, object>` (`MessageHeader.Bag`'s type), so the rehydrated key is `"sessionId"`.
4. `AzureServiceBusMessagePublisher.AddBrighterHeaders` looks the value up case-sensitively via `ASBConstants.SessionIdKey`, which is the literal `"SessionId"` (PascalCase). This lookup misses `"sessionId"`, so `ServiceBusMessage.SessionId` is never set.

This means **no** header-bag key set with a non-camelCase name can survive an outbox round-trip and still be found by a case-sensitive consumer, because `JsonNamingPolicy.CamelCase` always lower-cases the leading character(s).

As a side effect, the un-matched `"sessionId"` key also leaks into `ApplicationProperties`: `AzureServiceBusMessagePublisher`'s reserved-header filter (`ASBConstants.ReservedHeaders.Contains(h.Key)`) is also case-sensitive PascalCase, so it no longer recognises the lower-cased key as reserved. The lower-cased key showing up in application properties is the diagnostic fingerprint for this bug.

The **in-process/direct** send path (no outbox round-trip) is unaffected, since the bag is never JSON-serialised in that path.

The identical defect is independently duplicated in `Paramore.Brighter.MessageScheduler.Azure`'s `AzureServiceBusScheduler`, which declares its own `SessionIdKey = "SessionId"` constant and the same case-sensitive lookup/reserved-header pattern.

We also found the same *class* of defect affecting other transports that use a PascalCase-leading bag key, e.g. `Paramore.Brighter.MessagingGateway.RocketMQ`'s `HeaderNames.Keys = "Brighter-RocketMQ-Keys"` and `HeaderNames.Tag = "Brighter-RocketMQ-Tag"`, both looked up case-sensitively in `RocketMqMessageProducer`. By contrast, `Paramore.Brighter.MessagingGateway.AWSSQS`'s `HeaderNames.DeduplicationId = "messageDeduplicationId"` is already camelCase and survives the round-trip — consistent with the `JsonSerialisationOptions` doc comment's own advice to use camelCase bag keys, which isn't followed by the ASB or RocketMQ constants.

We reviewed the source in this repo directly (types and line references below are from the current `main`, commit `ecb6ba11b1b0d6d05a175d022ac59bb3ac92cace`, not a decompile) and can confirm every step above:

- `src/Paramore.Brighter/JsonConverters/JsonSerialisationOptions.cs` — `PropertyNamingPolicy = JsonNamingPolicy.CamelCase` (its own doc comment already flags the camelCasing behaviour as a footgun for bag keys).
- `src/Paramore.Brighter/Serialization/DictionaryStringObjectJsonConverter.cs` — `Write` applies `options.PropertyNamingPolicy?.ConvertName(propertyName)` to the key; `Read` does not reverse it and constructs a default (ordinal, case-sensitive) `Dictionary<string, object?>`.
- `src/Paramore.Brighter/MessageHeader.cs` — `Bag` is `Dictionary<string, object>` (case-sensitive).
- `src/Paramore.Brighter/RelationDatabaseOutbox.cs` — both the persist path and `GetContextBag`/rehydrate path use `JsonSerialisationOptions.Options` for the `HeaderBag` column.
- `src/Paramore.Brighter.MessagingGateway.AzureServiceBus/ASBConstants.cs` — `SessionIdKey = "SessionId"`.
- `src/Paramore.Brighter.MessagingGateway.AzureServiceBus/AzureServiceBusMessagePublisher.cs` — case-sensitive `Bag.TryGetValue(ASBConstants.SessionIdKey, ...)` and case-sensitive `ReservedHeaders.Contains(h.Key)`.
- `src/Paramore.Brighter.MessageScheduler.Azure/AzureServiceBusScheduler.cs` — independent duplicate of the same pattern.

No test in `tests/` exercises this round-trip (a mapper-only or in-process test passes and misses the bug — the outbox round-trip has to be exercised for it to surface).

### Proposed solution
To keep the fix's footprint small and avoid touching the shared JSON converter (which all transports and the outbox rely on) or widening `MessageHeader.Bag` to a case-insensitive dictionary (a behavioural change with a much larger blast radius), we'd propose scoping the fix to the affected Azure Service Bus call sites only:

1. In `AzureServiceBusMessagePublisher.AddBrighterHeaders`, replace the case-sensitive `Bag.TryGetValue(ASBConstants.SessionIdKey, ...)` with a case-insensitive lookup (e.g. a linear scan with `StringComparison.OrdinalIgnoreCase`, since `Bag` is small and stays a plain `Dictionary<string, object>`).
2. In the same method, change the reserved-header leak filter to `ASBConstants.ReservedHeaders.Contains(h.Key, StringComparer.OrdinalIgnoreCase)` so a lower-cased `sessionId` key is still excluded from `ApplicationProperties` after an outbox round-trip.
3. Apply the identical two changes to `AzureServiceBusScheduler` in `Paramore.Brighter.MessageScheduler.Azure`, which has its own independent copy of the same bug.

This corrects the reported defect without changing `JsonSerialisationOptions`, `DictionaryStringObjectJsonConverter`, or `MessageHeader.Bag`, so it doesn't affect any other transport, the outbox itself, or any code that may currently depend on the existing camelCasing behaviour. We're not proposing a fix for the RocketMQ instance of the same underlying class of bug in this issue — flagging it here for awareness/triage, since it may warrant a separate issue and fix scoped the same way.

Happy to raise the PR for the three changes above once the approach is agreed.

### To Reproduce
1. Configure a Brighter producer with the **MsSql outbox** (or any relational outbox) and the Azure Service Bus gateway, targeting a **session-required** ASB topic/queue.
2. Use a message mapper that sets `header.Bag["SessionId"] = "some-session-key"`.
3. Call `DepositPostAsync(request)` and let the sweeper (`ClearOutboxAsync`) dispatch it — do not post in-process.
4. Observe: Azure Service Bus rejects/dead-letters the message with a null session id. Inspecting the persisted outbox row shows the `HeaderBag` column contains `"sessionId"` (lower-cased), and the constructed `ServiceBusMessage.SessionId` is `null`.

Contrast: posting the same message **in-process** (bypassing the outbox) sets `SessionId` correctly, since the bag is never JSON round-tripped in that path.

### Exceptions (if any)
No .NET exception is thrown inside Brighter. The failure surfaces from Azure Service Bus itself when the message reaches a session-required entity:

```
Microsoft.Azure.ServiceBus... / Azure.Messaging.ServiceBus:
Session enabled entity doesn't allow a message whose session identifier is null.
```

### Further technical details
- Brighter version: confirmed against published NuGet packages 10.5.1 and 10.6.0; also confirmed present in this repo's `main` branch at commit `ecb6ba11b1b0d6d05a175d022ac59bb3ac92cace` (2026-05-18) via direct source review.
- `dotnet --info` (relevant subset):
  ```
  .NET SDK: 10.0.203
  Host: 10.0.10 (x64)
  .NET runtimes installed: Microsoft.NETCore.App 6.0.36, 7.0.20, 8.0.26, 8.0.29, 9.0.15, 10.0.7, 10.0.10
  ```
- OS: Windows 11 (10.0.26200)

MikeFreeman-Flagstone:
## Summary

Fixes #4250.

**Scope note:** #4250 originally reported this bug against both `AzureServiceBusMessagePublisher` and `AzureServiceBusScheduler`. Since filing it, I found that #4205 (merged, closing #4054) already fixed the publisher half with an `ASBConstants.IsBagKey`/`IsReservedHeader` helper, and #4215 (merged, closing #4151/#4214) fixed the underlying root cause (bag keys no longer get camelCased on write at all). Neither PR touched `AzureServiceBusScheduler`, which carries its own independent copy of `ASBConstants` and had the identical bug. This PR is now scoped to just that remaining gap.

`AzureServiceBusScheduler.ConvertToServiceBusMessage` looked up `Bag["SessionId"]` and filtered `ReservedHeaders` case-sensitively in PascalCase. An outbox-sourced message whose bag key comes back in a different casing (e.g. from an older camelCasing round-trip, or any other naming policy) would silently lose its `ServiceBusMessage.SessionId`, and the mismatched key would leak into `ApplicationProperties` instead of being recognised as reserved.

## Fix

Ports the exact `IsBagKey`/`IsReservedHeader` pattern #4205 introduced into `Paramore.Brighter.MessagingGateway.AzureServiceBus.ASBConstants`, into the scheduler's own nested `ASBConstants`, for consistency with the already-reviewed and merged approach rather than reintroducing a different one.

## Test plan

- [x] Scheduler project builds clean.
- [x] Confirmed against upstream `master`: upstream's existing `When_Converting_A_Message_With_A_Session_Id.cs` (added by #4205) still passes unaffected — this PR doesn't touch the publisher.
- [ ] No dedicated test for the scheduler itself: there is currently no test project for `Paramore.Brighter.MessageScheduler.Azure` in this repo. Applied by parity with the reviewed and tested publisher fix. Happy to add a test project as a follow-up if maintainers want that separated out.

### DIFF
--- a/src/Paramore.Brighter.MessageScheduler.Azure/AzureServiceBusScheduler.cs
+++ b/src/Paramore.Brighter.MessageScheduler.Azure/AzureServiceBusScheduler.cs
@@ -132,7 +132,7 @@ public class AzureServiceBusScheduler(
         azureServiceBusMessage.ApplicationProperties.Add(ASBConstants.ReplyToHeaderBagKey, message.Header.ReplyTo);
 
         foreach (var header in message.Header.Bag.Where(h =>
-                     !ASBConstants.ReservedHeaders.Contains(h.Key)
+                     !ASBConstants.IsReservedHeader(h.Key)
                      && !MessageHeader.IsLocalHeader(h.Key)))
         {
             azureServiceBusMessage.ApplicationProperties.Add(header.Key, header.Value);
@@ -144,11 +144,14 @@ public class AzureServiceBusScheduler(
         }
 
         var contentType = message.Header.ContentType ?? new ContentType(MediaTypeNames.Text.Plain);
-        
+
         azureServiceBusMessage.ContentType = contentType.ToString();
         azureServiceBusMessage.MessageId = message.Header.MessageId;
-        if (message.Header.Bag.TryGetValue(ASBConstants.SessionIdKey, out object? value))
-            azureServiceBusMessage.SessionId = value.ToString();
+        var sessionId = message.Header.Bag
+            .FirstOrDefault(h => ASBConstants.IsBagKey(h.Key, ASBConstants.SessionIdKey))
+            .Value;
+        if (sessionId is not null)
+            azureServiceBusMessage.SessionId = sessionId.ToString();
 
         return azureServiceBusMessage;
     }
@@ -167,6 +170,30 @@ public class AzureServiceBusScheduler(
                 LockTokenHeaderBagKey, MessageTypeHeaderBagKey, HandledCountHeaderBagKey, ReplyToHeaderBagKey,
                 SessionIdKey
             };
+
+        /// <summary>
+        /// Does a header bag <paramref name="key"/> refer to the reserved key <paramref name="reservedKey"/>?
+        /// Brighter's Outbox serializes the bag with <see cref="JsonSerialisationOptions.Options"/>, whose
+        /// <see cref="System.Text.Json.JsonSerializerOptions.PropertyNamingPolicy"/> rewrites bag keys — so a key
+        /// written as "SessionId" comes back transformed (for example "sessionId" or "session_id") depending on the
+        /// configured policy. We therefore match both the raw key and the key as the active naming policy would
+        /// render it, case-insensitively, so the lookup stays correct whatever policy the user configures.
+        /// </summary>
+        public static bool IsBagKey(string key, string reservedKey)
+        {
+            if (string.Equals(key, reservedKey, StringComparison.OrdinalIgnoreCase))
+                return true;
+
+            var namingPolicy = JsonSerialisationOptions.Options.PropertyNamingPolicy;
+            return namingPolicy is not null
+                   && string.Equals(key, namingPolicy.ConvertName(reservedKey), StringComparison.OrdinalIgnoreCase);
+        }
+
+        /// <summary>
+        /// Is a header bag <paramref name="key"/> one of the <see cref="ReservedHeaders"/>, accounting for any
+        /// naming-policy transformation applied during an Outbox round-trip? See <see cref="IsBagKey"/>.
+        /// </summary>
+        public static bool IsReservedHeader(string key) => ReservedHeaders.Any(reserved => IsBagKey(key, reserved));
     }
 
     /// <inheritdoc />


iancooper:
It might be worth understanding this fix first: https://github.com/BrighterCommand/Brighter/pull/4215

(Still awaiting a release, I think)

This removes the conversion that was being applied to bag key names as unwarranted, and leaves them as written. So I think it might also fix this, but generically.

MikeFreeman-Flagstone:
> It might be worth understanding this fix first: https://github.com/BrighterCommand/Brighter/pull/4215

Ah, yes, looks like it solves in on the publisher side, just verifying if its on scheduler side too.

MikeFreeman-Flagstone:
Does not make sense to resolve sessionIds on scheduled messages as the the point of sessions is for ordering. Closing PR.

Issue #4218, opened 2026-07-04
Make `Partitioner` Required on `KafkaPublication` in v11 and Default Guidance to Murmur2


lillo42:
### Problem Statement
`KafkaPublication` already exposes a `Partitioner` property, but it **defaults to `ConsistentRandom`** (CRC32-based). Most users never override it, so they silently get librdkafka's CRC32 partitioner without making an explicit choice.

`ConsistentRandom` has weaker distribution properties than `Murmur2` or `Murmur2Random`. Under certain key patterns (sequential IDs, similar prefixes), CRC32 can produce clustering and uneven partition load. Murmur2 provides better statistical spread across partitions.

### Current State
```csharp
public enum Partitioner
{
    Random,
    Consistent,
    ConsistentRandom,
    Murmur2,
    Murmur2Random,
}

public class KafkaPublication : Publication
{
    public Partitioner Partitioner { get; set; } = Partitioner.ConsistentRandom;  // silent default
    // ...
}
```

### Proposed Change for v11
Remove the default value and make `Partitioner` **required**, forcing the caller to make an explicit choice.

```csharp
public class KafkaPublication : Publication
{
    public required Partitioner Partitioner { get; set; }
    // ...
}
```

### Recommended Guidance
Update all documentation and samples to recommend **`Partitioner.Murmur2Random`** (or `Murmur2` if null keys are not expected). This provides better distribution across partitions compared to `ConsistentRandom`.

```csharp
new KafkaPublication
{
    Topic = new RoutingKey("my.topic"),
    Partitioner = Partitioner.Murmur2Random,  // explicit, better distribution
    // ...
}
```

### Why Murmur2?
- **Better statistical distribution**: Murmur2's avalanche properties spread keys more uniformly than CRC32, reducing hot partitions under common key patterns.
- **Ecosystem alignment**: Most Kafka clients (Java, librdkafka with explicit config) treat murmur2 as the standard for key-based partitioning.

### Migration Path for v11
1. **v11.0.0**:
   - Change `Partitioner` from `public Partitioner Partitioner { get; set; } = Partitioner.ConsistentRandom;` to `public required Partitioner Partitioner { get; set; }`.
   - Update all Kafka samples to show explicit `Partitioner = Partitioner.Murmur2Random`.
   - Update documentation to explain the choice and recommend murmur2.
2. **Migration guide**:
   - Every `KafkaPublication` instantiation must add `Partitioner = ...`.
[...]

PR #4255, opened 2026-07-24
Add Kafka publication partitioner analyzer and code fixes

lillo42:
## Description

Adds a Kafka publication partitioner analyzer with three new diagnostics:

- **BRT006** — warns when a `KafkaPublication` is created without an explicit `Partitioner` assignment.
- **BRT007** — warns when `Partitioner.ConsistentRandom` is used and recommends `Partitioner.Murmur2Random` for new publications.
- **BRT008** — warns when `Partitioner.Consistent` is used and recommends `Partitioner.Murmur2` for new publications.

This also introduces a new `Paramore.Brighter.Analyzer.CodeFixes` project and ships it in the analyzer NuGet package. The code fixes can add the recommended `Partitioner.Murmur2Random` assignment or replace legacy `ConsistentRandom`/`Consistent` values with the recommended Murmur2 alternatives.

Documentation for BRT006–BRT008 is included, together with analyzer and code-fix tests.

## Related Issues

- Related to #4218

## Type of Change

- [ ] Documentation update

## Checklist

- [ ] I have read the [Contributing Guide](../blob/master/CONTRIBUTING.md)
- [ ] My changes follow the existing code style and conventions

## Additional Notes

Existing publications that intentionally rely on `ConsistentRandom` or `Consistent` can keep their current partition assignment and suppress the warning if preserving the current key-to-partition mapping is required.

### Maintainer decisions

Two points raised in review are deliberate decisions, not oversights:

- **BRT006 ships at `Warning`, enabled by default.** Every `KafkaPublication` created without an explicit `Partitioner` is flagged, so consumers building with `TreatWarningsAsErrors` (this repo included) will see build failures on upgrade until they either set the partitioner explicitly or suppress the rule. This is intentional: omitting the assignment silently selects the `ConsistentRandom` default — the same value BRT007 discourages — so it is reported at the same severity (the argument is written up in `BRT006.md`).
- **The code fixes change runtime behaviour, and Fix All is deliberately not offered.** Applying a fix moves a publication to a Murmur2 partitioner, which re-partitions the topic — keys map to different partitions than before. Because of that, the fixes can only be applied one diagnostic at a time: each change must be an explicit, per-site decision by the developer, not a solution-wide action. Each diagnostic is individually suppressible for publications that must keep their current assignment, and the caveat is stated in the BRT006–BRT008 docs and in the code action title.


lillo42:
**`ConstructorAssignsPartitioner` over-matching (finding 1)**: accepted as a trade-off, and it's the cheap direction to be wrong. The constructor check is intentionally syntactic (analyzers must not call `Compilation.GetSemanticModel`  RS1030), so it's a heuristic by design. Its job is to prevent the expensive failure: `BRT006` firing on every instantiation of a subclass that configures itself in its constructor, with a Fix All then injecting initializers that silently override the subclass's own choice at every call site. The false negative you describe, a constructor that only configures a different publication inside an object initializer, is contrived by comparison, self-correcting (set the partitioner explicitly or suppress), and doesn't cascade through Fix All. If it shows up in real code we can tighten the scan to skip assignments inside `InitializerExpressionSyntax` ancestors in a follow-up; I don't want to grow the heuristic further without evidence.

**Descriptor wording vs BRT007.md (finding 4)**: this one is a deliberate maintainer decision, not an oversight. The rule's stated position is that the `Murmur2` family is preferred for key-distribution reasons, and the descriptor text and sample comments reflect that. Not changing it in this PR.

**Two-layout fix output (finding 6, second half)**:  the empty argument list is now dropped (`new KafkaPublication { … }`), which was the arbitrary-looking part. Single- vs multi-line layout is left to the formatter; the tests pin the current shapes.

PR #4263, opened 2026-08-03
Update RocketMQ.Client to 5.2.1 and fix RocketMQ test suite

lillo42:
## Description

Updates the RocketMQ client (`RocketMQ.Client` 5.1.0 → 5.2.1) and fixes the RocketMQ test suite, which had been silently broken (the RocketMQ CI job is currently disabled, so regressions went unnoticed).

Client upgrade (`772490a7a`):

- Bump `RocketMQ.Client` to 5.2.1 and migrate `RocketMessageConsumer` to its APIs — real `Nack` via `ChangeInvisibleDuration` instead of a no-op, and primary-constructor parameter capture in place of explicit fields.

Test suite fixes (`4990be010`):

- **Producer: empty Baggage property rejected.** `RocketMqMessageProducer` unconditionally added the `CE_baggage` property; an empty `Baggage` (the default) stringifies to `""`, which the client's `Message.Builder.AddProperty` rejects (`value should not be null or white space`, validation present since client 5.1.0). Every `Send`/`SendAsync` threw — this alone accounted for 24 of 35 test failures. The property is now only added when the baggage has entries.
- **Topic map typos.** `RocketMqMessageGatewayProvider.s_topicMap` used `When_requeing_...` keys vs. the actual `When_requeuing_...` test method names, so four requeue tests fell through to non-existent topics and died with `NotFoundException`.
- **`RocketConsumerFactoryDlqTests` rework.** The test could never pass: it built a subscription without a consumer group (which defaulted to `string.Empty` and fails the client's `SetConsumerGroup` regex), and it asserted via reflection on private fields removed in the client-upgrade commit. It now passes a random consumer group and asserts on the new public `RocketMessageConsumer.DeadLetterRoutingKey` / `InvalidMessageRoutingKey` properties instead of implementation details.
- **`docker-compose-rocketmq.yaml`:** image pinned (`apache/rocketmq:5.4.0`, matching the hardcoded mqadmin paths), proxy heap raised to `-Xmx512m`, `orders`/`orders-dlq`/`orders-invalid` topics added, and misleading port comments corrected (8081 is the gRPC endpoint clients use).

## Type of Change

- [ ] Documentation update

## Checklist

## Additional Notes

- Verified against a local RocketMQ broker (topics provisioned by the compose file with correct `message.type` attributes): all code-level failures are resolved, including the previously broken DLQ/reject and consumer-factory tests.
- Remaining suite instability is environmental, not Brighter code: `ReceiveMessage` long-polls through the local podman-forwarded proxy intermittently hang until the client deadline (`HTTP/2 CANCEL` → `DeadlineExceeded`). Reproduced standalone with the raw 5.2.1 client (no Brighter code) on both broker 5.4.0 and 5.5.0 — likely the rootless port-forwarding breaking long-lived gRPC streams, or a proxy-side long-poll issue (cf. apache/rocketmq#10615, fixed in the upcoming 5.5.1). Tests that receive timely broker responses all pass.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -138,7 +138,7 @@
     <PackageVersion Include="Quartz.Extensions.DependencyInjection" Version="3.18.1" />
     <PackageVersion Include="Quartz.Extensions.Hosting" Version="3.18.1" />
     <PackageVersion Include="RabbitMQ.Client" Version="7.2.1" />
-    <PackageVersion Include="RocketMQ.Client" Version="5.1.0" />
+    <PackageVersion Include="RocketMQ.Client" Version="5.2.1" />
     <PackageVersion Include="Serilog" Version="4.3.1" />
     <PackageVersion Include="Serilog.Extensions.Hosting" Version="10.0.0" />
     <PackageVersion Include="Serilog.Extensions.Logging" Version="10.0.0" />
--- a/docker-compose-rocketmq.yaml
+++ b/docker-compose-rocketmq.yaml
@@ -3,7 +3,7 @@ version: '3.8'
 services:
   # RocketMQ Nameserver (Service Discovery)
   nameserver:
-    image: apache/rocketmq
+    image: apache/rocketmq:5.4.0
     platform: linux/amd64
     container_name: rocketmq-nameserver
     ports:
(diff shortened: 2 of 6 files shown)


iancooper:
Just to let you know, the generator conformance tests will probably help fix the RocketMQ tests locally, although perhaps not the server side issues, so you should pause for those, and then do this work.

See #4297

PR #4264, opened 2026-08-06
Kafka: add ErrorLogLevel hook to classify/suppress non-fatal error-callback log level

### LINKED ISSUE #4226: Kafka: allow applications to classify/suppress log level of non-fatal consumer error-callback events
## Is your feature request related to a problem?

The Kafka consumer's librdkafka error callback logs **every** non-fatal error unconditionally at `Warning`, with no way for an application to suppress or downgrade a specific error code.

In `KafkaMessageConsumer.cs` (`SetErrorHandler`, ~L253-261):

```csharp
.SetErrorHandler((_, error) =>
{
    _hasFatalError = error.IsFatal;

    if (_hasFatalError)
        Log.FatalError(s_logger, error.Code, error.Reason, true);
    else
        Log.NonFatalError(s_logger, error.Code, error.Reason, false); // Warning, always
})
```

The concrete case that motivated this: against **Confluent Cloud**, the broker/LB idle-connection reaper closes the mostly-idle group-coordinator socket. librdkafka only notices on its next request, which surfaces via the *error* callback (not the log callback, so `log.connection.close=false` does not help) as:

```
Code: Local_TimedOut, Reason: GroupCoordinator: <broker>:9092: 1 request(s) timed out: disconnect
(average rtt 7-28 ms) (after ... ms in state UP), Fatal: False
```

This is benign and self-healing — librdkafka reconnects automatically — but it floods logs/alerting at `Warning`. Today the only mitigations are app-side:

- **Tune the cause** via `KafkaSubscription.ConfigHook` (`SocketKeepaliveEnable`, `ConnectionsMaxIdleMs`) — reduces but does not eliminate it.
- **Filter in the app's logging pipeline** — but Brighter assigns no explicit `EventId` to `NonFatalError`, so a precise filter has to inspect the structured `ErrorCode` state, and a coarse filter (by category/level) blinds you to *all* non-fatal errors, including ones you care about (auth, transport, partition).

There is no first-class, transport-level way to say "this specific non-fatal code is expected noise — log it at Debug (or not at all)".

## Describe the solution you'd like

An optional **error-classification hook** on `KafkaSubscription` that lets the application map a librdkafka `Error` to the `LogLevel` Brighter should log it at (including `LogLevel.None` to suppress). Logging classification only — it must **not** influence whether the error is treated as fatal.

```csharp
/// <summary>
/// Optional. Classify the log level for librdkafka error-callback events, e.g. to downgrade
/// or suppress expected non-fatal noise such as Local_TimedOut on an idle coordinator socket.
/// Return LogLevel.None to suppress. Does NOT affect fatal handling (Error.IsFatal is authoritative).
/// </summary>
public Func<Error, LogLevel>? ErrorLogLevel { get; set; }
```

Wired into the handler so the level is dynamic, while the fatal path is unchanged:

```csharp
.SetErrorHandler((_, error) =>
{
    _hasFatalError = error.IsFatal; // unchanged: IsFatal stays authoritative

    var level = _errorLogLevel?.Invoke(error)
                ?? (error.IsFatal ? LogLevel.Error : LogLevel.Warning); // preserves today's behaviour

    if (level != LogLevel.None)
        Log.KafkaError(s_logger, level, error.Code, error.Reason, error.IsFatal);
})
```

Application usage:

```csharp
var subscription = new KafkaSubscription<MyEvent>(
    // ...
    // downgrade the idle-coordinator timeout to Debug, leave everything else as-is
);
subscription.ErrorLogLevel = err =>
    err.Code == ErrorCode.Local_TimedOut ? LogLevel.Debug : LogLevel.Warning;
```

### Scope / guardrails
- **Logging classification only.** The hook cannot upgrade a non-fatal error to fatal or vice-versa; `Error.IsFatal` remains the sole driver of `_hasFatalError` and the resulting `ChannelFailureException`.
- **Default is no behaviour change.** When the hook is null, current levels (`Error` for fatal, `Warning` for non-fatal) are preserved exactly.
- Follows the existing `ConfigHook` precedent — an opt-in `Action`/`Func` on `KafkaSubscription` for things Brighter doesn't model directly.

## Describe alternatives considered

- **`ConfigHook` tuning only** (`SocketKeepaliveEnable`, `ConnectionsMaxIdleMs`) — good practice and recommended regardless, but reduces rather than eliminates the noise and doesn't generalise to other expected-but-noisy codes.
- **App-side log filtering** — works, but leaks a transport concern into every consuming app's logging config, and precise targeting is awkward because `NonFatalError` carries no explicit `EventId`.
- **A simple `IEnumerable<ErrorCode> SuppressedErrorCodes`** — simpler, but less expressive than a level map (can only drop, not downgrade). The `Func<Error, LogLevel>` supersets it.

## Additional context

Related but out of scope: `_hasFatalError = error.IsFatal;` **overwrites** the flag on each callback, so a fatal error immediately followed by a non-fatal one (librdkafka commonly emits bursts) could reset it to `false` before the consume loop checks it (`KafkaMessageConsumer.cs` ~L494). Worth a separate issue.

Happy to submit a PR (new property + handler wiring + a `KafkaError` dynamic-level `LoggerMessage` + tests) if the direction is agreeable.

IslamZakaria:
## Description
Adds `KafkaSubscription.ErrorLogLevel` (`Func<Error, LogLevel>?`), wired through `KafkaMessageConsumerFactory` into `KafkaMessageConsumer.HandleError`, so applications can classify or suppress the log level of non-fatal librdkafka error-callback events — e.g. downgrading `Local_TimedOut` idle-coordinator noise from Confluent Cloud instead of it flooding logs at `Warning`.

- `LogLevel.None` suppresses the log entirely.
- The hook never influences fatal handling: `_hasFatalError` remains latched purely from `Error.IsFatal`, regardless of what the hook returns.
- When the hook is null, current behavior is preserved exactly (fatal → `Error`, non-fatal → `Warning`).
- Replaces the fixed-level private `FatalError`/`NonFatalError` `LoggerMessage`s with a single dynamic-level `Log.KafkaError` via `ILogger.Log`, since `[LoggerMessage]` can't express a dynamic `Level`.

## Related Issues
Closes #4226

## Type of Change
- [ ] Documentation update

## Checklist

## Additional Notes
- **Shared test-infra change**: `tests/.../Initializer.cs` now uses `MinimumLevel.Debug()` for the TestCorrelator Serilog logger, so Debug-level log assertions are capturable (needed for the suppression/downgrade tests). Verified against a clean-`master` baseline: full Kafka suite (114 tests) has a byte-identical failure set before/after (76 pre-existing, broker-dependent failures), so this introduces no regressions.
- **Pre-existing failure, not introduced here**: `ConsumerConfigHookTests` fails on clean `master` too — it requires a live Kafka broker for `makeChannels: Create` → topic metadata lookup.
- `Log.FatalError`/`NonFatalError` were `private` (nested in a `private static partial class`); their only callers were `HandleError`, so removing them is **not** a public API break. No `PublicAPI.Shipped.txt`/`PublicApiAnalyzers` exist in this repo to update.
- **Binary compatibility note**: the new `errorLogLevel`/`ErrorLogLevel` constructor parameters are appended last with a default value (source-compatible, matches the existing pattern used for `TimeProvider`/`scheduler`/`groupProtocol`). This is source-compatible but, like any optional-parameter addition, is not binary-compatible for callers pre-compiled against the old signature without recompiling — inherent to this approach and consistent with how the constructors have grown previously.
- **Minor logging delta**: the new `Log.KafkaError` call has no `EventId` (defaults to `EventId(0)`), whereas the replaced `FatalError`/`NonFatalError` had generator-assigned opaque IDs. These were never documented or stable (the issue itself notes Brighter assigns no explicit EventId to `NonFatalError`), but flagging in case any consumer was keying off them.

### DIFF
--- a/src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageConsumer.cs
+++ b/src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageConsumer.cs
@@ -62,6 +62,7 @@ namespace Paramore.Brighter.MessagingGateway.Kafka
         private static readonly TimeSpan s_commitSyncTimeout = TimeSpan.FromSeconds(5);
         private readonly ITimer _sweeperTimer;
         private bool _hasFatalError;
+        private readonly Func<Error, LogLevel>? _errorLogLevel;
         private bool _isClosed;
         private readonly RoutingKey? _deadLetterRoutingKey;
         private readonly RoutingKey? _invalidMessageRoutingKey;
@@ -104,6 +105,7 @@ namespace Paramore.Brighter.MessagingGateway.Kafka
         /// <param name="makeChannels">Should we create infrastructure (topics) where it does not exist or check. Defaults to Create</param>
         /// <param name="configHook">Allows you to modify the Kafka client configuration before a consumer is created.</param>
         /// <param name="timeProvider">The <see cref="TimeProvider"/> used to create the timer for sweeping uncommitted offsets. Defaults to <see cref="TimeProvider.System"/> if not specified. Can be overridden for testing purposes.</param>
+        /// <param name="errorLogLevel">Allows you to classify or suppress the log level of non-fatal errors raised by the underlying Kafka consumer. Returning <see cref="LogLevel.None"/> suppresses the log entirely. Does not affect fatal handling, which is driven solely by <see cref="Error.IsFatal"/>. Defaults to <see langword="null"/>, logging fatal errors at <see cref="LogLevel.Error"/> and non-fatal at <see cref="LogLevel.Warning"/>.</param>
         /// <param name="deadLetterRoutingKey">If we support a dead letter topic what is the <see cref="RoutingKey"/></param>
         /// <param name="invalidMessageRoutingKey">If we support an invalid message topic what is the <see cref="RoutingKey"/></param>
         /// <param name="scheduler">Optional scheduler for delayed requeue operations. When provided, the lazily-created
@@ -132,7 +134,8 @@ namespace Paramore.Brighter.MessagingGateway.Kafka
             RoutingKey? invalidMessageRoutingKey = null,
             TimeProvider? timeProvider = null,
             IAmAMessageScheduler? scheduler = null,
-            IGroupProtocol? groupProtocol = null)
+            IGroupProtocol? groupProtocol = null,
+            Func<Error, LogLevel>? errorLogLevel = null)
         {
             if (groupId is null)
                 throw new ConfigurationException("You must set a GroupId for the consumer");
@@ -141,6 +144,7 @@ namespace Paramore.Brighter.MessagingGateway.Kafka
             
             _configuration = configuration ?? throw new ConfigurationException("You must set a KafkaMessagingGatewayConfiguration to connect to a broker");
             _scheduler = scheduler;
+            _errorLogLevel = errorLogLevel;
 
             _deadLetterRoutingKey = deadLetterRoutingKey;
             _invalidMessageRoutingKey = invalidMessageRoutingKey;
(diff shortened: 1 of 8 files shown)


iancooper:
@IslamZakaria, can you sign the CLA? We let you keep the copyright to your code, you just grant us perpetual use, for this purpose/license

iancooper (approved):
Thanks for this, @IslamZakaria . Nice, concise work with respect to existing code. Just sign the CLA, and if we can get a green Core and Kafka build, we are done. Congrats

IslamZakaria:
@iancooper Thanks. I actually signed the CLA already. Could you please check if it has been picked up correctly? If there's an issue with my GitHub account or the email associated with the CLA, let me know and I'll be happy to resolve it.

iancooper:
Let me check. I think it was complaining that you signed with an email that was not associated with the GitHub account

IslamZakaria:
but i already signed in with github account seems weird

iancooper:
> but i already signed in with github account seems weird

Seems happy now, the workflows can run

iancooper:
Thanks @IslamZakaria merged

Issue #4276, opened 2026-08-25
Flaky: JustSaying converter registration races on the shared static JsonSerialisationOptions.Options, failing build at random on any PR


iancooper:
One test in `Paramore.Brighter.Transforms.Adaptors.Tests` fails intermittently, which means **any** pull request to this repository can go red for reasons unrelated to its diff. The failure moves between target frameworks between runs, so it reads like noise rather than a defect.

## Symptom

```
JustSayingPropertySerializationTests.When_mapping_justsaying_event_should_serialize_all_properties_as_flat_values

System.InvalidOperationException: This JsonSerializerOptions instance is read-only or has already been used
   at Paramore.Brighter.Transformers.JustSaying.JsonConverters.RegisterConverters.Register()
```

Observed on `net10.0` (2026-08-07) and on `net9.0` while `net10.0` passed (PR #4275). It also passed 30/30 on both TFMs on three consecutive builds of that PR before failing on the fourth, whose diff was comments and a README — so a green run is not evidence it is fixed.

## Cause: a check-then-act race the existing guard cannot close

`src/Paramore.Brighter.Transformers.JustSaying/JsonConverters/RegisterConverters.cs`:

```csharp
lock (s_lock)
{
    // double-locking pattern
    if (s_register) { return; }

    if (!JsonSerialisationOptions.Options.Converters.IsReadOnly)   // :25  <-- passes
    {
        JsonSerialisationOptions.Options.Converters.Add(new TenantConverter());   // :27  <-- throws
        JsonSerialisationOptions.Options.Converters.Add(new IpAddressConverter());
    }

    s_register = true;
}
```

The throw comes from the `Add` at `:27`, **inside** the `IsReadOnly` check at `:25`. So the guard passed and the `Add` still threw. That is the tell: the lock protects `Register()` from other callers of `Register()`, but nothing protects the shared static `JsonSerialisationOptions.Options` from being frozen by an unrelated thread in the window between the check and the `Add`. `System.Text.Json` freezes an options instance the first time anything serializes with it.

## Why it fires in this assembly specifically

`tests/Paramore.Brighter.Transforms.Adaptors.Tests` has **no `xunit.runner.json` and no `CollectionBehavior` attribute**, so xUnit runs its test classes in parallel by default. Two of those classes have nothing to do with JustSaying and serialize with the same shared instance:

- `MassTransit/MassTransitMessageMapperTest.cs` — 3 references to `JsonSerialisationOptions.Options`
[...]

Issue #4272, opened 2026-08-24
`TodoApi` sample is missing from `Brighter.slnx`, so CI has never built it


iancooper:
`samples/WebAPI/WebAPI_mTLS_TestHarness/TodoApi/TodoApi.csproj` is on disk but is not
referenced by `Brighter.slnx`, so it is not part of any CI build.

### Evidence

- `Brighter.slnx` names **109** sample `.csproj` files; there are **110** under `samples/`.
  `TodoApi` is the only one unaccounted for.
- `ci.yml` builds with `dotnet build --configuration Release` at the repository root and no
  project argument, which resolves to `Brighter.slnx`. A project absent from the solution is
  absent from the build.
- It arrived in c8d27b2d1 (*Feature/3902 rabbitmq mutual tls*, #3953, 2026-02-01) and
  `git log -S 'WebAPI_mTLS_TestHarness' -- Brighter.slnx` returns nothing, so it has never
  been referenced. That is roughly six months of a sample nothing compiles.

### It is only an oversight — the fix is one line

I built it directly to check whether something was keeping it out:

```
dotnet build samples/WebAPI/WebAPI_mTLS_TestHarness/TodoApi/TodoApi.csproj -c Release
→ Build succeeded. 0 Error(s)
```

Nothing structural is in the way. It targets `net10.0`, which 21 other samples already do and
which CI installs; its one `PackageReference` (`Swashbuckle.AspNetCore`) is already pinned in
`Directory.Packages.props`. So adding a `<Project Path="…" />` entry under the
`/samples/WebAPI/` folder is all it needs.

*(It does emit `NU1903` for a transitive `Microsoft.OpenApi` 2.4.1 advisory — but that is not
hidden by this, since `WebAPI_Dapper`, `WebAPI_Dynamo` and `WebAPI_EFCore` reference
Swashbuckle too and are all in the solution. Separate matter, mentioned only so it is not a
surprise when the build starts running.)*

---

*Found while writing tutorial samples for the docs, where the plan leans on "samples are kept
honest because CI compiles them". That is true, but conditional on the solution entry, which
is worth knowing.*

PR #4280, opened 2026-08-26
samples: add Tutorials/04-Kafka, the sample behind tutorial rung 4

iancooper:
Rung 4 of the *Get Started* tutorial ladder, and the last sample the ladder needs. It is
`samples/Tutorials/02-FirstMessage` with the transport swapped for Kafka: one event over a
three-partition topic, to a consumer group you can scale by starting a second copy of the
receiver with no configuration change.

The tutorial page that quotes this code (`TutorialStreamingWithKafka.md`) follows in the Docs
repository, per the same convention as rungs 2 and 3 — the sample merges first.

## One readable delta from rung 2

Three of the seven files are byte-identical to `02-FirstMessage`'s, asserted with `cmp`:

```text
Greetings/GreetingEvent.cs                  identical
Greetings/Greetings.csproj                  identical
GreetingsReceiver/GreetingEventHandler.cs   identical
GreetingsSender/Program.cs                  Kafka publication, NumPartitions = 3, partition key
GreetingsSender/GreetingsSender.csproj      Kafka gateway instead of RMQ
GreetingsReceiver/Program.cs                KafkaSubscription, groupId
GreetingsReceiver/GreetingsReceiver.csproj  Kafka gateway instead of RMQ
```

Rung 3 changes only the sender; this rung changes both ends, because a transport is something
two processes have to agree on.

## Run, not just compiled

Against `docker-compose-kafka.yaml` from a torn-down broker, with both preconditions asserted
first — zero topics, zero receiver processes:

- One instance is assigned all three partitions
  (`Partition Added greeting.event : 0,greeting.event : 1,greeting.event : 2`).
- Nine greetings arrive **grouped by recipient rather than in send order**. The sender
  interleaves three recipients round-robin; the receiver prints alice #1–#3, then grace #1–#3,
  then mia #1–#3. Per-partition ordering is visible rather than argued about.
- A second instance triggers a rebalance. The first logs
  `Partitions for consumer revoked … [0],[1],[2]` and keeps a disjoint subset; the second takes
  the rest; together they cover all three.
- **The rebalance redelivers exactly three greetings, one per partition.** `commitBatchSize`
  defaults to 10 and each partition holds 3 messages, so the batch is never reached and the last
  message on each partition is uncommitted when partitions are revoked (`CURRENT-OFFSET 2`,
  `LOG-END-OFFSET 3`, `LAG 1` on all three). Nothing is lost; three are handled twice. That is
  at-least-once, and the README says so.

Both the rebalance and the redelivery count were observed on two clean runs, not one.

## Three things worth a reviewer's eye

**No message mapper.** `KafkaTaskQueue` writes one in order to set the partition key. It does
not have to: `JsonMessageMapper<T>` is the registered default for *both* the sync and the async
path (`ServiceCollectionMessageMapperRegistryBuilder`), and it already reads the key from the
request context onto the header (`JsonMessageMapper.cs:50`). So the sender sets
`RequestContextBagNames.PartitionKey` on a `RequestContext` and passes it to `Post`, and the
sample writes no mapper — which keeps it the same shape as rung 2.

**The recipient names are chosen, and the reason is in the code comment.** Kafka hashes the key
to pick a partition. The first three names tried — `alice`, `bob`, `carol` — all hashed to
partition 2. A single consumer holds every partition anyway, so all nine greetings arrived in
order and the sample looked perfect while demonstrating nothing; it took
`kafka-console-consumer.sh --property print.partition=true` to see it. `alice`, `grace` and
`mia` were checked against the broker and land on 2, 0 and 1.

**Reactor, deliberately.** `KafkaSubscription` already defaults to `MessagePumpType.Reactor`, so
the handler is a plain synchronous `RequestHandler<GreetingEvent>` — byte-identical to rung 2's
— and the single-threaded pump that makes per-key ordering hold is visible in the code.

## Solution registration

All three projects are in `Brighter.slnx` under `/samples/Tutorials/04-Kafka/`. Checked against
`git ls-files` rather than `find`:

```text
tracked .csproj: 252   named in Brighter.slnx: 251
in the repo but NOT in the solution:
    samples/WebAPI/WebAPI_mTLS_TestHarness/TodoApi/TodoApi.csproj
in the solution but NOT in the repo:
    (none)
```

The single unregistered project is the pre-existing #4272 case, untouched here — it is the
control for this check, not a finding.

## Note on CI

`build` is flaky on `master` independently of this branch — #4276, a shared-static race in
`RegisterConverters.Register()` that flips both TFM and test method between runs. If `build`
goes red, please check whether it is that before suspecting this sample.

### DIFF
--- a/Brighter.slnx
+++ b/Brighter.slnx
@@ -166,6 +166,12 @@
     <Project Path="samples/Tutorials/03-DurableOutbox/GreetingsReceiver/GreetingsReceiver.csproj" />
     <Project Path="samples/Tutorials/03-DurableOutbox/GreetingsSender/GreetingsSender.csproj" />
   </Folder>
+  <Folder Name="/samples/Tutorials/04-Kafka/">
+    <File Path="samples/Tutorials/04-Kafka/README.md" />
+    <Project Path="samples/Tutorials/04-Kafka/Greetings/Greetings.csproj" />
+    <Project Path="samples/Tutorials/04-Kafka/GreetingsReceiver/GreetingsReceiver.csproj" />
+    <Project Path="samples/Tutorials/04-Kafka/GreetingsSender/GreetingsSender.csproj" />
+  </Folder>
   <Folder Name="/samples/Validation/">
     <File Path="samples\Validation\README.md" />
     <Project Path="samples/Validation/DataAnnotationsSample/DataAnnotationsSample.csproj" />
--- a/samples/Tutorials/02-FirstMessage/README.md
+++ b/samples/Tutorials/02-FirstMessage/README.md
@@ -70,6 +70,6 @@ those is a concept a reader meeting a broker for the first time has to park, so 
 drops all four and adds nothing. It also has no message mapper: `JsonMessageMapper<T>` is
 Brighter's default, and a tutorial should not write code the framework already supplies.
 
-Rung 3 adds a durable Outbox to this sample; rung 4 swaps the transport for Kafka. Each is
-one readable delta from this one, which is why this is deliberately the smallest thing that
-still crosses a broker.
+[Rung 3](../03-DurableOutbox) adds a durable Outbox to this sample; [rung 4](../04-Kafka)
+swaps the transport for Kafka. Each is one readable delta from this one, which is why this is
+deliberately the smallest thing that still crosses a broker.
--- a/samples/Tutorials/03-DurableOutbox/README.md
+++ b/samples/Tutorials/03-DurableOutbox/README.md
@@ -124,3 +124,10 @@ The reference samples use Dapper, and Dapper is a fine choice. This one uses
 claim of this rung is that your write and Brighter's write share a transaction, and an
 explicit assignment argues that better than an extension method that takes `tx` as its third
 argument.
+
+## Where this sits on the ladder
+
+This rung and [rung 4](../04-Kafka) are each one delta from
+[rung 2](../02-FirstMessage) rather than from each other: this one keeps RabbitMQ and adds
+durability, rung 4 keeps the shape and swaps the transport for Kafka. Neither depends on the
+other, so they can be read in either order.
--- a/samples/Tutorials/04-Kafka/Greetings/GreetingEvent.cs
+++ b/samples/Tutorials/04-Kafka/Greetings/GreetingEvent.cs
@@ -1,0 +1,44 @@
+#region Licence
+/* The MIT License (MIT)
+Copyright © 2026 Ian Cooper <ian_hammond_cooper@yahoo.co.uk>
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the “Software”), to deal
+in the Software without restriction, including without limitation the rights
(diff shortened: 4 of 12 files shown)


iancooper:
Thanks — the finding is right and fixing it properly turned up something neither the README nor the review had.

**You are right that the stated cause could not be the whole story.** I measured it rather than reasoning about it. Immediately after all nine greetings are handled:

```text
TOPIC           PARTITION  CURRENT-OFFSET  LOG-END-OFFSET  LAG
greeting.event  0          -               3               -
greeting.event  1          -               3               -
greeting.event  2          -               3               -
```

Nothing committed at all, on any partition. So the batch path really does commit nothing — `_offsetStorage.Count % _maxBatchSize` runs 1→9 and never hits 10, exactly as you say — and every committed offset in the original README came from `sweepUncommittedOffsetsInterval`, which the text never mentioned.

**Where the correction does not hold is the consequence.** The prediction was that waiting past the sweep commits everything, so a later rebalance redelivers nothing. It does not. After the sweeper fires, each partition sits at `CURRENT-OFFSET 2`, `LAG 1` — and stays there:

```text
t+15s   p0 current=2 end=3 lag=1     p1 current=2 end=3 lag=1     p2 current=2 end=3 lag=1
t+30s   p0 current=2 end=3 lag=1     p1 current=2 end=3 lag=1     p2 current=2 end=3 lag=1
t+45s   p0 current=2 end=3 lag=1     p1 current=2 end=3 lag=1     p2 current=2 end=3 lag=1
t+60s   p0 current=2 end=3 lag=1     p1 current=2 end=3 lag=1     p2 current=2 end=3 lag=1
```

`LAG 1` is the stable end state, not a window. Your reading of `CommitAllOffsets` is what explains it: it drains a `ConcurrentBag` in no particular order into one `_consumer.Commit(list)`, so where a partition appears several times the offset that sticks is the last one written, not the highest. The sweeper is not a catch-up mechanism that converges.

**So the count is timing-dependent, but in the opposite direction, and never zero:**

| Second instance started | Redelivered (of 9) |
|---|---|
| Before the first sweep, within a few seconds | **7** |
| After the sweep | **3** — one per partition, on two separate runs |

That also confirms your second point directly. Started early, with a full offset bag, `CommitOffsetsFor`'s `tpo.Offset.Value > ptc.Offset.Value` filter commits nothing against a position of 3, and the redelivery count goes *up*, not down.

The README now prints both offset tables, names both commit paths, and gives that table instead of claiming "exactly three". The at-least-once conclusion is unchanged and is stronger for it: **no timing makes the number zero.**

All three nits fixed, and the first was a genuine defect — `bob`/`carol` in a comment fifteen lines below the comment explaining they were discarded, in a file whose blocks get quoted verbatim onto a docs page. Also `using System;` removed and the count interpolated. Rebuilt and re-run end to end after the edits: same assignment, same nine greetings, same grouping.

Two of your "not findings" are going into the tutorial page rather than the sample, which is where they belong: the SASL/TLS-via-`ConfigHook` sentence, and the note that dispose-flush runs through `OutboxProducerMediator` rather than the registry.

Issue #1, opened 2014-09-19
Support for multiple Application Layer Protocols in Task Queues


iancooper:
See the restms_support branch

PR #4282, opened 2026-08-28
Spec 0036: scoped lifetime per pipeline

### LINKED ISSUE #4256: Scoped should mean per-chain for mappers and transforms, with opt-in adoption of a caller's scope
### Summary

Brighter has three scopes for objects it creates via supplied factories for handlers, transformers, and mappers.

- **Transient**: We create the object, use it, then release it. 
- **Scoped**: We create the objects at the beginning of the request pipeline, and we tear them down at the end of the request pipeline. Where Brighter is used as a Command Dispatcher from an ASP.NET application, this scope is inherited from the HTTP Request; where Brighter is used as a Command Dispatcher from a Proactor or Reactor pump, the Proactor/Reactor should create a scope when it gets a message from the pump (before deserializing it via a transformer/mapper pipeline) and tears the scope down when the message terminates with ack/reject/requeue etc and we move to the next message on the pump. To be clear the lifetime is one message (to allow for per-message state).  Fresh message, fresh scope. 
- ** Singleton**: We create the objects once, and they last for the remainder of the program lifetime, usually at program startup.

Today, **scoped** does not work this way. Under `MapperLifetime.Scoped` (and the equivalent transform lifetime), the scope is bound to the **Dispatcher/process lifetime**, so state can be cached across messages and is only reclaimed at host shutdown.

### Current behaviour (as of #4254)

- `BuildDispatcher` (`Paramore.Brighter.ServiceActivator.Extensions.DependencyInjection/ServiceCollectionExtensions.cs`) is a **singleton** factory, so `ServiceProviderMapperFactory` / `ServiceProviderTransformerFactory[Async]` capture the **root** `IServiceProvider`.
- Under `MapperLifetime.Scoped`, `ServiceProviderLifetimeScope.GetOrCreateScoped` creates **one** `IServiceScope` off that root provider and caches every scoped mapper in it for the factory's entire lifetime. Fresh message → *same* scope.
- Nothing on the consumer pump (`Reactor`/`Proactor`/`Dispatcher`) opens a per-message DI scope: a grep for `CreateScope` / `IServiceScopeFactory` across `Paramore.Brighter.ServiceActivator` finds no call site. Handlers get per-resolution scopes via `ServiceProviderHandlerFactory`, but there is no unifying message-boundary scope.

### Why this matters

- A `Scoped` mapper/transform holds state (and any `IDisposable` it owns) from first use until process exit, rather than per message.
- The reviewer's suggestion on #4254 to dispose the consumer factories when the `Dispatcher` disposes only addresses **shutdown hygiene**; it does not make the scope per-message and would entrench "scope == Dispatcher lifetime" as the model.

### Proposed direction

> [!NOTE]
> **Superseded.** The direction below — one DI scope per dequeued message, shared by the mapper registry, the transform factories and the handler factory — was the starting point, but exploring it showed the scope unit should be the **chain** (one handler pipeline, or one transform pipeline), with the message as the *bound* rather than the scope itself. That model preserves [ADR 0039](https://github.com/BrighterCommand/Brighter/blob/master/docs/adr/0039-scoping-dependencies-inline-with-lifetime-scope.md), leaves the handler side already correct, and needs no change to the pump loop.
>
> **See the design narrative in the comment below** for the current direction, the evidence behind it, the constraints that shape it (notably: Microsoft's DI scopes do not nest), and the open questions.

~~Open a DI scope per dequeued message on the consumer pump and resolve the mapper registry, transform factories, and (ideally, aligned) handler factory from **that** scope, disposing it when the message completes (acked/rejected). This makes `Scoped` mean per-message on the consumer, matching the request-scope model, and removes cross-message state retention.~~

This remains an architectural change with its own test surface — deliberately **out of scope** for the memory-leak fix in #4254, which is confined to reclaiming the per-resolution `Transient` scopes that leaked in #4252.

### Related

- Follow-up from PR #4254 (fixes #4252), finding 4 of the latest review.

iancooper:
Closes #4256.

**Implementation is complete — all 82 tasks in [`tasks.md`](specs/0036-scoped-lifetime-per-pipeline/tasks.md) are done.** This was in draft while the design (7 ADRs, requirements, task breakdown) was visible during implementation; it now contains the full implementation, test suite and documentation and is ready for review.

## The problem

`ServiceLifetime.Scoped` did not mean *per-pipeline* for message mappers and transforms — in practice a `Scoped` mapper was cached for the life of the process. This made `Scoped` behave differently for Brighter artefacts than a user's intuition from ASP.NET Core, and gave no way for a pipeline to join a caller's existing scope, so a handler and a mapper in the same HTTP request couldn't share a `Scoped` `DbContext`.

## What changed

- `HandlerLifetime`, `MapperLifetime` and `TransformerLifetime` now govern a **pipeline-scoped** DI scope: a `Scoped` participant resolves from one DI scope shared by every `Scoped` participant on that pipeline, disposed when the pipeline ends.
- A new `Paramore.Brighter.Extensions.AspNetCore` package lets an ASP.NET Core pipeline **adopt** an ambient request scope instead of creating its own — one line, `AddBrighterRequestScope()`. Not opting in changes nothing.
- `ValidatePipelines()` gained seven new startup checks for common lifetime/scope-registration mistakes (three errors, four warnings), each pointing at a new guidance page: [`docs/guides/lifetimes-and-scoping.md`](docs/guides/lifetimes-and-scoping.md) — the lifetime model, the adoption truth table, a decision guide for choosing a lifetime triple, and a troubleshooting entry for each validation message.
- `Publish` subscribers and the consumer pump deliberately **never** adopt an ambient scope, preserving [ADR 0039](docs/adr/0039-scoping-dependencies-inline-with-lifetime-scope.md)'s per-subscriber isolation.
- **Thirteen breaking-change items**, catalogued as a single `release_notes.md` entry — the headline one is `MapperLifetime.Scoped` no longer caching across messages (no compatibility flag; migration is `MapperLifetime = Singleton`). Six interfaces gain members — the four mapper/transformer factories plus the two mapper registries — source- and binary-breaking on `netstandard2.0`, since no default interface member can absorb it, so every implementation in the repo moves in one commit.

## Design (7 ADRs, all `Accepted`)

Read in this order — compile-dependency order, not numeric:

| ADR | Decision |
| --- | --- |
| [0070](docs/adr/0070-per-pipeline-di-scope-for-mapper-and-transform-factories.md) | Per-pipeline DI scope shared by the mapper and transform factories |
| [0071](docs/adr/0071-pipeline-scope-handle-for-handler-pipelines.md) | Handler pipelines take their DI scope as a pipeline scope handle |
| [0072](docs/adr/0072-ambient-scope-adoption-seam.md) | Adopting an ambient DI scope — the resolution-source hand-off |
| [0073](docs/adr/0073-aspnet-core-request-scope-package.md) | ASP.NET Core's request scope as Brighter's ambient scope — a package of its own, one line to opt in |
| [0074](docs/adr/0074-lifetime-validation-evaluation-site.md) | Where the scope-configuration rules are evaluated |
| [0075](docs/adr/0075-publish-and-pump-scope-suppression.md) | Suppressing adoption for `Publish` subscribers and the consumer pump |
| [0076](docs/adr/0076-scope-affinity-option-and-write-through.md) | The affinity option, and how one setting reaches all four registration paths in any order |

[`requirements.md`](specs/0036-scoped-lifetime-per-pipeline/requirements.md) — 27 FRs, 10 NFRs, 55 acceptance criteria, the contract every test asserts against.

## Review process already applied

The design and task breakdown went through adversarial review rounds before implementation started, and every task went through TDD (test written and reviewed before implementation, full regression run after):

| Artefact | Rounds | Findings | Outcome |
| --- | --- | --- | --- |
| requirements | 2 | — | approved at revision 30 |
| design (7 ADRs) | 6 | — | all seven `Accepted` |
| tasks | 3 | 11 + 10 + 8 = **29** | all fixed, none Critical |

## Scope of the diff

462 files under `src/`/`tests/` (24,453 insertions), plus specs/ADRs/docs — 510 files total against `master`.

## Regression status

`Core.Tests`, `Extensions.Tests` and `Extensions.AspNetCore.Tests`, both net9.0/net10.0, run clean except two pre-existing flakes unrelated to this work (both confirmed passing in isolation, both predate this branch):
- `When_configuring_json_serialisation.Should_preserve_existing_options` (`Extensions.Tests`) — a parallel-test-execution race on a shared static `JsonSerializerOptions`.
- `PackageReferencedButExtensionNotCalledTests` (`Extensions.AspNetCore.Tests`) — a `WebApplicationFactory` parallel-hosts disposal race.

## Where a reviewer's attention is most valuable

1. **Is `Scoped`-means-per-pipeline the right default**, given it's a breaking change? (`docs/adr/0070-per-pipeline-di-scope-for-mapper-and-transform-factories.md`)
2. **ADR 0075's suppression rule** — `Publish` subscribers and the consumer pump deliberately do *not* adopt a caller's scope. The decision most likely to surprise.
3. **The seven validation rules and their messages** (`ScopeConfigurationRules.cs`) — each names the guidance page; `docs/guides/lifetimes-and-scoping.md` §6 is meant to be self-sufficient for fixing any of them without reading source.
4. **The breaking-change catalogue in `release_notes.md`** — one entry, thirteen items; flag if anything is missing or mis-stated.

https://claude.ai/code/session_01WzFDNrg2Q18Zkm9MpR5TXv


iancooper:
@claude Please re-review this PR. Since your last review (2026-09-16), 3 new commits (`fd77dcece`, `dbe70e0b0`, `d4ac64b8e`) were pushed that address all 3 findings from that review:

1. `d4ac64b8e` — `HandlerLifetimeScope` now has a real async disposal path (`DisposeAsync()` + finalizer, mirroring `TransformPipelineAsync`'s existing pattern), threaded through `IAmALifetime` (now `IAsyncDisposable` too) and `PipelineBuilder<T>`, with `CommandProcessor`'s async send/publish paths moved to `await using`. Scoped to the pipeline scope's own disposal only — individual handler `Release` stays sync, since `IAmAHandlerFactorySync`/`IAmAHandlerFactoryAsync.Release` have no async variant.
2. `dbe70e0b0` — `ServiceProviderLifetimeScope.ResolveOwnedArtefactCache`'s fallback no longer uses the no-lock `LazyInitializer.EnsureInitialized` overload (which could run its factory more than once under concurrent first callers and silently leak the losing `ScopedArtefactCache`); it now matches the `Interlocked.CompareExchange`-and-dispose-the-loser pattern already used by `EnsureRootScopePublished` in the same file.
3. `fd77dcece` — `ScopeConfigurationRules.DefeatedOptIn()`'s error factory no longer force-unwraps a `ScopeAffinityOverride` registered by factory delegate (`ImplementationInstance == null`), which previously threw a raw `NullReferenceException` instead of a validation message.

Each fix has its own regression test (linked in the commit). Please verify each fix actually addresses its finding, and re-check the rest of the diff for anything new.

iancooper:
Thanks for the re-review — response to each finding:

1. **`DefeatedOptIn` silence (medium) — fixed** in `ac46633de`. `DefeatedOptIn()` no longer declines when the override's value is unreadable; it keeps firing the Error and falls back to a value-free message ("An affinity override is registered, but the resolved `IBrighterOptions` was supplied by the application...") when it can't name the affinity. Regression test updated: the first fact now asserts specifically no `NullReferenceException` (a `PipelineValidationException` is the correct outcome under `throwOnError: true`), the second asserts both the `UnreadableOverride` Warning and the `DefeatedOptIn` Error are reported together, since they're two different problems that happen to co-occur.
2. **Predicate gap — closed for free** by #1's fix. `DefeatedOptIn` no longer special-cases unreadability at all, so there's no second predicate for `UnreadableOverride`'s trigger to diverge from.
3. **Release-notes gap (medium) — fixed** in `ec60be748`. Added a fourteenth breaking-change catalogue item for `IAmALifetime` also gaining `IAsyncDisposable`, and extended its per-interface migration bullet with the `DisposeAsync()` implementation shape.
4. **Finalizer cost — acknowledged, no change.** `~HandlerLifetimeScope()` is a deliberate mirror of `TransformPipelineAsync`'s existing pattern (same house style, same safety-net reasoning) rather than something arrived at by accident. Leaving as-is.
5. **`DisposeAsync()` abort-on-throw — accepted for now.** Matches the pre-existing `Dispose()`'s own behaviour, and nothing in-repo exercises it (`HandlerLifetimeScope` never throws from dispose). Will revisit if a third-party `IAmALifetime` implementation actually hits it in practice.

All three regression suites (`Core.Tests`, `Extensions.Tests`, `Extensions.AspNetCore.Tests`, both net9.0/net10.0) clean apart from the same two pre-existing flakes noted in earlier passes on this PR.

iancooper:
@claude Please do one more re-review pass — likely the last one for this PR. Since your previous re-review, 2 more commits were pushed addressing findings #1 and #3 from that pass (findings #2, #4 and #5 were resolved without a code change - see the comment above for details):

1. `ac46633de` — `DefeatedOptIn()` no longer declines to fire when the affinity override's value is unreadable (registered by factory delegate). It now keeps reporting the Error unconditionally when the opt-in is genuinely defeated, falling back to a value-free message ("An affinity override is registered, but the resolved `IBrighterOptions` was supplied by the application rather than by Brighter...") only when it can't name the affinity value. Regression test `DefeatedOptInMeetsUnreadableOverrideTests` rewritten: one fact asserts specifically no `NullReferenceException` (a `PipelineValidationException` is now the correct outcome under `throwOnError:true`), the other asserts both the `UnreadableOverride` Warning and the `DefeatedOptIn` Error fire together.
2. `ec60be748` — docs only. Added a 14th breaking-change catalogue item to `release_notes.md` recording that `IAmALifetime` also gained `IAsyncDisposable`, and extended its per-interface migration bullet with the `DisposeAsync()` implementation.

Please verify fix #1 actually closes the gap (an app whose opt-in is genuinely defeated should still fail startup validation when `throwOnError:true`, even if the override was registered by factory delegate), and do a final check of the whole PR for anything not yet caught.

iancooper:
All three findings from the final re-review addressed:

1. **Docs guide §6 stale (medium, NFR-10) — fixed** in `386efee78`. "Defeated opt-in" now notes the affinity-omitting message variant; "Unreadable override" no longer unconditionally claims the override "still takes effect" and points at "Defeated opt-in" for the case where that claim doesn't hold.
2. **`ResolveOwnedArtefactCache` resolver-vs-Dispose leak (low) — fixed** in `35ee1b3bf`. `ResolveOwnedArtefactCache` now re-checks `_disposed` after winning the publish and reclaims-and-disposes if a concurrent `Dispose()` already ran, mirroring `EnsureRootScopePublished`'s existing pattern. All four teardown paths (`Dispose`, `DisposeAsync`, `DisposeSurfacing`, `DisposeSurfacingAsync`) now claim `_ownedFallbackCache` atomically via `Interlocked.Exchange` instead of a direct read, so exactly one side ever disposes a given cache instance — `ScopedArtefactCache.Dispose()` isn't idempotent, so a double-dispose would have under-counted rather than no-op'd. Regression test drives the race through public surface only and reproduced the leak in 3-12/200 trials against the unfixed code across three runs.
3. **Stale doc comment (nit) — fixed** in `386efee78`. `ContainerRegistrationSnapshot`'s `<remarks>` no longer says "Nothing calls this type yet".

All three regression suites (`Core.Tests`, `Extensions.Tests`, `Extensions.AspNetCore.Tests`, both net9.0/net10.0) clean apart from the same two pre-existing flakes noted throughout this PR's review history.

That's everything from this review pass. Thanks for the thorough look across all three rounds — moving to merge next.

Issue #4240, opened 2026-07-17
[Feat] Test Generator: Brighter-provided Reject/DLQ/requeue-with-delay flows as universal transport needs Generator Tests


iancooper:
## Problem

The messaging gateway generator's optional feature flags — `HasSupportToDelayedMessages` and `HasSupportToDeadLetterQueue` — are treated as *native*-capability switches, and are set to whatever makes the generated suite green. This is wrong on two counts:

1. **They gate behaviour Brighter provides universally.** Brighter falls back to its own scheduler for delayed send/requeue when a transport has no native delay, and provisions a dead-letter (and invalid-message) producer driven by the `Reject` flow when a transport has no native DLQ. `Reject(Message, MessageRejectionReason?)` is on `IAmAMessageConsumerSync`/`Async` for **every** transport. So these are universal conformance obligations, not opt-in capabilities.

2. **The flags are already mis-declared against the code.** Postgres has native delay (delay column in the insert) and native/provided DLQ, yet declares `HasSupportToDelayedMessages: false` and `HasSupportToDeadLetterQueue: false`. AWS SQS has native `DelaySeconds` (≤15 min) and declares delay `false`. RocketMQ is the only config declaring delay `true`, so the delayed-delivery test runs for exactly one transport.

The root cause is timing: **the Reject→DLQ / invalid-channel / requeue-via-producer / requeue-via-scheduler features post-date the generator.** Because they couldn't be expressed as templates, contributors hand-wrote them per transport instead. The result is broad but inconsistent duplication.

## Evidence: the same tests, hand-written per transport

Inventory of hand-written (non-generated) coverage across transport test projects:

| Transport | Reject→DLQ | Reject→invalid | Requeue w/ delay | Plain requeue | Delayed send | Fallback ladder / no-channel |
|---|---|---|---|---|---|---|
| RMQ.Async | ✅ | — | ✅ (producer) | — | — | — |
| RMQ.Sync | ✅ | — | ✅ (producer) | ✅ | ✅ | — |
| Kafka | ✅ | ✅ | ✅ (producer + scheduler) | ✅ | — | ✅ (+ metadata stamping) |
| AWS / AWS.V4 | ✅ | ✅ | — | — | — | ✅ |
| Redis | ✅ | ✅ | ✅ (producer) | ✅ (zero-delay) | — | ✅ |
| PostgresSQL | ✅ | ✅ | ✅ (native SQL) | ✅ | — | ✅ |
[...]

PR #4283, opened 2026-08-29
fix/4253 Kafka timestamp timezone lossy

### LINKED ISSUE #4253: Kafka gateway: legacy TimeStamp header drifts by host UTC offset per round-trip (timezone-lossy write/read)
## Summary

The Kafka gateway's **legacy `TimeStamp` header** is written and read in a timezone-lossy
way, so a message's `Header.TimeStamp` drifts by the host's UTC offset on every round-trip
through Kafka. On a non-UTC host this produces both a wrong wall-clock representation (every
hop) and a **genuine instant drift that accumulates on each re-publish** (e.g. a requeue,
which is two hops → the timestamp advances by one host-offset). On a UTC host the defect is
invisible, which is why it survives CI.

## Affected component

`src/Paramore.Brighter.MessagingGateway.Kafka` — the legacy `TimeStamp` header path only.
The CloudEvents time path (`ce_time`) is **correct** and should be treated as the reference.

## Root cause

Two facets, one on each side of the round-trip:

**Write side — offset is dropped.**
`KafkaDefaultMessageHeaderBuilder.cs:59-60` serialises the timestamp as:

```csharp
var timeStampAsString = message.Header.TimeStamp.DateTime != default
    ? message.Header.TimeStamp.DateTime.ToString(CultureInfo.InvariantCulture)
    : DateTimeOffset.UtcNow.DateTime.ToString(CultureInfo.InvariantCulture);
headers.Add(HeaderNames.TIMESTAMP, timeStampAsString.ToByteArray());
```

`DateTimeOffset.DateTime` returns the clock-time component with `Kind = Unspecified` and
**no offset marker** in the string. The offset is lost on the wire.

**Read side — assumed-UTC is converted to host-local.**
`KafkaMessageCreator.cs:252` re-parses it as:

```csharp
if (DateTime.TryParse(lastHeader!.FromByteArray(), DateTimeFormatInfo.InvariantInfo,
        DateTimeStyles.AssumeUniversal, out DateTime timestamp))
    return new HeaderResult<DateTimeOffset>(timestamp, true);
```

`AssumeUniversal` **without** `AdjustToUniversal` treats the offset-less string as UTC and
then **converts it to host-local time** (`Kind = Local`). The implicit `DateTime → DateTimeOffset`
conversion then stamps the host offset on it.

**Net effect (host at UTC+O):**
- **Hop 1** (original send, value at UTC+00:00): `.DateTime` → UTC wall-clock → read re-anchors
  to local, so the *instant is preserved* but the value is now represented at `+O`. Wall-clock
  string is off by `O`.
- **Hop 2+** (any re-publish, e.g. a requeue — the value is now at `+O`): `.DateTime` writes the
  **local** wall-clock offset-less, and the read re-assumes it as UTC → the **instant drifts by
  `+O` per subsequent hop**.

Observed on a **BST (+01:00)** host via the conformance requeue tests (two hops):
`KafkaMessageAssertion.cs:50` compares `expected.Header.TimeStamp.ToString("HH:mm:ss")` against
`actual...ToString("HH:mm:ss")` and fails with `Expected 20:50 / Actual 22:50` — `+2h` (1h of
instant drift + 1h of representation offset).

The CloudEvents path is unaffected because it round-trips with offset intact:
`KafkaDefaultMessageHeaderBuilder.cs:87` writes `TimeStamp.ToRfc3339()` (RFC 3339 carries the
offset) and the fallback read `KafkaMessageCreator.cs:268` uses offset-aware
`DateTimeOffset.TryParse`.

## Impact

- On any **non-UTC host**, `Message.Header.TimeStamp` is wrong after consuming a Kafka message:
  wrong wall-clock every time, and a genuinely wrong instant after any re-publish/requeue.
- This is **production** behaviour, not just a test artifact — any consumer that reads
  `Header.TimeStamp` (ordering, TTL/age checks, auditing) on a non-UTC-configured host is
  affected.
- It is invisible on UTC hosts (including CI), so it is easy to ship.

## How it was found

Surfaced during the universal transport conformance work (spec 0036, issue #4240). The Kafka
requeue/redeliver conformance tests (FR-22/15/16) failed deterministically on a BST host and
passed under `TZ=UTC`. The conformance ledger records those behaviours as `Pass` (they conform
in the reference UTC environment); this timezone defect was deliberately **carved out as its
own bugfix** rather than fixed inside the conformance ledger.

## Reproduction

1. On a host whose local time is **not** UTC (e.g. BST / `TZ=Europe/London` in summer):
2. Run a Kafka send→requeue→redelivery round-trip and compare the redelivered
   `Header.TimeStamp` wall-clock to the sent one — it is off by the host offset (×2 after a
   requeue). Equivalent repro in this repo:
   ```
   dotnet test tests/Paramore.Brighter.Kafka.Tests --framework net10.0 \
     --filter "FullyQualifiedName~WhenRequeuingAFailedMessageShouldBeRedelivered"
   ```
   Fails on a BST host at `KafkaMessageAssertion.cs:50`; passes when prefixed `TZ=UTC`.

## Constraints on the fix (why the naive one-liner was reverted)

A previous speculative one-sided change to the **read** side only
(`AssumeUniversal` → `AssumeUniversal | AdjustToUniversal` at `KafkaMessageCreator.cs:252`) was
**reverted** — it regressed a hand-written offset expectation and does not address the write
side that drops the offset. The correct fix must treat **write and read as a pair** and keep
these existing unit tests green (or update them deliberately, with rationale):

- **Write-side format is pinned:** `When_converting_brighterheader_to_kafkaheader.cs:60`
  asserts the header equals `message.Header.TimeStamp.DateTime.ToString(InvariantCulture)`.
  Changing the on-wire format (e.g. to ISO-8601 with offset / `Z`) breaks this test by design.
- **Read-side is instant-tolerant:** `When_converting_kafkaheader_to_brighterheader.cs:69`
  asserts `Assert.Equal(TimeStamp, readTimeStamp, TimeSpan.FromSeconds(5))` (DateTimeOffset
  equality is instant-based, so it passes single-hop today but does not catch multi-hop drift).

## Suggested direction (for `/bugfix` to confirm)

Align the legacy `TimeStamp` header with the already-correct CloudEvents approach: serialise
with the offset preserved (round-trip as UTC, or RFC 3339 / ISO-8601 with offset) on the write
side, and parse offset-aware (or `AssumeUniversal | AdjustToUniversal` so the value stays UTC)
on the read side — the two changes must land together. Update the two constraint unit tests
above to the new format with rationale.

## Suggested regression test

A **multi-hop** round-trip test (send → requeue/redeliver → assert same instant AND same
UTC wall-clock) that runs green under a non-UTC `TZ`, so it fails today and would only pass
once both sides are fixed. The existing single-hop instant-tolerant assertion does not catch
the accumulating drift.

## References

- `src/Paramore.Brighter.MessagingGateway.Kafka/KafkaDefaultMessageHeaderBuilder.cs:59-63` (write, offset dropped)
- `src/Paramore.Brighter.MessagingGateway.Kafka/KafkaDefaultMessageHeaderBuilder.cs:87` (CloudEvents write — correct, RFC 3339)
- `src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageCreator.cs:247-260` (read, `AssumeUniversal` local conversion)
- `src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageCreator.cs:267-270` (CloudEvents read fallback — correct, offset-aware)
- `tests/Paramore.Brighter.Kafka.Tests/MessagingGateway/Reactor/When_converting_brighterheader_to_kafkaheader.cs:60` (write-format constraint)
- `tests/Paramore.Brighter.Kafka.Tests/MessagingGateway/Reactor/When_converting_kafkaheader_to_brighterheader.cs:69` (read instant-tolerant constraint)
- `tests/Paramore.Brighter.Kafka.Tests/KafkaMessageAssertion.cs:50` (conformance assertion that surfaces it)
- Related: #4240 (universal transport conformance), ADR 0046 (Kafka DLQ), ADR 0067 (conformance rollout & deferral governance)

Abhishek84313:
## Description

The Kafka gateway's legacy `TimeStamp` header was written and read in a timezone-lossy way, so
`Message.Header.TimeStamp` drifted by the host's UTC offset on every round-trip through Kafka.

Two facets, one on each side, fixed together:

- **Write** (`KafkaDefaultMessageHeaderBuilder`) serialised `TimeStamp.DateTime` — the clock-time
  component, `Kind = Unspecified`, formatted with no offset marker. The offset never reached the wire.
- **Read** (`KafkaMessageCreator.ReadTimeStamp`) parsed with `DateTimeStyles.AssumeUniversal` but
  *without* `AdjustToUniversal`, so the offset-less string was taken as UTC and then converted to
  host-local; the implicit `DateTime` to `DateTimeOffset` conversion stamped the host offset on it.

On a host at UTC+O this gave a wrong wall clock on every hop, and — because a re-publish then wrote
that *local* wall clock offset-less — a genuinely wrong instant, advancing by O on each subsequent
hop. A requeue is two hops. On a UTC host it is invisible, which is why it survived CI.

The fix aligns the legacy header with the CloudEvents `ce_time` header beside it, which already
round-trips correctly:

- Write the header as `TimeStamp.ToRfc3339()` — UTC, offset-bearing. This also matches the legacy
  timestamp header written by the Redis, GCP PubSub, RocketMQ and AWS gateways.
- Read with `DateTimeOffset.TryParse(..., AssumeUniversal | AdjustToUniversal)`, keeping the value
  anchored to UTC instead of re-stamping it with the host offset.

**Wire-compatible in both directions.** `AssumeUniversal` is deliberately retained rather than
replaced by plain offset-aware parsing, so a new consumer still reads the legacy offset-less values
in flight from old producers — taking them as UTC and leaving them there. In the other direction, an
old consumer's `DateTime.TryParse(..., InvariantInfo)` parses an RFC 3339 `...Z` string, so old
consumers keep working against new producers.

Also fixed, same defect class and same method: the binary Unix-ms fallback read
`DateTimeOffset.FromUnixTimeMilliseconds(...).DateTime`, whose trailing `.DateTime` discarded the
(zero) offset and let the implicit conversion re-stamp the host offset, corrupting the instant for
any producer writing a binary timestamp.

Diagnosis, evidence and scope decisions are recorded in
`bugfixes/0017-kafka-timestamp-header-timezone-lossy/bugfix.md`.

## Related Issues

Fixes #4253

Surfaced by the universal transport conformance work (#4240); first noticed as the 9 unrelated
failures recorded in the Verification section of `bugfixes/0004-kafka-hasfatalerror-not-latched`.

## Type of Change

- [ ] Documentation update

The on-wire format of the `TimeStamp` header changes (invariant-culture `DateTime` to RFC 3339). It
is not marked breaking because it round-trips against both older and newer peers, as described above;
the pinned write-format unit test changes by design.

## Checklist

Notes on the above, so reviewers know what each tick actually covers:

- **Documentation** — checked the in-repo docs (`docs/transports/`, `docs/Kafka-DeadLetterQueue-Usage.md`,
  the Kafka ADRs 0046/0055). The legacy `TimeStamp` header's wire format is not documented anywhere,
  so there is nothing to update here. CONTRIBUTING also asks that changes be documented in the
  BrighterCommand **Docs** repository — that is a separate repo and has not been touched.
- **XML documentation** — nothing to add: no public API surface changed. Both edits are inside private
  methods (`AddBrighterHeaders`, `ReadTimeStamp`), and per CONTRIBUTING we do not add documentation
  comments to non-exported members. The rationale is carried in inline comments at both sites instead.
- **Code style** — test methods follow `When_[condition]_should_[expected_behavior]`, one `[Fact]` per
  file named for the method, `//arrange //act //assert` structure, Evident Data in the fixture, and no
  mocks (real builder and creator throughout).

## Additional Notes

**Regression tests** — two broker-free facts, green under any host timezone:

- `tests/.../Reactor/When_round_tripping_a_timestamp_should_preserve_the_instant_across_hops.cs` —
  round-trips a timestamp at `+05:00` and asserts the instant and its UTC wall clock survive, that what
  is read back is anchored to UTC (`Offset == TimeSpan.Zero`, which is what makes a re-publish
  idempotent), and that re-publishing what was read produces byte-identical `TimeStamp` header bytes,
  so drift cannot accumulate over hops.
- `tests/.../Reactor/When_reading_a_timestamp_from_an_older_producer_should_treat_it_as_utc.cs` — an
  offset-less invariant-format value is read as UTC and left there, pinning the backward-compatibility
  contract above.

Because the input carries an explicitly non-UTC offset, the writer loses information no reader can
recover, so these fail on a UTC host too — a stronger guarantee than the issue's BST-only reproduction,
and CI-safe. Verified red against the pre-fix source (`Failed: 2, Passed: 0`), each for the right
reason: `Expected 2024-06-15T13:45:30+05:00 / Actual 2024-06-15T19:15:30+05:30`, and
`Expected 00:00:00 / Actual 05:30:00` for the offset.

`When_converting_brighterheader_to_kafkaheader.cs:60` pinned the old wire format and is updated to
`TimeStamp.ToRfc3339()` with a comment giving the rationale — that test exists to pin the format, so
changing the format changes it by design. The instant-tolerant read assertion at
`When_converting_kafkaheader_to_brighterheader.cs:69` still passes and is left alone; the new tests
cover the multi-hop drift it cannot see.

### Reviewer notes

**A separate pre-existing defect, found but deliberately not fixed here.** A consumed Kafka message
cannot be re-published and re-consumed in-process: `KafkaMessageCreator` puts `TopicPartitionOffset`
into `Header.Bag`, `KafkaDefaultMessageHeaderBuilder.AddUserDefinedBagHeaders` writes the bag back out
as a header (it is in neither `BrighterDefinedHeaders.HeadersToReset` nor `MessageHeader.IsLocalHeader`),
and the next read then throws `ArgumentException: An item with the same key has already been added.
Key: TopicPartitionOffset` in `ReadBagEntry` — which `CreateMessage` swallows into a
`Message.FailureMessage`. The `ce_*` headers are duplicated by the same route. This is what blocked an
end-to-end two-hop *read* assertion, so the second hop is asserted at the wire level instead. Out of
scope for this PR; it probably deserves its own issue.

**Two things not verified locally, worth confirming in CI:**

- **`net9.0` was not run** — that runtime is not installed on the machine this was developed on, so only
  `net10.0` of the `net9.0;net10.0` matrix was exercised. The change introduces no API absent from either.
- **The live-broker suite was not run** — no Kafka broker was available. The broker-dependent tests fail
  locally with `Local: Broker transport failure` / `1/1 brokers are down` both with and without this
  change. The conformance requeue tests named in the issue (FR-22/15/16) therefore remain unverified
  end to end and should be green on a broker before merge.


iancooper:
Thanks @Abhishek84313 I will look.

Do you want to raise an issue for this one:

"A consumed Kafka message
cannot be re-published and re-consumed in-process: KafkaMessageCreator puts TopicPartitionOffset
into Header.Bag, KafkaDefaultMessageHeaderBuilder.AddUserDefinedBagHeaders writes the bag back out
as a header (it is in neither BrighterDefinedHeaders.HeadersToReset nor MessageHeader.IsLocalHeader),
and the next read then throws ArgumentException: An item with the same key has already been added. Key: TopicPartitionOffset in ReadBagEntry — which CreateMessage swallows into a
Message.FailureMessage. The ce_* headers are duplicated by the same route. This is what blocked an
end-to-end two-hop read assertion, so the second hop is asserted at the wire level instead. Out of
scope for this PR; it probably deserves its own issue."

src/Paramore.Brighter.MessagingGateway.Kafka/KafkaDefaultMessageHeaderBuilder.cs
```diff
@@ -56,11 +56,16 @@ private static void AddBrighterHeaders(Headers headers, Message message)
             headers.Add(new Header(HeaderNames.TOPIC, message.Header.Topic.Value.ToByteArray()));
             headers.Add(new Header(HeaderNames.MESSAGE_ID, message.Header.MessageId.Value.ToByteArray()));
             
-            var timeStampAsString = message.Header.TimeStamp.DateTime != default
-                ? message.Header.TimeStamp.DateTime.ToString(CultureInfo.InvariantCulture)
-                : DateTimeOffset.UtcNow.DateTime.ToString(CultureInfo.InvariantCulture);
+            //Write the timestamp as RFC 3339, normalized to UTC, so that its offset survives the wire.
```
iancooper, line 59: Let's drop the inline comments, or shorten them to a line

src/Paramore.Brighter.MessagingGateway.Kafka/KafkaMessageCreator.cs
```diff
@@ -248,15 +248,19 @@ private static HeaderResult<DateTimeOffset> ReadTimeStamp(Headers headers)
         {
             if (headers.TryGetLastBytesIgnoreCase(HeaderNames.TIMESTAMP, out var lastHeader))
             {
-                //Additional testing for a non unixtimestamp string
-                if (DateTime.TryParse(lastHeader!.FromByteArray(), DateTimeFormatInfo.InvariantInfo, DateTimeStyles.AssumeUniversal, out DateTime timestamp))
+                //Additional testing for a non unixtimestamp string. Parse offset-aware: we write RFC 3339, so
```
iancooper, line 251: Again, inline, we probably only want a single line comment, if not clear from the code

iancooper (commented):
The explanatory comments are fine in a test, because they add clarity if it breaks, but agents tend to pepper the code with them, and we could reduce this to nothing or a single line

iancooper:
Just some cosmetic cleanup @Abhishek84313 also, feel free to add yourself to CONTRIBUTORS.md

Abhishek84313:
@iancooper 
Agreed — trimmed. Both multi-line blocks in src are down to one line each. In KafkaDefaultMessageHeaderBuilder it's just "RFC 3339 keeps the offset on the wire; without it the reader re-anchors to host-local time". In KafkaMessageCreator I kept the existing "Additional testing for a non unixtimestamp string" comment and extended that same line rather than replacing it, since the AssumeUniversal | AdjustToUniversal pairing is the one bit that isn't readable from the code — it's what a future cleanup is most likely to simplify straight back into the bug.

In the tests the arrange/act/assert markers are back to bare. I left three short lines where the code doesn't state its own reason: why the fixture timestamp is at +05:00 rather than UTC (that's what makes it fail on a UTC host too, so it's really the point of the test), why there's a second Build call, and what the byte-equality assert is guarding. Happy to cut those as well if you'd rather have none.

That's 14 comment lines down to 5, tests unchanged and green.

One thing worth flagging: the creator's line is now around 150 characters, longer than its neighbours. If that's over the width you want, I'd sooner drop the added clause than wrap it back to two lines — let me know which you'd prefer.

Abhishek84313:
> Thanks @Abhishek84313 I will look.
> 
> Do you want to raise an issue for this one:
> 
> "A consumed Kafka message cannot be re-published and re-consumed in-process: KafkaMessageCreator puts TopicPartitionOffset into Header.Bag, KafkaDefaultMessageHeaderBuilder.AddUserDefinedBagHeaders writes the bag back out as a header (it is in neither BrighterDefinedHeaders.HeadersToReset nor MessageHeader.IsLocalHeader), and the next read then throws ArgumentException: An item with the same key has already been added. Key: TopicPartitionOffset in ReadBagEntry — which CreateMessage swallows into a Message.FailureMessage. The ce_* headers are duplicated by the same route. This is what blocked an end-to-end two-hop read assertion, so the second hop is asserted at the wire level instead. Out of scope for this PR; it probably deserves its own issue."

Yes — though I need to walk back part of what I wrote, because I'd overstated it.

I've since found CleanBagForResend in KafkaMessageConsumer, which strips TopicPartitionOffset and all the ce_* keys, and it's called on every resend path — Requeue, RequeueAsync, and DLQ/reject via RefreshMetadata. So the live requeue→redelivery path is safe, and the conformance tests are right to pass. My "a consumed message cannot be re-published and re-consumed" was only true of a raw builder→creator round-trip, which is what I hit in the test, not of the gateway's own paths. Sorry for the noise.

What's left is still worth an issue, just a smaller one than I made it sound. The guard is a hand-maintained deny-list living in the consumer rather than a property of the bag key itself. TopicPartitionOffset looks like exactly what MessageHeader.IsLocalHeader / RegisterLocalHeader exist for — a bag key that should never be serialised out — and registering it there would make the builder skip it on every republish route for free, instead of each caller having to remember CleanBagForResend. The failure mode when someone forgets is also quiet in an unhelpful way: ReadBagEntry throws, CreateMessage catches it and returns a Message.FailureMessage, so the message is lost with only a log line.

Happy to raise that as "Kafka: TopicPartitionOffset should be a local header rather than a consumer-side deny-list" if you think it's worth tracking, or to drop it if you'd rather not carry the ticket.

iancooper:
>Happy to raise that as "Kafka: TopicPartitionOffset should be a local header rather than a consumer-side deny-list" if you think it's worth tracking, or to drop it if you'd rather not carry the ticket.

Let's do that

iancooper:
> @iancooper Agreed — trimmed. Both multi-line blocks in src are down to one line each. In KafkaDefaultMessageHeaderBuilder it's just "RFC 3339 keeps the offset on the wire; without it the reader re-anchors to host-local time". In KafkaMessageCreator I kept the existing "Additional testing for a non unixtimestamp string" comment and extended that same line rather than replacing it, since the AssumeUniversal | AdjustToUniversal pairing is the one bit that isn't readable from the code — it's what a future cleanup is most likely to simplify straight back into the bug.

Hmmm, don't see that in the current PR. Can you check that got committed and pushed?

iancooper (approved):

iancooper:
Thanks @Abhishek84313 going to merge

PR #4284, opened 2026-08-29
Fix 4240

Abhishek84313:
## Description

Design work for the messaging gateway test generator: a spec and an ADR, no code changes.

The generator's `HasSupportToDelayedMessages` and `HasSupportToDeadLetterQueue` flags read as
"does this broker have this natively?", but they sit in front of templates that exercise behaviour
Brighter supplies *for* a transport — the `Reject` routing ladder, the Brighter-provisioned dead
letter and invalid message channels, and scheduler-backed delay. In practice they became "set this
to whatever keeps the suite green", and the declarations have drifted from the code.

Because the generator couldn't express those behaviours, contributors hand-wrote them once per
transport. Reject→DLQ and the fallback ladder now exist on essentially every modern transport,
reject→invalid on most, requeue-with-delay on six, rejection metadata on one. GCP Pub/Sub has none.
Delayed send is generated for 3 of 21 gateway configurations.

This PR proposes making those behaviours generator-owned and on by default, re-scoping the two flags
to native capability only, and replacing the escape hatch with **named conformance waivers** that
carry a reason and a tracking issue.

Adds:

- `specs/0036-generator-universal-rejection-tests/` — requirements, verified inventory, task breakdown
- `docs/adr/0070-generator-owned-rejection-and-delay-conformance.md` — the design (status: Proposed)
- `specs/README.md` and regenerated `docs/adr/index.md`

### Three corrections to the issue's premises, found while verifying

1. **"Ungated" isn't reachable today.** Six of the 21 gateway configurations can't satisfy the
   Brighter-provided rejection tests: `RmqSubscription` (both variants) exposes `DeadLetterRoutingKey`
   as get-only DLX config and implements neither `IUseBrighterDeadLetterSupport` nor
   `IUseBrighterInvalidMessageSupport`; GCP Pub/Sub and Azure Service Bus use native DLQ and implement
   neither. Ungating outright lands red CI on GCP ×4 and RMQ.Async ×2 with no in-scope fix — hence
   waivers rather than a bare ungate.
2. **Delayed send runs for three configurations, not one** — AWS `SqsStandard`, AWS.V4 `SqsStandard`,
   RocketMQ. Substance of the issue unchanged.
3. **MQTT, RMQ.Sync and Azure Service Bus have no `test-configuration.json`**, so templates never
   reach them and their hand-written tests can't be retired by this work. MSSQL is also mis-declared
   (`HasSupportToDeadLetterQueue: false` while `MsSqlSubscription` implements both interfaces), which
   the issue doesn't list alongside Postgres and SQS.

### Two deviations from the issue, argued in the ADR

- **Not `InMemoryScheduler`** for the generated tests. It takes an `IAmACommandProcessor` and fires by
  posting through it, so every gateway test would need a command processor, producer registry and
  outbox — the opposite of the stated intent of not testing across gateway *and* scheduler. A
  generated `RecordingMessageScheduler` covers both the wiring assertion and real redelivery.
- **`bool setupDeadLetterQueue` is replaced, not extended.** A boolean can't express "invalid channel
  only", which the fallback-ladder tests need. Cost: all 21 provider classes break in one change —
  called out in the ADR as the design's largest cost.

Both defects in the issue are confirmed as described: the `requeuing_with_delay` template calls
`Requeue(received)` with no `TimeSpan?`, so it never exercises a delayed requeue; and `SkipTest`
substring-matches the full template path, so that filename trips both `requeuing` and `with_delay`.

## Related Issues

Refs #4240 (design only — implementation follows in a separate PR)

## Type of Change

- [x] Documentation update

## Checklist

- [ ] I have added/updated XML documentation for any public API changes — N/A, no code changes
- [ ] I have added/updated tests as appropriate — N/A, design only; the ADR is *about* tests

ADR carries frontmatter per `.agent_instructions/adr_frontmatter.md`; `docs/adr/index.md` regenerated
from frontmatter rather than hand-edited.

## Additional Notes

Deliberately stopped at the ADR rather than implementing. Phase 1 breaks all 21 provider classes at
once, and per CONTRIBUTING.md the ADR should get review before that lands. `tasks.md` has the
six-phase breakdown; Phase 0 (the two template defects) is self-contained and could go first.

Open questions for reviewers:

1. Waivers vs. a straight ungate — is accepting red CI on GCP and RMQ.Async the better forcing
   function? The ADR argues no, but that's a judgement call about how hard to push the commitment.
2. The waived-transport gaps (RabbitMQ/GCP/ASB rejection interfaces, `IAmAChannelFactoryWithScheduler`
   on four more factories) are scoped out here and need their own issues — happy to open them.
3. Whether behaviour 4's weaker shared assertion ("no throw, no redelivery") is enough, given the
   transports genuinely differ on what "no channels configured" does.


iancooper (commented):
Sorry, but this one is still in development; it hasn't been checked in yet. There are plenty of bugs in the free issues. We try to update the labels on in-flight issues. You will notice that the issue carries a "2 - In Progress" tag.

PR #4297, opened 2026-09-01
FEAT: Generated transport conformance tests should cover new reject/dlq/requeue paths

iancooper:
## Generated transport conformance tests - should cover new reject/dlq/requeue paths

Contributes to #4240, which stays **open** deliberately.

All 46 `Deferred` cells in the conformance ledger, and every generated `Skip` marker, resolve to
#4240. Closing it on merge would leave each of those pointing at a closed issue, and
`LedgerSkipCrossCheckAudit` never queries the tracker (ADR 0067 step 7), so nothing in CI would
ever notice - a silent skip with extra steps, in a PR whose central claim is that there are none.
#4240 therefore remains the umbrella tracker for the deferred behaviours, and carries a comment
recording what this PR completed.

While generating the transport conformance tests, I added new functionality around reject/dlq/requeue. This functionality can now be universal; where the message-oriented middleware doesn't support it, we can fall back to a Brighter substitute.

This replaces the four legacy per-transport gate keys, which controlled generation of functionality for these features, dependent on native support, with a **generated canonical conformance suite** that runs
against every gateway.

As some gateways have not yet implemented Brighter support for missing features, we also added a checked-in **conformance ledger** (`specs/0036-universal-transport-conformance-tests/conformance-status.md`) recording per-configuration conformance for 11 canonical behaviors (FR-2, 4, 5, 6, 7, 8, 9, 15, 16, 17, 22).

Every canonical test's `Skip` is driven by this ledger, so a behavior is either proven against a real broker or carries a signed-off `Deferred -> #4240` marker. No silent skips.

### Status — 62 of 62 tasks; phases 0–6 complete

All **24 wired configuration rows are resolved — zero `Unknown` cells**, across all twelve targeted
transports:

| Configuration | Result |
|---|---|
| AWS V3 x4 + V4 x4 | 11 Pass, except `SqsFifo` FR-9 `Deferred` (SQS FIFO rejects per-message delay) and `Sns*` FR-9 `Fixed` |
| Kafka Classic / Consumer / PartitionKey | 11/11 Pass |
| PostgresSQL | 11/11 Pass (native visibility-timeout redelivery) |
| MSSQL, Redis | 10 Pass + FR-16 `Deferred` (destructive read) |
| RMQ.Async Classic / Quorum | 10 Pass + FR-5 `Deferred` (invalid-channel routing) |
| RMQ.Sync | 10 `Fixed` + FR-5 `Deferred`, mirroring RMQ.Async |
| MQTT | 10 `Fixed` + FR-16 `Deferred` |
| RocketMQ | 9 `Fixed` + FR-2 / FR-15 `Deferred` (upstream no-op `Requeue`) |
| GCP Pull / PullOrdering | 4 Pass + FR-9 `Fixed` + 6 `Deferred` (emulator-only verification) |
| GCP Stream / StreamOrdering | all 11 `Deferred` — streaming pull hangs on the emulator, excluded in CI, no real creds |
| AzureServiceBus | 9 Pass + FR-5 / FR-9 `Deferred` — verified against the live namespace `azure-ci` already uses. FR-5 is a platform difference (ASB dead-letters natively; there is no separate invalid-message channel), FR-9 a real defect, raised as [#4318](https://github.com/BrighterCommand/Brighter/issues/4318) |

46 of the 264 cells are `Deferred`, each carrying a signed-off `#4240` marker.

### Transport `src` changes (deliberately localized)

- `Paramore.Brighter.MessagingGateway.RocketMQ` — guard the `Baggage` property; empty baggage crashed every send.
- `Paramore.Brighter.MessagingGateway.GcpPubSub` — `GcpPullMessageConsumer.Receive/ReceiveAsync` ignored `timeOut` and long-polled, blocking the pump; now bound via `CallSettings` expiration. A null timeout returns null settings, leaving the client's own per-method default in force — `Expiration.None` would have overridden it with no deadline at all.
- `Paramore.Brighter.MessagingGateway.AWSSQS` (V3 + V4) — sync `SendWithDelay` passed `TimeSpan.Zero`.
- `Paramore.Brighter.MessagingGateway.MQTT` — `Receive` returned immediately on an empty buffer, so every caller needed an external sleep. The MQTT event handler now releases a semaphore per message: `Receive` blocks on it, `ReceiveAsync` awaits it and honours the caller's cancellation token. The producer takes an `InstrumentationOptions` constructor parameter, as every other gateway producer does, rather than always tracing at `All`.
- `Paramore.Brighter.MessagingGateway.RMQ.Async` — **`RmqSubscription`'s `isDurable` default changes from `false` to `true`** (both constructors). See the behaviour-change note below.

Everything else is test-harness, generator template, or ledger work.

> ### ⚠️ One user-visible behaviour change — `RMQ.Async` subscriptions now default to durable queues
>
> `RmqSubscription`/`RmqSubscription<T>` in `Paramore.Brighter.MessagingGateway.RMQ.Async` change
> `bool isDurable = false` to `bool isDurable = true` (`RmqSubscription.cs:115` and `:180`).
>
> **Why.** RabbitMQ 4.3 blocks transient non-exclusive queues by default, so the previous default
> makes every affected subscription fail to declare against a 4.3+ broker — [#4355](https://github.com/BrighterCommand/Brighter/issues/4355).
>
> **What it means for existing users.** A caller who never passed `isDurable` and relied on the
> transient default will now get a **durable** queue: queue definitions survive broker restart, and
> the declare will **fail with a `PRECONDITION_FAILED` channel error against an existing queue that
> was declared non-durable**, since durability is immutable after declaration. Callers who want the
> old behaviour should pass `isDurable: false` explicitly. Nothing else about the subscription changes.
>
> **The change is deliberately asymmetric, and `#4355` stays open.** `RMQ.Sync`'s `RmqSubscription`
> still defaults to `false` (`:106`, `:165`). `RMQ.Sync` is independently non-functional against
> RabbitMQ 4.3 — a measured 46 failures, all on `transient_nonexcl_queues`, **reproduced unchanged on
> a pre-branch commit**, so it is pre-existing rather than this PR's doing. Fixing it is out of scope
> here; #4355 tracks the remainder. Until then, **"the Brighter default" differs between the two RMQ
> transports** and should not be quoted without saying which.

### Terminal cleanup — done

Gated on the zero-`Unknown` ledger, and performed in the order the acceptance criteria require (the
`test-configuration.json` keys last, so each earlier step was provably a no-op):

- the four legacy templates and their generated copies deleted — 8 templates + 72 copies, 80 files;
- the `SkipTest` gate branches and the `LegacyGatedTemplates` closed list stripped;
- the three retired properties removed from `MessagingGatewayConfiguration`;
- the three gate keys removed from every `test-configuration.json` (9 files, 63 occurrences).

Regeneration after each step was a **content no-op** — not one generated file differs — which is the
proof that nothing was still reading those keys. The three *retained* capability gates
(`HasSupportToPublishConfirmation`, `HasSupportToValidateBrokerExistence`,
`HasSupportToValidateInfrastructure`) keep their meaning and are still tested.

One retained gate was found to be **mis-declared in the same way the retired ones were**, and is
corrected here rather than left. `Kafka / Consumer` was the only Kafka configuration setting
`HasSupportToValidateInfrastructure: false`, so `Classic` and `PartitionKey` ran `assume_channel` and
`validate_channel` while `Consumer` ran neither. Measured against a live broker, that configuration
splits cleanly: `validate_channel` **passes** both variants, `assume_channel` **fails** both,
deterministically — the flag was suppressing two working tests in order to suppress two that do not
work. A narrower gate, `HasSupportToDetectMissingInfrastructureOnAssume`, now skips `assume_channel`
alone; `HasSupportToValidateInfrastructure` is unchanged in meaning, so MQTT and Redis are unaffected.
The Kafka suite goes 188 -> **190 pass / 0 fail**, by exactly the two tests recovered.

The underlying cause is a real behavioural difference, declared rather than papered over: `EnsureTopic()`
returns immediately on `OnMissingChannel.Assume` without an admin call, and the test compose disables
topic auto-creation, so the topic genuinely does not exist. Captured against a live broker, the classic
consumer raises `ChannelFailureException` wrapping `ConsumeException: Subscribed topic not available:
<topic>: Broker: Unknown topic or partition`, while the KIP-848 consumer returns `MT_NONE` with no error
at all.

**Fixing that error surfacing in `KafkaMessageConsumer` is deliberately not attempted here** — it sits
outside this PR's localized `src` boundary — so it is raised as #4299 instead. The new flag is a
workaround for that defect, not a statement of a platform limit, and the code comment says so: when
#4299 is fixed, drop the flag from `Kafka / Consumer` and let the default return to `true`.

### CI audit — no silent skips

A read-only, network-free audit (it never queries the issue tracker) enforces the deferral trail in
three ways:

- every `Skip` in a canonical template or a generated `Skip` must match `Deferred: #<n>` — a bare or
  reasonless `Skip` fails the build. Hand-written gateway tests are out of scope by design: they are
  not generated, so a `Skip` on one is a maintainer's explicit choice, not something the ledger
  emitted;
- every `Deferred` ledger cell must carry both an issue link and a recorded sign-off;
- every generated test's `Skip` must agree with **its own** (configuration × behaviour) ledger cell, in
  both directions — a cell flipped without regenerating fails, whether that leaves a test wrongly
  skipped or wrongly running. The expected value comes from the generator's own emitter, so the audit
  cannot drift from what generation produces.

The audits assert their own non-vacuity (files scanned, markers found, and that both branches were
exercised), and each failure mode is covered by a synthetic-tree test, so they cannot pass by checking
nothing.

### Out of scope
Fixing all brokers that don't currently implement Brighter support when there is no native support is out of scope for this work.

Every `Deferred` cell resolves to **#4240**, this PR's own tracking issue, by maintainer ruling — no
per-deferral follow-up issues are raised. The `#NNNN` pre-audit placeholders are fully reconciled: none
remains in the ledger or in any generated `Skip` marker.


iancooper:
## `aws-ci` on this PR is an account quota problem, not a conformance failure — and #4298 is the fix

Recording this so the red `aws-ci` here is not misread.

The 5 failures (all `AWS / SnsFifo` Reactor, including the basic `posting_a_message` companion) annotate as:

```
Amazon.SimpleNotificationService.Model.TopicLimitExceededException
  : Could not complete request: topic quota exceeded
```

That is a **shared AWS account** limit, not SnsFifo behaviour. The same job fails the same way on #4306, which changes nothing AWS-related. **#4298** has the diagnosis: the account had accumulated 26,381 orphaned resources, including exactly **1,000 FIFO SNS topics in `eu-west-1`** — the AWS default quota — because every generated gateway test leaked its topic and queue on every run.

⚠️ **No AWS conformance conclusion drawn from this PR's CI is trustworthy until #4298 has merged and the account has been reaped.** A `TopicLimitExceededException` during setup surfaces downstream as `Assert.NotEqual() … Actual: MT_NONE`, which is indistinguishable from a genuine behaviour failure.

### This PR conflicts with #4298 — 8 files, plus a trap

`git merge-tree` reports conflicts in all eight AWS gateway providers (`Sns{Fifo,Standard}`, `Sqs{Fifo,Standard}` × `AWS.Tests`, `AWS.V4.Tests`); both PRs edit `CreateSubscription`, `CleanUp` and `CleanUpAsync`. `ci.yml` does **not** conflict.

The trap, in full on [#4298](https://github.com/BrighterCommand/Brighter/pull/4298#issuecomment-5572916301): #4298's reaper tracks the topic, the channel and the DLQ, which is complete against `master` — but **this PR adds the invalid-message channel**, a resource type that does not exist at the branch point. Resolving the merge must wrap that name in `_reaper.TrackQueue(...)` too, or the merged result leaks precisely the resources this PR introduces.

**Suggested order: #4298 first**, then merge `master` here and resolve the eight files once.

### The other two red jobs are unrelated to AWS

- **`gcp-ci`** — down from 21 failures to 10 after `6c1fd91d2`. The remainder is led by FR-15 (`requeuing_with_zero_delay`), which is completely untouched by that fix because it goes through `ModifyAckDeadline(ackId, 0)` rather than the ack deadline. Under diagnosis.
- **`azure-ci`** — the ASB round trip fails in CI; all 11 ledger cells stay `Deferred` and `1aec68263` corrects the note explaining why. See the closed probe #4308.

iancooper:
@claude Please review this PR — but **scoped**, and mind your turn budget.

The label-triggered run ([35090214505](https://github.com/BrighterCommand/Brighter/actions/runs/35090214505)) exhausted all 40 turns and failed having ticked only its first checkbox: it never pulled the diff. This PR is 947 files, so an unscoped review cannot finish.

**Do not** read the generated tree, the PR body, or the ~20 prior review rounds. The design is settled — do not re-litigate it.

**737 of the 947 changed files are generated** from templates. Review only the following, in this order, and stop when your budget is half gone and post what you have:

**1. The 9 production files — the entire production risk of this PR:**

- `src/Paramore.Brighter.MessagingGateway.RMQ.Async/RmqSubscription.cs` — **read this one first.** It changes default queue durability, which changes behaviour for existing users.
- `src/Paramore.Brighter.MessagingGateway.MQTT/MQTTMessageConsumer.cs`, `MQTTMessageProducer.cs`, `MqttMessageConsumerFactory.cs`
- `src/Paramore.Brighter.MessagingGateway.AWSSQS/SnsMessageProducer.cs` and `src/Paramore.Brighter.MessagingGateway.AWSSQS.V4/SnsMessageProducer.cs` (a matched pair — check they agree)
- `src/Paramore.Brighter.MessagingGateway.GcpPubSub/GcpPullMessageConsumer.cs`
- `src/Paramore.Brighter.MessagingGateway.RocketMQ/RocketMqMessageProducer.cs`

**2. `tools/Paramore.Brighter.Test.Generator/Templates/MessagingGateway/Shared/*.liquid`** — three templates, each fanning out to ~24 generated copies, so a defect there is multiplied by 24.

**What to report:** only findings you would block a merge on, or that are plainly wrong. Correctness and resource lifetimes over style. Note explicitly if the two `SnsMessageProducer` copies have diverged. If a file is clean, say so in one line and move on.

A partial review that gets posted is worth far more than a complete one that never does.

iancooper:
Thanks — that was a useful review, and finishing inside the budget after the scoping made the difference. Disposition of all four findings below. Two were actioned, two did not survive verification, and I have set out the evidence for those rather than just disagreeing.

## Actioned

**1. Three undocumented behaviour changes — confirmed, fixed in `52abdf702`.**

You were right that `release_notes.md` had detailed sections for the RMQ durability and MQTT receive changes and nothing for these. Verified each against the diff before writing:

- **SNS `SendWithDelay`** now leads the new notes, for the reason you gave: the sync overload forwarded `TimeSpan.Zero`, so a delayed send published immediately, and honouring the delay makes the no-scheduler path reachable for the first time — a call that looked like it worked now throws `ConfigurationException`. The note calls that out explicitly as the upgrade hazard.
- **GCP `timeOut`** — noted, including that a null timeout leaves the client's own expiration in force and a `DeadlineExceeded` there is still logged and rethrown rather than reported as an empty receive.
- **MQTT producer tracing** — noted, including that the `InstrumentationOptions.All` default carries `RequestBody`, so message bodies reach producer spans.

**2. MQTT null-write / throw in the dispatch loop — confirmed, fixed in `715b5562b`.**

Also confirmed your note that it is pre-existing: the diff is only `_messageQueue.Enqueue(message!)` → `_messages.Writer.TryWrite(message!)`; the `Deserialize` and the `!` are untouched. Fixed anyway, since as you say this branch makes the handler the sole feed for a now-blocking `Receive`.

Deserialisation moved to an `internal static TryDeserialiseMessage` that returns null for both the null-document and malformed cases and logs which; the handler writes only a non-null message. The seam and the `InternalsVisibleTo` mirror `RocketMqMessageProducer.AddHeaderProperties` in this same branch, so the case is covered in `build` rather than only in `mqtt-ci`.

**One limit stated plainly:** removing the `catch` fails the malformed test with the real `JsonException`, so that half is proven. The **null** half of the test passes before and after the change — `Deserialize` already returned null — so the new protection there is the handler's `if (message is not null)`, which stays broker-only and uncovered.

## Not actioned, with evidence

**3. "an unrequested default change" — the framing does not hold; the residue is real and is now [#4365](https://github.com/BrighterCommand/Brighter/issues/4365).**

`WriteProducerEvent` is called by **13 of the 14** transport producers already — AWSSQS ×2, AWSSQS.V4 ×2, AzureServiceBus, GcpPubSub, Kafka, MsSql, Postgres, Redis, RMQ.Async, RMQ.Sync, RocketMQ, and `InMemoryMessageProducer`. MQTT was the only one without it. `instrumentationOptions` defaulting to `InstrumentationOptions.All` matches `RmqMessageProducer` (both halves) and `InMemoryMessageProducer`. So this brings MQTT into line with the convention; dropping the calls would make it the outlier again.

Your underlying observation is still right and is filed: all three construction sites are inside `MqttMessageConsumer`, none passes the argument, and there is no producer registry factory in the MQTT package — so the parameter is unreachable and `All` is unconditionally in force. On MQTT that is currently *unconfigurable*, which the other gateways are not. #4365 proposes threading it through the consumer.

**4. "only `Baggage` gained a guard" — this one is not accurate.**

`AddHeaderProperties` guards **nine** headers, not one: `Baggage`, `Type`, `Subject`, `DataSchema`, `CorrelationId`, `ReplyTo`, `DataRef`, `TraceParent` and `TraceState`.

The properties written unconditionally are `MessageId`, `Topic`, `HandledCount`, `MessageType`, `TimeStamp`, `Source`, `SpecVersion` and `ContentType`/`DataContentType` — all **required** CloudEvents fields, which is what the remark means by "every *optional* header has to be checked". The remark reads as accurate to me; happy to be shown otherwise if you disagree on `ContentType` specifically, which is the one arguable case.

---

@claude please re-review **against this feedback**, and keep it scoped — the label-triggered run before yours died on `Reached maximum number of turns (40)` without reading anything, so budget accordingly.

Read only these four files, which are the whole of `1efd77b86..715b5562b`:

- `src/Paramore.Brighter.MessagingGateway.MQTT/MQTTMessageConsumer.cs`
- `src/Paramore.Brighter.MessagingGateway.MQTT/Paramore.Brighter.MessagingGateway.MQTT.csproj`
- `tests/Paramore.Brighter.MQTT.Tests/When_a_payload_cannot_be_deserialised_should_be_dropped_rather_than_written.cs`
- `release_notes.md` (the three new `###` sections only — AWS SNS, GCP Pub/Sub, MQTT producer telemetry)

Four questions, in priority order:

1. Does `TryDeserialiseMessage` actually close the hole you raised? The catch is `JsonException` only. I probed ten payloads through the real `JsonSerialisationOptions` — `null`, malformed JSON, a bare string, a number, `[]`, a bad `messageType` enum, a bad `timeStamp`, a non-base64 body, a junk `source` URI, a non-GUID `messageId` — and **every** failure came back as `JsonException` (the last two deserialised successfully). So I did not widen it, on the grounds that a broader catch would swallow bugs. If you can name a payload that escapes as something else, that is the finding I most want.
2. Do the three release notes describe the behaviour changes **correctly**? They were written from the diff; contradict anything that misstates what the code does.
3. On finding 4 — do you accept the nine guards, or is there a specific unguarded header you think is genuinely optional?
4. Anything in the new `internal`/`InternalsVisibleTo` seam you would not ship.

Please do **not** re-review the rest of the PR, and do not restate findings already dispositioned above.

iancooper:
Note that we don't allow [InternalsVisibleTo] as tests MUST NOT couple to implementation details

iancooper:
@claude Please do a **fresh, broad review** of this PR. Read this whole comment as the prompt.

### Why I'm asking again

The last round ([run 35149743029](https://github.com/BrighterCommand/Brighter/actions/runs/35149743029)) was **scoped to one question** — it was titled "Scoped review — `InternalsVisibleTo`", and all four of its checklist items were about that. It found the thing I asked it to find and stopped. **That means everything else in this PR went unexamined in that round.** I want the search that finding cut short.

### `InternalsVisibleTo` is already fixed — do not re-report it

Commit `52334e83a` took **option 1 from your own list** ("promote the two methods to genuine public, documented API … first-class public members of a small purpose-built type"):

- **MQTT** — new `public static class MqttMessageCreator` with `CreateMessage(byte[] payload, object? topicPrefix)`. The consumer's arrival handler now calls it. It is the inverse of the already-public `MqttMessagePublisher.CreateMqttMessage`, and matches the existing public `KafkaMessageCreator`.
- **RocketMQ** — new `public static class RocketMqMessagePublisher` with `CreateRocketMqMessage(message, publication, delay, timeProvider)`. The **entire** builder block moved out of `RocketMqMessageProducer.SendWithDelayAsync`; `AddHeaderProperties` is now **private** inside the new type.
- Both `<InternalsVisibleTo>` lines are gone. `grep -rn "InternalsVisibleTo" src/ tests/` now returns only a prose comment in `SpannerBoxMigrationRunner.cs`.

**Do review that change adversarially** (see below) — just don't re-litigate the attribute itself.

### Scope — this is the part that makes the review finishable

The PR is ~949 changed files, but **736 of them are generated** and are not the review surface. **Skip every path containing `/Generated/`.** Also skip the PR body and the ~20 prior review rounds; I don't need them summarised back.

The real surface is:

- **`src/` — 10 files.** The two new types above, plus `MQTTMessageConsumer.cs`, `MQTTMessageProducer.cs`, `MqttMessageConsumerFactory.cs`, `RocketMqMessageProducer.cs`, `SnsMessageProducer.cs` (×2, AWSSQS and AWSSQS.V4), `GcpPullMessageConsumer.cs`, `RmqSubscription.cs`.
- **`tools/Paramore.Brighter.Test.Generator/` — 7 C# files** (`CanonicalBehaviours.cs`, `ConformanceLedger.cs`, `Generators/*.cs`, `Configuration/MessagingGatewayConfiguration.cs`) **and 29 `.liquid` templates**. The templates are the highest-leverage files here: one bug there is multiplied across all 736 generated files.
- **`tests/Paramore.Brighter.Test.Generator.Tests/ConformanceAudit/` — 19 audit files.** These are the guardrails; I'd like them checked for rules that can pass vacuously.
- **`tests/**/MessagingGateway/*Provider.cs` — 24 hand-written provider files.**

### What I actually want looked at

1. **Is the `RocketMqMessagePublisher` extraction behaviour-preserving?** Diff the new `CreateRocketMqMessage` against the block it replaced in `RocketMqMessageProducer.SendWithDelayAsync`. Ordering, the delay/`TopicType` conditions, the `Keys`/`Tag` fallbacks, `PartitionKey`. I believe it is identical; please try to prove otherwise.
2. **The new public API itself.** These are shipped packages. Naming, XML docs, nullability, and whether a `static` class is the right call. Note `MqttMessageCreator` deliberately keeps the `MqttMessageConsumer` logger category and the original log message text so existing log scraping doesn't break — tell me if that's the wrong trade.
3. **The `.liquid` templates and generator.** Anything that would produce subtly wrong tests across the tree.
4. **The audits** — can any of them pass while the thing they guard is broken?
5. **Anything the earlier rounds plausibly missed** because they were looking at a 12-file slice.

### Ground rules

- **Post a partial review rather than run out of road.** If you're running low, post what you have with a note on what you didn't reach. A partial review I can act on beats a complete one I never see.
- **Verify before you claim.** Two of four findings in an earlier round didn't survive checking (one was factually wrong — it claimed only `Baggage` had a guard when nine fields did). Quote the line you're relying on.
- Known and not worth reporting: `gcp-ci`/`azure-ci` are red for unrelated reasons, `rocketmq-ci` is commented out in `ci.yml`, and `postgres-ci` has an intermittent 1-of-258 failure in a hand-written test this PR doesn't touch.

iancooper:
## Reviewer guide: the product-code surface is 10 files, +532/−154

This PR is 963 files and ~102k lines, but **almost all of it is generated tests, templates and the conformance ledger**. The code that ships to users is this, and it is the whole of it:

```
git diff origin/master...HEAD -- src/
```

| file | ± | what to look at |
|---|---|---|
| `MessagingGateway.MQTT/MQTTMessageConsumer.cs` | +138 −25 | the largest behavioural change. `Receive` used to return immediately on an empty buffer, so every caller needed an external sleep. A `SemaphoreSlim` is now released per arriving message; `Receive` blocks on it and `ReceiveAsync` awaits it, honouring the caller's cancellation token. **Worth the closest read of anything here** — it is a blocking/async change in a pump path |
| `MessagingGateway.RocketMQ/RocketMqMessagePublisher.cs` | +158 −0 | new file, but **not new logic** — it is the message-building block lifted out of the producer below |
| `MessagingGateway.RocketMQ/RocketMqMessageProducer.cs` | +3 −102 | the other half of that extraction. The 102 removed lines become a 3-line call. ⚠️ **See the coverage caveat below** |
| `MessagingGateway.MQTT/MqttMessageCreator.cs` | +81 −0 | new; mirrors `KafkaMessageCreator` / `RmqMessageCreator` |
| `MessagingGateway.GcpPubSub/GcpPullMessageConsumer.cs` | +58 −6 | `Receive`/`ReceiveAsync` only — the requeue/reject paths are **untouched**. Binds the pull to the caller's `timeOut` via `CallSettings` expiration and maps `DeadlineExceeded` to an empty message. A null timeout deliberately yields null settings so the client's own per-method default stays in force |
| `MessagingGateway.AWSSQS/SnsMessageProducer.cs` | +37 −7 | sync `SendWithDelay` passed `TimeSpan.Zero` |
| `MessagingGateway.AWSSQS.V4/SnsMessageProducer.cs` | +37 −7 | ⭐ **the V3 and V4 hunks are byte-identical** — verified, so reviewing one reviews both |
| `MessagingGateway.MQTT/MQTTMessageProducer.cs` | +10 −1 | takes an `InstrumentationOptions` ctor parameter instead of always tracing at `All` |
| `MessagingGateway.RMQ.Async/RmqSubscription.cs` | +8 −4 | ⚠️ **the only user-visible behaviour change in the PR** — `isDurable` defaults `false` → `true`. See the callout in the PR description |
| `MessagingGateway.MQTT/MqttMessageConsumerFactory.cs` | +2 −2 | wiring for the above |

## Three things I would not want a reviewer to take on trust

**1. RocketMQ's 263 lines are barely executed.** `rocketmq-ci` is commented out in `ci.yml` (*"TODO: Rafael Andrade is working on how to run RocketMQ on GHA"*). This PR adds a `build`-job step for the broker-free tests, but that covers **3 of the project's 56 tests** and exercises the publisher seam only. The extraction was reviewed as behaviour-preserving by diffing the moved block — textually identical bar `mqPublication`→`publication` and `connection.TimerProvider`→the parameter — but **CI is not proving it**. The ledger's 9 `Fixed` RocketMQ cells rest on a one-off local run against a live RocketMQ 5.5.0 broker (21 pass / 2 skip per variant), not on anything repeatable.

**2. GCP now claims nothing.** All 24 GCP ledger cells are `Deferred -> #4240`. `gcp-ci` is green at **86 passed / 52 skipped** — green partly *by skipping*. 28 gateway tests do still execute (send, receive, activity-context propagation, multi-thread post), which is what covers the `Receive` change above; but GCP conformance for requeue/nack/reject is now unproven, and the root cause has its own issue, [#4321](https://github.com/BrighterCommand/Brighter/issues/4321). Real Pub/Sub does not redeliver promptly on `ModifyAckDeadline(ackId, 0)`; the earlier green cells came from the **emulator**.

**3. The bot review did not reach everything.** It read the FR-7, FR-16 and FR-23 templates in full and skimmed the rest of the 29; it spot-checked 3 of 24 provider files. **Absence of findings in the other templates and providers is not evidence about them.**

## Known-flaky jobs, both pre-existing

- [#4376](https://github.com/BrighterCommand/Brighter/issues/4376) — `postgres-ci`, hand-written gateway tests using a fixed sleep plus one read. `master` is 10/10 green; the branch adds load and tips them over. Neither failing test is touched by this PR.
- [#4330](https://github.com/BrighterCommand/Brighter/issues/4330) — `kafka-ci`, `Subscribed topic not available`. Reproduces on `master`.

A red run on either is not a signal about this PR.

PR #3996, opened 2026-01-29
Add MessageGateway test generator

lillo42:
# Overview

 Extends the \Paramore.Brighter.Test.Generator` tool to automatically generate standardized tests for messaging gateway implementations (RabbitMQ, AWS SNS/SQS, Azure Service Bus, etc.), reducing code duplication and
   ensuring consistent test coverage across all gateway implementations.

 # What's Changed

 ## Configuration System

- Added MessagingGatewayConfiguration class with support for:
	- Test class naming and categorization
	- Custom message factories and assertion helpers
	- Gateway-specific publication/subscription configuration
- Async operation timing controls
- Extended TestConfiguration with `MessagingGateway` and `MessagingGatewaies` properties
- Added comprehensive XML documentation for all configuration classes

 ## Generator Implementation

- New MessagingGatewayGenerator supporting both:
	- Reactor pattern (synchronous operations)
    - Proactor pattern (asynchronous operations)
- Follows the same architecture as the existing OutboxGenerator for consistency
- Generates provider interfaces and test implementations from Liquid templates

 ## Message Factory & Assertions

- Refactored message creation to use MessageConfiguration for complete control over test message properties
- New `IAmAMessageFactory` interface and `DefaultMessageFactory` implementation
- New `IAmAMessageAssertion` interface and `DefaultMessageAssertion` implementation
- Factories now track created messages via `CreatedMessages` property

##   Templates

- `MessagingGateway/Proactor/` - Async test templates
- `MessagingGateway/Reactor/` - Sync test templates
- Provider interfaces and assertion implementations

## Documentation

- Added ADR 0037 documenting the design decisions
- Added comprehensive XML documentation to all configuration classes and templates

## Benefits

- ✅ Eliminates code duplication across gateway implementations
- ✅ Ensures consistent test coverage for all messaging gateways
- ✅ Reduces maintenance burden when test patterns change
- ✅ Simplifies adding new gateway implementations with complete test coverage

PR #4298, opened 2026-09-01
fix(aws): stop generated gateway tests leaking SNS topics and SQS queues

iancooper:
## Description

CI could no longer create AWS topics or queues. The account had accumulated **26,381 orphaned resources** across two regions, including **exactly 1,000 FIFO SNS topics in `eu-west-1`** — the AWS default quota for FIFO topics per account per region. That quota is what CI was hitting, and it is why most of the leftovers were `.fifo`.

**The leak.** The generated MessageGateway tests never deleted the infrastructure they created. All eight AWS providers' `CleanUp`/`CleanUpAsync` purged and disposed the channel and producer, then stopped — no `DeleteTopicAsync`/`DeleteQueueAsync`. They also built a `ChannelFactory` inline and discarded it, so they held no handle to delete with. Every generated test leaked its topic and queue on every run, pass or fail. The hand-written fixtures and the scheduler tests already did this correctly and were not the problem.

**Why `clean_failed_tests_aws_assets.sh` never caught up with it:**

1. The tag query filters on `Environment=Test`, but the gateway only stamps `Source=Brighter` (`AWSMessagingGateway.cs:386`), so it returned nothing.
2. The naming fallback listed the hand-written fixtures' prefixes, which predate the test generator (#3996). **0 of 26,381** leaked names matched it.
3. `list-queues` was called without `--page-size`. SQS returns at most 1000 queues and no `NextToken`, so 20,900 queues looked like 1,000.
4. One AWS CLI call per resource could not clear a real backlog inside the cleanup workflow's timeout.

And `test_clean_failed_tests_aws_assets.sh` only ever created resources tagged `Environment=Test`, so it passed while the script found none of the actual leaks.

**The fix.** Providers now track every topic and queue name they hand out — including dead letter queues — and delete them from teardown via a new `AwsTestResourceReaper`. Reaping *names* rather than created objects also covers a test that failed while standing its infrastructure up, when there is no producer or channel left to dispose, and catches resources created as a side effect of a name (a DLQ, or a topic auto-created by a producer).

The teardown abstraction itself is not new — `CleanUpAsync` on `IAmAMessageGatewayProactorProvider` is exactly that hook. AWS simply was not honouring it. GCP and Kafka already delete their infrastructure through the same hook.

## Type of Change

- [ ] Documentation update

## Checklist

### Verification

- 27 generated AWS tests run against the live account → **0 resources left behind** (previously ~27 leaked).
- A test was deliberately forced to fail *after* a full message round trip → still **0 left behind**, which is the "cleans up even on a failed run" requirement tested rather than assumed.
- `test_clean_failed_tests_aws_assets.sh`: **24/24 pass**, including 4 new cases covering untagged, convention-named, FIFO resources. The existing negative test still proves untagged resources that do *not* match the convention survive.
- The 26,381 orphaned resources have been purged; all seven regions checked now report 0 topics and 0 queues.

The review-response commits have been verified as far as they can be locally: both projects build, and the offline reaper tests pass. The new live-account tests and the updated sweep test have not been run against AWS yet — CI is their first run.

### This PR carries a workflow-scoped change

`.github/workflows/aws-cleanup.yml` is in the diff: the cleanup workflow's `timeout-minutes` goes from 5 to 30, the diagnostic resource counts are fixed, and the sweep's age guard is wired to the trigger. Pushing this branch therefore needs a token with the `workflow` scope.

### Review response

Addressing https://github.com/BrighterCommand/Brighter/pull/4298#issuecomment-5493816881:

- **The reap now runs from a `finally`.** It was the last statement of each provider's `CleanUp`, after the channel purge and producer dispose — both of which reach AWS and can fail. `PurgeQueue` is throttled to one call per queue a minute, so a throw there was reachable under CI load, and it skipped the reap entirely.
- **`ReapAsync` can no longer throw past teardown.** Creating the SNS and SQS clients sat outside the guarded region, so a failure there replaced the test's own result — the opposite of what the method's summary promised. Each sweep guards its own client, and the tracked names are cleared in a `finally` so a half-completed run cannot leave names it will never retry.
- **The sweep no longer deletes queues a live run may be using.** Queues younger than `CLEANUP_MIN_AGE_SECONDS` (default an hour) are left for the next pass. The guard applies on *every* trigger, not only the six-hourly one: CI runs on every push and pull request and declares no concurrency group, so the run that just finished is never the only one in the account. What this defers is a leak from a run that died, by up to an hour — the sweep fires again on the next CI completion, and the schedule is only the floor.
- **Topic ARNs are composed from `RegionEndpoint.PartitionName` and `Amazon.Arn`**, the way `ValidateTopicByArnConvention` builds them, instead of reading the partition out of the caller identity's ARN by position. A failed STS lookup is now remembered rather than retried per topic.
- **The SQS providers no longer track a channel name that is never created.** `CreateSubscription` replaces it with the publication's queue, which `GetOrCreateRoutingKey` already tracks — so tracking it *after* the override would cost the same wasted round trip rather than removing it.
- **Diagnostic counts fixed.** `grep -c` prints `0` *and* exits 1, so the `|| echo "0"` fired too and the queue count came out as `Total queues: 0 0`; `|| true` is what was wanted. The topic counts had the same per-page `length(@)` undercount the queue count was already fixed for.
- **Test 8 polls to a deadline** instead of reading once, and takes its hex id from `uuidgen` rather than `openssl`.

Two items were not taken as written:

- `FindTopicAsync` keeps no cancellation token — the SDK offers no overload that takes one.
- Tracking the SQS channel name inside `CreateSubscription` would not have removed the round trip, for the reason above; the name is left untracked instead.

**New tests.** `MessageGatewayProviderCleanUpTests` stands infrastructure up through a provider and then looks for it in AWS after `CleanUpAsync`, so the reap is verified against the account rather than the fixture's bookkeeping; a second case tears down through a channel that throttles on purge and asserts both that the failure surfaces and that the topic and queue went anyway. Deletion is eventually consistent, so the assertions poll. `AwsTestResourceReaperTeardownFailureTests` covers the swallow-and-drain contract offline, with no credentials needed.

### Review response (round 3)

Addressing https://github.com/BrighterCommand/Brighter/pull/4298#issuecomment-5574971046:

- **The dependency commit is reverted to the AWS bump it was named for.** `Ensure using current AWS` also moved FluentValidation, two Google.Cloud packages, Grpc.Core.Api and MongoDB.EntityFrameworkCore, and added a `Pomelo.EntityFrameworkCore.MySql` entry to the net10.0 group. Two did harm: `FluentValidation 11.12.1` does not exist, so NuGet floated to 12.0.0, which dropped netstandard2.0 — `Paramore.Brighter.Validation.FluentValidation` failed to restore and **took the whole build with it**, which is why every test job in the last run was skipped rather than run. And the only project referencing Pomelo targets `net8.0;net9.0` precisely because Pomelo has no net10 support, so pinning it in the net10.0 group pulled 9.0.0 into four WebAPI samples' net10 restore graph against EF Core Relational 10.0.11 (NU1608). The `AWSSQSV4` bump stays; it lines the package up with the other AWS SDK pins, already on 4.0.100.11.
- **STS is asked once per run, not once per test.** The account id was cached on the reaper, and every generated test builds its own provider — and so its own reaper — in a constructor xUnit runs per test method, so "we ask STS once" meant ~100 round trips per project. Only success is cached: the reaper's own offline tests run in-process against a deliberately broken connection, and a shared failure flag would let them decide STS is unavailable for every other test in the run.
- **The `FindTopicAsync` fallback is gone.** It is the one call in this SDK surface taking no cancellation token, so the one call `DeleteTimeout` cannot bound, and it pages every topic in the account per topic — an unbounded scan on the path only reached when something is already wrong. (A note on the suggested `Task.WhenAny` race: it bounds the *wait* but does not cancel the call, which keeps running.) An unavailable account id now reports once and leaves those topics to the account sweep.
- **Every provider's tracking is covered, not just the one with a live test.** Naming a resource and creating one are separate steps, so the naming is asserted offline for all four providers per project — routing key, channel name and a dead letter queue the live test never sets up — with no credentials, so it runs in `aws-mock-ci`. Asserting against the *subscription* rather than the generated names is what earns it: both SQS providers deliberately discard the channel name they are handed and subscribe with the publication's queue, which was asserted only by a comment and is now executable. Verified non-vacuous by dropping a `TrackQueue` call and watching exactly that case fail.

One finding not taken: **STS against `floci` is not untested ground.** Run [33605421747](https://github.com/BrighterCommand/Brighter/actions/runs/33605421747), at `ea5bc425`, already contained the STS-calling reaper (`cd882b5e7` is its ancestor) and `aws-mock-ci` passed in **4m12s** against its 8-minute cap. The predicted cost — a failed STS call plus an unbounded `ListTopics` per test — did not materialise, and the two changes above remove the mechanism anyway.

That same run is the best argument for the `aws-ci` timeout raise: it passed in **7m55s** against an 8-minute cap, with the live teardown tests already present. Five seconds of margin.

### Two changes to CI's own defaults

Called out rather than left to be found in the diff:

- `aws-ci` and `aws-scheduler-ci` move from `us-west-2` to `eu-west-1`. The cleanup workflow has only ever swept `eu-west-1`, so CI's own region was never reaped — which is half of why this went unnoticed for so long.
- `aws-mock-ci` gains `--filter "LiveAWS!=true"`. The new teardown tests stand infrastructure up in a real account and then look for it; localstack cannot answer for them.

### Footgun worth knowing

The script and workflow operate on the ambient AWS region. The workflow pins `eu-west-1`, but local test runs leak into each developer's own default region — that is where a separate `us-east-1` pile came from. Documented in the script header rather than widening the workflow's scope.

### Review response (round 2)

Addressing https://github.com/BrighterCommand/Brighter/pull/4298#issuecomment-5573512293 and https://github.com/BrighterCommand/Brighter/pull/4298#issuecomment-5573515915:

- **Topics are age-guarded now, so the sweep cannot delete one out from under a live run.** This was the reviewer's blocking finding and it was right: the widened name pattern matches the bulk of the AWS suite, the sweep fires on every CI completion, and topics were guarded by nothing. SNS reports no creation time, so a topic's age cannot be read — only established. The first sweep to match a topic stamps it `BrighterSweepFirstSeen` and defers; a later sweep deletes it once the stamp is old enough. A topic created moments ago always survives its first sweep, and nothing is deferred forever. Needs `sns:TagResource` and `sns:ListTagsForResource`; losing `TagResource` would defer every topic indefinitely, so that failure is reported rather than swallowed. The self-contradicting comment in the script header is gone.
- **`CLEANUP_MIN_AGE_SECONDS` is validated.** With no `set -e`, a non-numeric value made `-gt 0` print a bash error and evaluate false — silently disabling the only guard between the sweep and a live run. It now refuses to run.
- **`--dry-run` previews the skips as well as the deletes.** The name sweep filtered by age *before* the listing, so a developer who ran the tests locally and then dry-ran to see what leaked saw nothing at all. The tag sweep had the inverse asymmetry. Both now filter in both modes and print `[DRY RUN] Would skip (too young): <name>`, and the trailing count reads `Matched N …, acted on M` rather than "Found" a number that was neither.
- **`Consumer-DLQ-Async` joins `TEST_PREFIXES`.** Its fixture deletes both queues in `Dispose`, but a throw in the constructor — two real `CreateAsyncChannel` calls — skips `Dispose`, which is exactly the case the name sweep exists for.
- **The reaper is safe to block on.** `Reap()` blocks on `ReapAsync`, and without `ConfigureAwait(false)` its continuations post back to the synchronisation context xUnit installs, bounded by `maxParallelThreads`. Enough threads blocked at once and no worker is left to run them. `ConfigureAwait(false)` throughout, and a remark on `Reap()` saying why, since it is easy to undo by accident.
- **The delete budget is per resource, not per sweep.** One 30-second budget covered STS plus every topic and queue; expiring mid-sweep cancelled the rest, and since the tracked names are dropped either way, nothing retried them.
- **Swallowed failures are reported.** A reaper that reaps nothing looked exactly like one that works — the invisibility this PR exists to end. Failures now name the resource on stderr, including the STS lookup whose fallback turns teardown quietly slow rather than loudly broken.
- **The regression test proves arrange before asserting absence.** It asserted only `QueueDoesNotExistException`/`NotFoundException`, which hold vacuously if nothing was ever created; a producer that stopped creating its topic eagerly, a wrong ARN or a drifted region would all have passed. Arrange now polls until `GetQueueUrl` and `GetTopicAttributes` both succeed.
- **`aws-ci` goes to 12 minutes**, since it runs both projects sequentially and each now carries two tests that poll a deletion to a deadline. The 90-second deadline stays: SQS's sixty-second visibility window is documented worst case, so shortening it would trade a timeout for a flake.

New sweep tests cover the guard directly: a young queue and topic are deferred rather than deleted, the dry run previews that deferral, the topic carries its first-seen stamp afterwards, a non-numeric guard exits non-zero having deleted nothing, and releasing the guard deletes what it deferred. The existing negative test still proves untagged, non-conventional resources survive.

Not taken: adding a `concurrency` group to `ci.yml` — a repo-wide change to every job, well outside this PR — and gating the topic sweep behind `workflow_dispatch`, which does not help, since CI runs on every push and a live run is as likely during the scheduled sweep as during a completion-triggered one. Left as follow-ups, both now filed: the sweep still reports deletion failures as warnings only and goes green regardless (#4315), and `brightertestbucket-<guid>` S3 buckets are swept by nothing — `S3LuggageUploadTests` never deletes its bucket at all (#4316).

Verified locally: both projects build, the offline reaper tests pass, and the sweep was exercised end to end against a stubbed AWS CLI — deferral, stamping, release once the stamp ages, dry-run symmetry, the validation refusal, and a denied `TagResource` all behave as described. The live-account tests and the updated sweep test still have CI as their first real run.

---

https://claude.ai/code/session_012BpmKpyGj1G6VZ2EibRZoq


lillo42 (approved):

iancooper:
Thanks — this was a good catch list. All of it is addressed in `f6d4d6e03..d2dbb077d`, except two items I did not take as written and one gap I have left open deliberately. Point by point:

**1. `Reap()` not in a `finally`.** Fixed in all eight providers, sync and async. You were right that this is the one that matters: the throttle on `PurgeQueue` makes it a reachable path, not a theoretical one, and it skipped the reap entirely.

**2. `ReapAsync` masking a test result.** Fixed. Client creation now sits inside the guarded region — each sweep guards its own client, so a failure creating the SNS client no longer skips the queues as well — and `_topics.Clear()`/`_queues.Clear()` moved into a `finally`, which makes the single-attempt contract real. The XML doc now says what the code does.

**3. The scheduled sweep vs. an in-flight run.** Age guard added: queues younger than `CLEANUP_MIN_AGE_SECONDS` are left for the next pass, with the lookups running at the sweep's parallelism and only against names that already matched. A queue whose age cannot be read is treated as old, since the likeliest reason is that it has already gone.

Two things the suggestion needed that were not obvious from the diff:

- The guard has to be off for the `workflow_run` path. CI has finished by then, nothing is in flight, and leaving it on would defer every leak from that run by six hours. The workflow now sets it per trigger — `3600` on `schedule`, `0` otherwise.
- `test_clean_failed_tests_aws_assets.sh` creates its fixtures seconds before invoking the sweep, so the guard would have skipped all of them and failed Test 8. It exports `CLEANUP_MIN_AGE_SECONDS=0`, since what that script tests is which resources are matched, not when the sweep declines to act.

**Smaller points.** `ResolveTopicArnAsync` now builds ARNs with `Amazon.Arn` from `RegionEndpoint.PartitionName`, the way `ValidateTopicByArnConvention` does, so STS supplies only the account id and the positional parse is gone. A failed identity lookup is remembered rather than retried per topic. Diagnostic counts fixed — `|| true` for the queue count (reproduced locally: the old form really did print `Total queues: 0 0`), and the topic counts had the same per-page `length(@)` undercount, so they count names now. Test 8 polls to a deadline and takes its hex id from `uuidgen`. The PR description now says the workflow-scoped change is in the diff.

**Not taken as written:**

- **`FindTopicAsync` cancellation token** — there is no overload that takes one in this SDK version; adding it fails the build with CS1501. Left as-is with a comment saying why.
- **Tracking the SQS channel name inside `CreateSubscription`** — after the override the effective name *is* the routing key's queue, which `GetOrCreateRoutingKey` already tracks. Tracking it there would cost the same `GetQueueUrl` and caught exception, just against an already-deleted queue instead of a never-created one. The name is simply left untracked, with a remark recording why.

**Test coverage.** Two new suites:

- `MessageGatewayProviderCleanUpTests` (live account) stands infrastructure up through a provider and then looks for it in AWS after `CleanUpAsync` — so the reap is verified against the account, not the fixture's own bookkeeping. The second case tears down through a channel that throws `PurgeQueueInProgressException` on purge, and asserts both that the failure still surfaces and that the topic and queue went anyway. Revert any of the eight `finally` blocks and it fails. Assertions poll to a deadline for exactly the eventual-consistency reason you raised about Test 8.
- `AwsTestResourceReaperTeardownFailureTests` covers the swallow-and-drain contract offline, through a connection whose client configuration fails, so point 2 is pinned on every build without credentials.

**One gap left open.** SNS reports no creation time, so topics cannot be age-guarded the way queues can and remain exposed to the scheduled sweep. Narrowing the scheduled path to the tag query for topics would close it, at the cost of making a leaked topic wait for the next CI completion. Recorded in the commit message and the PR description rather than fixed here.

CI is the first run for the live tests and the updated sweep test — they cannot run locally without credentials.

iancooper:
Good round. #1, #4 and #6 were all confirmed against the workflows before acting, and #1 was mine — your original suggestion was just "add an age guard"; disabling it for `workflow_run` was my addition, on reasoning that turns out to be false at the account level. Fixed in `1707fa106..ea5bc425b`.

**#1 — guard on every trigger.** Confirmed: no `concurrency:` in either `ci.yml` or `aws-cleanup.yml`, and CI fires on every push and every PR. `CLEANUP_MIN_AGE_SECONDS` is now `3600` unconditionally rather than per-trigger. Your point that this defers nothing is what makes it free: the providers reap their own resources now, so the sweep is only a backstop for runs that die before they can, and those are caught by the next pass.

**#4 — LocalStack.** Confirmed, and worse than the timeout: `CredentialsChain` returns `BasicAWSCredentials` and `USEast1` whenever `AWS_SERVICE_URL` is set, so the live tests were pointed at floci with fake credentials. They now carry `[Trait("LiveAWS", "true")]` and `aws-mock-ci` filters `LiveAWS!=true`. Verified both directions locally: the mock filter lists no tests from that class, `aws-ci`'s `Fragile!=CI` still lists both. A runtime skip was the alternative and I didn't take it — a test that passes without asserting is its own hazard.

**#6 — region.** Confirmed, and it is the finding with the longest reach. On CI there is no profile store, so `CredentialsChain` falls through to `new EnvironmentVariableAWSRegion().Region`, which reads `AWS_REGION` — and the job-level `env` does win over what `configure-aws-credentials` exports to `$GITHUB_ENV`. The suites have been running in `us-west-2` while the sweep swept `eu-west-1`, which means the backstop has never once looked where `aws-ci` leaks. `aws-ci` and `aws-scheduler-ci` now name the region their credentials step already named. That is a real change to where the AWS suites run, so it is worth a second pair of eyes.

**#3 — the NRE.** Confirmed. The ARN is composed from `RegionEndpoint.PartitionName` and the caller identity, the way `ResolveTopicArnAsync` does, so the eventually-consistent `ListTopics` scan is gone along with the null.

**#2 — the leaking leak test.** Fixed with `IAsyncLifetime.DisposeAsync` calling `CleanUpAsync(null, null, [])`, exactly as you suggested and for the reason you gave.

**#5 — no timeout.** `ReapAsync` now applies a thirty second timeout when the caller supplies no cancellable token. No test: provoking a hung SDK call is more machinery than the guarantee is worth, so it is asserted by construction rather than by test.

**Nits taken:** the teardown-throws test disposes the channel and producer the throwing purge left behind; `AwsTestResourceReaper` records that tracking is unsynchronised; the sweep header notes the `TEST_PREFIXES` / `GENERATED_TEST_PATTERN` asymmetry you describe; `hex_id` falls back to `od -An -tx1 -N16 /dev/urandom` — you were right that the comment claimed more than `uuidgen` delivers.

**Declined:** the `static Lazy` for STS, keeping the reaper self-contained as you offered.

---

Two I'd push back on.

**#7 — one test method per file.** `testing.md:9` is followed by `testing.md:10`: *"If you decide to use multiple test cases per file, for example shared complex set up, name the file after the happy path test method and the class after the shared behavior."* Both files have shared complex setup — a constructor arrange and a `CreateInfrastructureAsync` helper in one, a shared failing-connection fixture in the other — are named after their happy-path method, and carry `[Behavior]Tests` class names, which you separately note is correct. That is the documented exception rather than a violation of the rule above it.

**The `Category=AWS` nit.** It is the most-used category in the project, not one nothing else uses: 74 occurrences in `Paramore.Brighter.AWS.Tests` against 36 `Sns` and 40 `Sqs`. `[Collection("SnsStandard")]` I have also left off deliberately — the fixture's names are GUID-based and shared with nothing, so joining that collection would serialise it against unrelated tests for no isolation gained.

CI is again the first real run for the live tests; the offline reaper tests and both filter directions are verified locally.

iancooper:
@claude Let's re-review the latest, and look at the comments

iancooper:
Heads-up from #4297 (the transport-conformance branch), which this PR collides with. Two things, one of them a trap.

## 1. The conflict

`git merge-tree` against `master` (`cd50a43a9`) reports **8 conflicting files** — every AWS gateway provider, in both projects:

```
tests/Paramore.Brighter.AWS.Tests/MessagingGateway/{SnsFifo,SnsStandard,SqsFifo,SqsStandard}MessageGatewayProvider.cs
tests/Paramore.Brighter.AWS.V4.Tests/MessagingGateway/{SnsFifo,SnsStandard,SqsFifo,SqsStandard}MessageGatewayProvider.cs
```

Both PRs edit the **same three methods** — `CreateSubscription`, `CleanUp`, `CleanUpAsync`. Per file, #4297 adds ~107 lines (the new conformance surface) and this PR ~38 (the reaper).

Good news: **`.github/workflows/ci.yml` does *not* conflict.** This PR's `AWS_REGION: us-west-2 -> eu-west-1` and `--filter "LiveAWS!=true"` edits sit in different hunks from #4297's `timeout-minutes: 8 -> 20` on `aws-ci`/`aws-mock-ci`, so the two merge cleanly.

## 2. 🪤 The trap: resolving the conflict "correctly" still reintroduces a leak

The reaper wraps every name the provider hands out — `TrackTopic` for the topic, `TrackQueue` for the channel and for the DLQ. That is complete **against `master`**, where the signature is:

```csharp
public SqsSubscription CreateSubscription(
    RoutingKey routingKey, ChannelName channelName,
    OnMissingChannel makeChannel, bool setupDeadLetterQueue = false)
```

#4297 replaces that last parameter with two routing keys and introduces **a resource type that does not exist on `master`: the invalid-message channel**.

```csharp
public SqsSubscription CreateSubscription(
    RoutingKey routingKey, ChannelName channelName, OnMissingChannel makeChannel,
    RoutingKey? deadLetterRoutingKey = null,
    RoutingKey? invalidMessageRoutingKey = null)
```

and then stands up a real channel for it:

```csharp
subscriptionName: new SubscriptionName(subscription.InvalidMessageRoutingKey!.Value),
channelName:      new ChannelName(subscription.InvalidMessageRoutingKey!.Value),
```

That is another SQS queue per test, created on every run. This PR cannot know about it — it does not exist on the branch point.

**So a resolution that looks right — keep #4297's expanded signature, keep this PR's tracking — still leaks.** Whoever resolves the merge must also wrap #4297's invalid-message routing key and channel name in `_reaper.TrackQueue(...)`, or the quota problem returns via the resources #4297 adds, and the `FifoTopicLimitExceeded` failures come straight back.

## 3. Suggested merge order: **this PR first**

- It fixes the root cause that is currently red on **both** PRs — `aws-ci` fails on #4306 too, and that PR changes nothing AWS-related.
- #4297 takes AWS generated files from **168 to 312**. Merging it first roughly doubles the leak rate against an account already at the 1,000 FIFO-topic quota.
- Landing this first also means #4297 resolves the 8 files once, with the reaper pattern already in the tree to copy — rather than this PR having to rebase onto a doubled surface.

Nothing here asks for a change to this PR. Recorded so it does not get lost between the two.

PR #4300, opened 2026-09-04
fix: generate the Sync outbox suite for a singular Outbox configuration

### LINKED ISSUE #4238: Test Generator: single `Outbox` config generates Async tests only, silently skipping the Sync suite
## Summary

In `tools/Paramore.Brighter.Test.Generator`, a `test-configuration.json` using the singular `Outbox` form generates **only the Async outbox suite**. The Sync suite is never emitted, and nothing reports this. The plural `Outboxes` form correctly generates both.

Projects using the singular form therefore have async-only outbox conformance coverage today: `Paramore.Brighter.DynamoDB.Tests` and `Paramore.Brighter.MongoDb.Tests`.

## Evidence

`Generators/OutboxGenerator.cs:49-61`:

```csharp
if (configuration.Outbox != null)
{
    await GenerateAsync(configuration, "Outbox", "Outbox", configuration.Outbox);

    var prefix = configuration.Outbox.Prefix;
    await GenerateAsync(
        configuration,
        Path.Combine("Outbox", prefix, "Generated", "Async"),
        Path.Combine("Outbox", "Async"),
        configuration.Outbox,
        filename => SkipTest(configuration.Outbox, filename)
    );
}
```

Two distinct problems:

1. **No Sync call.** Only the `Async` template folder is rendered. Compare the `Outboxes` branch (lines 62-89), which renders `Outbox/Sync` and `Outbox/Async`.
2. **The first call is a no-op.** It passes template folder `"Outbox"`. `Templates/Outbox/` contains only the `Sync/` and `Async/` subdirectories and no `.liquid` files of its own, and `BaseGenerator` enumerates with `SearchOption.TopDirectoryOnly`, so it renders nothing. It looks like it was meant to be the Sync call.

Additionally, unlike the `Outboxes` branch, the singular branch never defaults `Prefix` from a key and never dot-prefixes it, so a single `Outbox` with no explicit `Prefix` emits path `Outbox/Generated/Async/` with namespace `<Namespace>.Outbox.Async`.

## Expected

The singular `Outbox` form should generate both the Sync and Async suites, consistent with `Outboxes`.

## Suggested fix

Replace the no-op call with the missing Sync call:

```csharp
await GenerateAsync(
    configuration,
    Path.Combine("Outbox", prefix, "Generated", "Sync"),
    Path.Combine("Outbox", "Sync"),
    configuration.Outbox,
    filename => SkipTest(configuration.Outbox, filename)
);
```

Note this is a behaviour change for DynamoDB and MongoDB: they will gain a Sync suite that has never run, and it may not pass immediately. Worth landing behind a look at whether those providers implement `IAmAnOutboxSync` at all.

## Workaround

Use `Outboxes` with a single entry, which generates both suites.

## Notes

Found while documenting the generator for `docs/factories/tests`. The workaround is documented there, and this issue is linked from that page.

mazyn:
## Description

A `test-configuration.json` using the singular `Outbox` form generated only the Async and Causation suites. The Sync suite was never emitted, and nothing reported it. The plural `Outboxes` form was unaffected.

The Sync call was dropped from the singular branch by a formatting pass in f16a0c896 (#3996). It left behind a vestigial call passing template folder `"Outbox"`:

```csharp
await GenerateAsync(configuration, "Outbox", "Outbox", configuration.Outbox);
```

`Templates/Outbox/` contains only the `Sync/`, `Async/` and `Causation/` subdirectories and no `.liquid` files of its own, and `BaseGenerator` enumerates with `SearchOption.TopDirectoryOnly` — so that call rendered nothing. The same commit correctly removed the equivalent no-op from the `Outboxes` branch, which is why only the singular form regressed.

This replaces the no-op with the missing Sync call, so the singular branch now mirrors the plural one: Sync, then Async, then Causation.

## Related Issues

Fixes #4238

## Type of Change

- [ ] Documentation update

## Checklist

- [x] I have added/updated XML documentation for any public API changes — no public API changed; the existing class summary already documented "producing synchronous and asynchronous test variants" and is now accurate

### This does not change any test outcomes

The affected projects are **not** async-only today. `Outbox/Generated/Sync/` is committed in all three, and `DynamoDBOutboxProvider` and `MongoDbOutboxProvider` both implement `IAmAnOutboxProviderSync` — so those Sync suites already compile and run. They were *orphaned* from the generator: still executing, but no longer refreshed, and free to drift silently as the templates evolved.

Regenerating all three projects after this change produces a **zero diff**. Every emitted file — Shared helpers, Outbox Sync, Async and Causation — is byte-identical to what is already committed. No provider infrastructure (DynamoDB Local, Mongo) is needed to review this, because the generated sources cannot change.

### Affected projects

- `Paramore.Brighter.DynamoDB.Tests`
- `Paramore.Brighter.DynamoDB.V4.Tests`
- `Paramore.Brighter.MongoDb.Tests`

### Validation performed

| Check | Result |
| --- | --- |
| New test against `master` | Fails: `a singular Outbox should generate the Sync suite` |
| Generator suite after fix | 20/20 passed |
| Generate from `master` worktree into scratch dir | `Async`, `Causation` — 18 files, no `Sync` |
| Generate from this branch into scratch dir | `Async`, `Causation`, `Sync` — 31 files |
| Regenerate all three projects in place | Zero diff — `git status` clean under `tests/` |
| `dotnet csharpier check` on both changed files | Clean |

### Deliberately out of scope

- **`Prefix` handling is untouched.** Unlike the `Outboxes` branch, the singular branch never defaults `Prefix` from a key nor dot-prefixes it. That is inert for all three affected projects, since none sets a `Prefix`. There is a latent bug here — a singular `Outbox` *with* an explicit `Prefix` renders the namespace `<Namespace>.OutboxFoo.Sync`, missing the dot — but fixing it would relocate files and namespaces, so it belongs in its own change.
- **The duplication between the two branches is left alone.** Extracting the shared Sync/Async/Causation triple would prevent exactly this class of regression, but it is a structural change and belongs in a separate Tidy First commit.

### Suggested follow-ups

1. `When_generating_with_multiple_outboxes_should_complete_without_error.cs` contains a copy-paste of the *single*-outbox arrangement (it sets `Outbox`, not `Outboxes`), so the plural path has no coverage there.
2. The singular-`Outbox`-with-`Prefix` namespace bug described above.
3. No CI job runs `generate-test.sh` and fails on a dirty tree. That check is what would have caught this regression at the time, and would catch the next one.


tools/Paramore.Brighter.Test.Generator/Generators/OutboxGenerator.cs
```diff
@@ -48,9 +48,15 @@ public async Task GenerateAsync(TestConfiguration configuration)
...
             var prefix = configuration.Outbox.Prefix;
+            await GenerateAsync(
+                configuration,
+                Path.Combine("Outbox", prefix, "Generated", "Sync"),
+                Path.Combine("Outbox", "Sync"),
+                configuration.Outbox,
+                filename => SkipTest(configuration.Outbox, filename)
+            );
+
             await GenerateAsync(
```
iancooper, line 60: The "Async" call should also probably pick up path:

```
var prefix = configuration.Outbox.Prefix;
    await GenerateAsync(
        configuration,
        Path.Combine("Outbox", prefix, "Generated", "Async"),
        Path.Combine("Outbox", "Async"),
        configuration.Outbox,
        filename => SkipTest(configuration.Outbox, filename)
```

iancooper: The original issue does suggest this code in the comment, but it's unclear that both need change

iancooper:
Thanks @mazyn. I think we need the fix around the Async path too

iancooper:
Note on **suggested follow-up 3** ("no CI job runs `generate-test.sh` and fails on a dirty tree"): that check is worth having, but it would **not** have caught this bug, and it will not catch the next one of this shape.

A dirty-tree check catches drift in files the generator **writes**. It is structurally blind to a file the generator **stopped writing**. Regeneration does not delete anything, so an orphaned file just sits there and the tree stays clean — which is exactly what happened here. The 37 `Outbox/Generated/Sync/` files across the three affected projects have been orphaned since f16a0c896, and a regenerate-and-diff job would have passed on every commit in between. (It stayed benign only because `Templates/Outbox/Sync/` happens not to have changed since; `git log f16a0c896..HEAD -- tools/Paramore.Brighter.Test.Generator/Templates/Outbox/Sync` is empty. Any edit to those templates would have silently divorced the three projects from them.)

The complementary check is an **expected-set audit**: compute the set of files the current configuration would produce, and compare it with what is on disk in both directions.

I modelled that set — mirroring `BaseGenerator`'s template enumeration, both `SkipTest` predicates, and the prefix logic — and diffed it against the tree:

| | expected | on disk | orphans (on disk, not expected) | missing (expected, not on disk) |
|---|---|---|---|---|
| `master` today | 1259 | 1298 | **39** | 0 |
| with this PR | 1296 | 1298 | **2** | 0 |

The 39 are the 37 `Sync` files this PR reclaims, plus one pre-existing pair unrelated to it: `tests/Paramore.Brighter.RocketMQ.Tests/MessagingGateway/Generated/{Reactor,Proactor}/When_posting_a_message_with_partition_key_via_the_messaging_gateway_should_be_received.cs`. No `.liquid` of that name has ever existed (checked across all of history, not just deletions) — it was hand-written into a `Generated/` directory under an `<auto-generated>` header at #3996, and it asserts a partition-key round trip that no current template covers.

Two things follow:

1. **The "missing" direction is the one that matters here.** Under this PR's own fix it reads 0, but against `master` before the fix it would have reported the 37 absent `Sync` files — i.e. it would have failed CI the day #3996 landed, which is precisely what #4238 needed and did not have.
2. **This PR makes such an audit introducible green.** After it merges the orphan count drops to 2, both pre-existing and both traceable, so the check can go in as a passing guard rather than a red one.

So: worth adding as **follow-up 4**, distinct from follow-up 3 rather than a variant of it. The two assertions are different — one guards content, the other guards membership — and neither subsumes the other.

Nothing here asks for a change to this PR.

iancooper:
Correction to the table above: the row I labelled `master` was measured on the #4297 branch, not on `master`. I have since measured `master` (`cd50a43a9`) directly:

| tree | expected | on disk | orphans | missing |
|---|---|---|---|---|
| `master` (`cd50a43a9`) | 741 | 780 | **39** | 0 |
| PR #4297 branch (`f80127f04`) | 1259 | 1298 | **39** | 0 |
| PR #4297 branch, with this PR applied | 1296 | 1298 | **2** | 0 |

The totals differ because #4297 adds templates and wired configurations. **The orphan set is the same 39 paths in both trees**, so nothing in the point above changes — including that this PR accounts for 37 of them and that the remaining 2 are the pre-existing RocketMQ pair.

Tracking the guard as #4305.

iancooper:
@mazyn Ran some checks and going to merge

PR #4308, opened 2026-09-07
PROBE (do not merge): observe the ASB canonical suite in CI

iancooper:
# ⚠️ PROBE — DO NOT MERGE, DO NOT REVIEW

**This branch deliberately carries false ledger values.** It exists to make `azure-ci` execute the Azure Service Bus canonical conformance suite exactly once, so its real results can be observed. It will be closed unmerged as soon as the run reports.

## Why

The conformance ledger defers **all 11** Azure Service Bus cells, on the recorded grounds that:

> **this is a deferral of VERIFICATION, not a declaration of non-conformance** — no ASB behaviour has been observed to fail; none has been observed at all.
>
> …the credentials `ASBCreds.cs` requires (`BrighterTestsASBConnectionString` / `BrighterTestsASBNameSpace`) are **both unset** … so every canonical test would fail on client construction rather than on a behaviour.

That is accurate for a local machine. **It is not accurate for CI.** `.github/workflows/ci.yml` sets:

```yaml
  azure-ci:
    env:
      BrighterTestsASBConnectionString: ${{ secrets.BRIGHTERTESTS_ASB_CONNECTION_STRING }}
```

So `azure-ci` already runs against a real namespace — `iancooper-lisbon.servicebus.windows.net` appears in its failure output. The ledger itself says what would close these deferrals:

> **To close these deferrals**: supply `BrighterTestsASBConnectionString` … against a real namespace … flip the cells, and regenerate.

CI already satisfies the precondition. The evidence is reachable.

## The cycle this breaks

It is reachable but not *obtainable*, because the two ends depend on each other:

- a `Deferred` cell makes the generator emit `[Fact(Skip = "Deferred: #4240 …")]`
- a skipped test is never executed, in CI or anywhere
- so CI can never produce the observation that would justify changing the cell

Flipping the cells is the only way to cut that loop. Hence a throwaway probe rather than a change to #4297.

## What is in this branch

The 11 ASB cells set to `Pass`, and the suite regenerated — which removes the `Skip` from **22** generated tests (11 behaviours × Reactor/Proactor).

Regeneration touched **nothing else**: `git status` after `./generate-test.sh` showed exactly 22 ASB files plus the ledger row. No other project's generated output differs by a byte, which also re-confirms the generator is deterministic with respect to everything except the cell that changed.

`dotnet build` on the ASB project: **0 errors**.

## What happens next

1. Read the per-behaviour pass/fail from this PR's `azure-ci` job.
2. Write the **observed** values into the ledger on `feature/4240-universal-transport-conformance-tests` — `Pass` only where both variants actually passed, `Deferred` with a real reason everywhere else, per the ledger's FR-14 rule.
3. Close this PR **unmerged** and delete the branch.

## Known independent problem

`azure-ci` on #4297 also shows **6 deterministic failures** in the *non-canonical* companions — basic post/receive, multi-message, multi-thread post — with `ServiceBusException: SubCode=40901 / 40900, Status: 409 (Conflict)` on topic-and-subscription management. Those are not ledger-gated and so were never skipped. They are tracked separately; the canonical results read from this probe must be interpreted with that contention in mind, since a 409 during setup can present as a behaviour failure.


iancooper:
## Probe complete — closing unmerged, as planned

`azure-ci` ran the un-skipped canonical suite once. **161 tests, 131 passed, 30 failed.**

All 11 canonical behaviours failed, both variants — **and so did the basic post/receive companion**. 52 of the failures are `Assert.NotEqual() Failure: … Actual: MT_NONE`, alongside `ServiceBusException: SubCode=40900 / 40901, Status: 409 (Conflict)` on topic-and-subscription management.

### The result is inconclusive, and no cell can be flipped from it

This is the caveat the probe was opened with, and it is what happened. A transport that cannot complete a **plain round trip** in this environment cannot have "does Nack redeliver" judged against it — every canonical failure is downstream of the round trip rather than evidence about the behaviour. The 409s point at contention on the shared namespace, not per-behaviour non-conformance.

So all eleven cells stay `Deferred -> #4240`.

### What the probe did establish

1. **The recorded reason for the deferral was wrong.** The ledger said no ASB behaviour "has been observed at all" because the credentials are unset. CI has them and has been running against a real namespace all along. Corrected on the #4240 branch in `1aec68263` — the cells are untouched, only the note.
2. **The failure is foundational, not per-behaviour.** Worth knowing before anyone tries to certify ASB again: fix the round trip first, and the 409 contention with it.
3. **One distinct defect**, the only ASB failure with a cause of its own: both variants of `When_sending_a_delayed_message_should_deliver_after_delay` fail with `System.UriFormatException: Invalid URI: The format of the URI could not be determined.` That is a real fault in the delayed-send path, not a missed message.

### Closing

Doing exactly what the description said: closed unmerged, branch deleted. Nothing here was ever intended to land, and the false cell values it carried die with it.

Issue #4309, opened 2026-09-07
Generated Azure Service Bus conformance tests never receive: the subscription is created lazily, after the send


iancooper:
## Summary

The generated Azure Service Bus conformance tests (#4240) never receive the message they send. The cause
is an ordering gap: **`AzureServiceBusChannelFactory` provisions no infrastructure**, and the ASB
subscription is instead created lazily on the first *receive*. The test sends before that happens, and an
ASB topic with no subscription attached silently **discards** the message.

## The mechanism

1. `AzureServiceBusChannelFactory.CreateAsyncChannel`
   (`src/Paramore.Brighter.MessagingGateway.AzureServiceBus/AzureServiceBusChannelFactory.cs:75-87`)
   constructs a `ChannelAsync` around a consumer and returns. It creates nothing on the broker.
   `CreateAsyncChannelAsync` (line 97-99) is just `Task.FromResult(CreateAsyncChannel(...))`.
2. The subscription is created inside `AzureServiceBusTopicConsumer.EnsureChannelAsync`
   (`AzureServiceBusTopicConsumer.cs:79`, the `CreateSubscriptionAsync` call is at `:98`), which runs
   only from `AcknowledgeAsync` (`AzureServiceBusConsumer.cs:106`), `ReceiveAsync` (`:177`),
   `NackAsync` (`:243`), `RejectAsync` (`:300`) and `PurgeAsync`
   (`AzureServiceBusTopicConsumer.cs:76`). There is no sync `EnsureChannel` — the abstract member at
   `AzureServiceBusConsumer.cs:372` is async-only and every sync entry point is sync-over-async, so
   both pumps hit the same lazy path.
3. Every generated test does: `CreateProducerAsync` → `CreateChannelAsync` (creates **nothing**) →
   **send** → receive. Note that `CreateProducerAsync` does **not** create the topic either:
   `AzureServiceBusMessageProducerFactory.CreateAsync` is `Task.FromResult(Create())` (`:96-99`) and
   `Create()` (`:67-94`) only `new`s up producer instances. The **topic** is created lazily during
   the send, via `SendWithDelayAsync` → `GetSenderAsync` → `EnsureChannelExistsAsync`
   (`AzureServiceBusMessageProducer.cs:248`), and
   `AzureServiceBusTopicMessageProducer.EnsureChannelExistsAsync` (`:61-88`) creates the topic
   **only**, never a subscription.
[...]

PR #4311, opened 2026-09-07
fix(asb): accept a relative CloudEvents source and dataschema

### LINKED ISSUE #4310: Azure Service Bus gateway cannot receive a message with a relative CloudEvents source (UriFormatException)
## Summary

`AzureServiceBusMessageCreator` constructs a `System.Uri` from the CloudEvents `source` and `dataschema`
attributes using the **absolute-only** `new Uri(string)` overload. CloudEvents defines `source` as a
[URI-reference](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#source-1), which **MAY be
relative**. Any Brighter message carrying a relative source therefore cannot be received over Azure Service
Bus — the receive throws `System.UriFormatException` and the message is never delivered to the pipeline.

## The offending code

`src/Paramore.Brighter.MessagingGateway.AzureServiceBus/AzureServiceBusMessageCreator.cs`

- `GetSource` (line 309) ends with `return new Uri(source!);` (line 331)
- `GetCloudEventsDataSchema` (line 151) ends with `return new Uri(dataSchema);` (line 173)

Both methods already handle the *missing* and *empty* cases by falling back to a default
(`new Uri("http://goparamore.io")`) — they just don't handle the *present but relative* case.

The write side is fine: `AzureServiceBusMessagePublisher.cs:87` writes
`message.Header.Source?.ToString()`, which round-trips a relative URI to a bare string correctly. It is
only the read side that cannot parse it back.

## Observed failure

From the `azure-ci` job of the (now closed) probe PR #4298-era run
[`34127954869` / job `101763183832`](https://github.com/BrighterCommand/Brighter/actions/runs/34127954869):

```
System.UriFormatException : Invalid URI: The format of the URI could not be determined.
   at System.Uri..ctor(String uriString)
   at Paramore.Brighter.MessagingGateway.AzureServiceBus.AzureServiceBusMessageCreator.GetSource(IBrokeredMessageWrapper)
      in /_/src/Paramore.Brighter.MessagingGateway.AzureServiceBus/AzureServiceBusMessageCreator.cs:329
   at Paramore.Brighter.MessagingGateway.AzureServiceBus.AzureServiceBusMessageCreator.MapToBrighterMessage(IBrokeredMessageWrapper)
      in .../AzureServiceBusMessageCreator.cs:85
   at Paramore.Brighter.MessagingGateway.AzureServiceBus.AzureServiceBusConsumer.ReceiveAsync(Nullable`1, CancellationToken)
      in .../AzureServiceBusConsumer.cs:220
   at Paramore.Brighter.ChannelAsync.ReceiveAsync(Nullable`1, CancellationToken) in .../ChannelAsync.cs:153
```

It surfaced in both the Reactor and Proactor variants of
`When_sending_a_delayed_message_should_deliver_after_delay`.

## Why this matters beyond one test

The generated transport-conformance test suites (#4240) build their messages with a shared template,
`tools/Paramore.Brighter.Test.Generator/Templates/DefaultMessageBuilder.cs.liquid`, which defaults the
source to a relative URI:

```csharp
private Uri? _source = new Uri(Uuid.NewAsString(), UriKind.Relative);
```

**All 17 generated `DefaultMessageBuilder.cs` copies carry that same default.** Every other transport
receives such a message without complaint; Azure Service Bus is the only one that throws. That asymmetry
is what identifies this as a gateway defect rather than a test artefact.

## Suggested fix

Use `Uri.TryCreate(value, UriKind.RelativeOrAbsolute, out var uri)` in both methods and fall back to the
existing `http://goparamore.io` default when parsing fails, mirroring how the missing/empty branches
already behave. Fix **both** methods — otherwise the first message with a relative `dataschema`
reproduces the identical failure.

## Note

This defect is currently **masked** by #4309 (the generated ASB tests never receive a message at all,
because the subscription is created lazily after the send). Fixing that one is what exposes this.

Found while investigating why the Azure Service Bus conformance suites were deferred in #4240.

iancooper:
Fixes #4310.

## The bug

CloudEvents defines both `source` and `dataschema` as URI-**references**, which may be relative.
`AzureServiceBusMessageCreator` built them with the single-argument `new Uri(string)` overload, which
defaults to `UriKind.Absolute` and throws `UriFormatException` on anything relative. The guards above
each call covered *absent* and *empty* but never *present-but-relative*.

The failure lands in the consumer's receive loop, not at a boundary where it could be reported — the
message is simply undeliverable:

```
System.UriFormatException : Invalid URI: The format of the URI could not be determined.
   at System.Uri..ctor(String uriString)
   at AzureServiceBusMessageCreator.GetSource(IBrokeredMessageWrapper)
   at AzureServiceBusMessageCreator.MapToBrighterMessage(IBrokeredMessageWrapper)
   at AzureServiceBusConsumer.ReceiveAsync(Nullable`1, CancellationToken)
   at Paramore.Brighter.ChannelAsync.ReceiveAsync(...)
```

**Azure Service Bus was the only gateway with this defect.** An exhaustive scan of `src/` for a
single-argument `new Uri(...)` applied to wire data found exactly two hazardous sites, both in ASB.
Every other transport — RMQ (both), SQS (+V4), GCP Pub/Sub, Kafka, Redis, RocketMQ — already parses
wire URIs with `Uri.TryCreate(..., UriKind.RelativeOrAbsolute, ...)`, as do all four Outbox
implementations. This brings ASB into line with that existing house pattern rather than inventing one.

It surfaced through the generated transport conformance suites (#4240), whose shared message builder
defaults the source to a bare GUID. All 17 generated `DefaultMessageBuilder.cs` copies carry that
identical default, and only ASB threw — that asymmetry is what identifies the gateway, not the
builder, as the defective side.

## Also fixed, on the same fault line

- **Write side.** `AzureServiceBusMessagePublisher` assigned the `Uri` *object* for `dataschema`,
  asymmetric with `source` two lines above. The Service Bus SDK serialises a `Uri` application
  property via `Uri.AbsoluteUri` (confirmed in `AmqpMessageConverter::TryGetAmqpObjectFromNetObject`
  in the pinned 7.20.2), which throws `InvalidOperationException` for a relative URI — so publishing
  a relative dataschema failed inside the SDK at send time, before reaching a broker.
- **`dataschema` no longer fabricates a value.** `GetCloudEventsDataSchema` returned
  `http://goparamore.io` when the dataschema was absent, empty or unparseable. That is the documented
  default for `Source`, not for `DataSchema`; `MessageHeader.DataSchema` is nullable and every other
  backend yields `null`. ⚠️ **This is a deliberate behaviour change.** It also makes the
  previously-dead null-guard on the publisher live, so ASB stops re-publishing an invented dataschema
  on every requeue.
- **`RelationDatabaseOutbox`** read `DataSchema` with `UriKind.Absolute`, silently dropping a relative
  dataschema to `null` on read — inconsistent with the `Source` reader in the same file and with every
  other Outbox. It lost data rather than failing, so nothing caught it.
- **`GetSource`'s fallback** now uses `MessageHeader.DefaultSource` rather than a hard-coded literal.
  `JustSayingTransform` compares against that constant, an identity that held only because the
  literals happened to match. (Structural; no behaviour change.)

## Tests

Four regression tests, all broker-free, each red for a distinct reason before the fix:

| test | failed before with |
| --- | --- |
| `When_receiving_a_message_with_relative_cloud_events_uris_should_preserve_them` | `UriFormatException`, thrown at the **Act** line — reproduces the CI stack locally |
| `When_publishing_a_message_with_a_relative_dataschema_should_write_it_as_a_string` | `Expected: typeof(string)` / `Actual: typeof(System.Uri)` |
| `When_receiving_a_message_with_no_dataschema_should_leave_it_null` | `Expected: null` / `Actual: http://goparamore.io/` |
| `When_storing_a_message_with_a_relative_dataschema_should_read_it_back` | `Assert.Equal() Failure: Values differ` |

The write-side test asserts the observable wire contract rather than the SDK throw, because that throw
happens inside AMQP conversion at send time, which no test reaches without a broker.

## Verification

`Core.Tests` 978 passed · `Sqlite.Tests` 153 passed · `InMemory.Tests` 153 passed · the 4 new tests green.

The broker-backed ASB tests need credentials not available locally and fail in their constructors with
`ASB ConnectionString or Namespace not set` — pre-existing and unrelated.

⚠️ **Not yet exercised end-to-end against a real broker.** #4309 still blocks the generated ASB suite
upstream of receive: those tests send before any subscription exists, so almost none of them ever
receive a message. Only the two delayed-send tests got far enough to hit this defect at all. Once
#4309 is fixed, the generated ASB suite exercises this path properly.

The full diagnosis is recorded in `bugfixes/0019-asb-relative-cloudevents-source/bugfix.md`.


iancooper:
Thanks — this was a useful review. Addressed in 64eae7539. I verified the two findings that assert live
defects against the code rather than taking them on trust; both held.

## Points 1–3 (wanted before merge)

**1. Release note.** Added under `## Master` in `release_notes.md`, with an explicit upgrade instruction
("if you read `Header.DataSchema` off an ASB message without a null check, add one") rather than just a
description of the change.

**2. Outbox coverage — went with the template.** You were right that the hand-written test guarded
SQLite sync only while the reader serves five providers. Added a Liquid template pair under
`Templates/Outbox/{Sync,Async}/` and regenerated: **27 tests across all eight outbox suites**, including
the async path and **Spanner**, neither of which the hand-written test touched. The hand-written test is
deleted as redundant.

Verified non-vacuous rather than assumed — reverting just the reader change with
`git checkout HEAD~1 -- src/Paramore.Brighter/RelationDatabaseOutbox.cs` fails all four SQLite variants
(Text/Binary × Sync/Async), and they pass with it restored.

I did *not* take the cheaper option of flipping `_dataSchema` to relative in the builder. It would have
been free coverage, but it changes what every generated test exercises on all 17 suites, which is both a
default change and a much wider blast radius than this PR should carry.

**3. Relative-source publish coverage.** Added the round-trip test you suggested, pinning `source` and
`dataschema` across publish → receive in one test. It passed the moment I wrote it, which proves
nothing, so I reverted the gateway fix and re-ran: `UriFormatException`, as expected. Non-vacuous.

## Points 4–6

**4.** Hoisted to `private static readonly Uri s_defaultSourceUri`, matching `JustSayingTransform.cs:248`.

**5 — this one was a real bug, thank you.** Confirmed your reading of `BrokeredMessageWrapper.cs:20`: an
AMQP-null property arrives as a null *value*, and `property.ToString()` threw `NullReferenceException`. I
wrote the test first — it reproduces reliably. Now guarded with `property is not string dataSchema`, which
collapses the two `if` blocks exactly as you suggested; the two methods now read identically. Squarely on
the fault line this PR set out to close, so in scope.

**6.**
- Surviving literals → **#4313**. `Publication.cs:64` is the more significant of the two, since it seeds
  the default source for every publication that does not set one.
- `bugfixes/.current-bug` → **#4314**. Not theoretical: it blocked a branch switch for me during this
  work, and `git log` shows a prior `chore: allow housekeeping of the current bug` commit dealing with
  the same thing.
- Log level → lowered to `Debug`, with a comment saying why. Absence is the normal outcome now that it is
  no longer substituted.
- Wire-literal rationale → comment added to the three tests, so the next reader does not "tidy" it.

**One correction.** The `TryCreate` fallbacks are not quite unreachable: `RelativeOrAbsolute` also fails on
input over 65519 characters, which nothing guards. Reachable, just not worth a test. Left as-is, which is
what you wanted anyway.

## Verification

`Core.Tests` 978 passed · `Sqlite.Tests` 156 passed · generator audits 19 passed · `AzureServiceBus.Tests`
120 passed. All eight affected outbox test projects build.

The 10 ASB failures are pre-existing and environmental — each fails in ~1 ms in its constructor with
`ASB ConnectionString or Namespace not set`, because the broker-backed tests need credentials that are not
available locally. CI has them.

Fair point on your caveat, and worth restating: this still is not exercised end-to-end against a real
broker, because #4309 blocks the generated ASB suite upstream of receive.

Issue #4312, opened 2026-09-07
Azure Service Bus producers throw a 409 when several threads race to create the same topic


iancooper:
## Summary

Azure Service Bus producers create their topic/queue lazily on the send path, using an unsynchronised
check-then-act that does not tolerate losing the race. When several threads publish to the same
not-yet-created topic at once, they all observe "does not exist", they all call `CreateTopicAsync`, and
the losers get a 409 that is rethrown rather than absorbed:

```
Azure.Messaging.ServiceBus.ServiceBusException : SubCode=40900. Conflict. You're requesting an
operation that isn't allowed in the resource's current state.
```

(`SubCode=40901`, "Another conflicting operation is in progress", appears from the same race.)

## The offending code

`src/Paramore.Brighter.MessagingGateway.AzureServiceBus/AzureServiceBusTopicMessageProducer.cs:62-89`
— `EnsureChannelExistsAsync`:

```csharp
if (await _administrationClientWrapper.TopicExistsAsync(channelName))   // check
{
    TopicCreated = true;
    return;
}
...
await _administrationClientWrapper.CreateTopicAsync(channelName);       // act
TopicCreated = true;
```

There is no lock, and the `catch` rethrows every exception unconditionally. `TopicCreated` is set only
*after* a successful create, so it cannot serialise concurrent callers either.

`AzureServiceBusQueueMessageProducer.EnsureChannelExistsAsync` (`:62-89`) has the identical shape for
`QueueExistsAsync` / `CreateQueueAsync`.

This runs on the **send** path — `SendWithDelayAsync` → `GetSenderAsync` → `EnsureChannelExistsAsync`
(`AzureServiceBusMessageProducer.cs:248`) — so any concurrent first publish to a new topic can hit it.

## The asymmetry that shows the fix

**Both consumers already handle this correctly.** `AzureServiceBusTopicConsumer.cs:103` and
`AzureServiceBusQueueConsumer.cs:112` both do:

```csharp
catch (ServiceBusException ex)
{
    if (ex.Reason == ServiceBusFailureReason.MessagingEntityAlreadyExists)
    {
        Log.MessageEntityAlreadyExists(s_logger, Topic, _subscriptionName);
        _subscriptionCreated = true;      // losing the race is success
    }
    else { ... }
}
```

**Neither producer does** — `grep -rn 'MessagingEntityAlreadyExists' src/Paramore.Brighter.MessagingGateway.AzureServiceBus/`
[...]

PR #4317, opened 2026-09-07
PROBE (do not merge): confirm #4309 — provision the ASB subscription before the send

iancooper:
## ⚠️ EXPERIMENT — DO NOT MERGE

This is a probe, in the same spirit as #4308: it exists to **confirm or refute** the root-cause
hypothesis for #4309. Read the `azure-ci` result, then close it unmerged.

`azure-ci` only runs on `pull_request`, never on a plain branch push — which is why this has to be a PR
at all.

## The hypothesis under test

Full triage in `bugfixes/0018-asb-subscription-created-after-send/bugfix.md`.

`AzureServiceBusChannelFactory` provisions nothing on the broker — `CreateAsyncChannel` builds a
`ChannelAsync` around a consumer and returns. The ASB **subscription** is created lazily, inside
`AzureServiceBusTopicConsumer.EnsureChannelAsync`, which only runs from
`ReceiveAsync`/`AcknowledgeAsync`/`NackAsync`/`RejectAsync`/`PurgeAsync`. Every generated test sends
before it first receives, and an ASB topic with **no subscription attached silently discards** the
message. The first `ReceiveAsync` then creates the subscription — too late — and every read returns
`MT_NONE`.

Two independent controls point the same way:

1. The hand-written ASB tests pass, and they pass because they **pre-create the subscription**
   explicitly — `When_posting_a_message_via_the_producer.cs:65` — before building a channel and
   sending. Three for three across the hand-written tests that round-trip through the broker.
2. The **only** generated test that ever received a message is the delayed-send one, the only test
   whose send is deferred. Its 2 s before-delay `ReceiveAsync` creates the subscription at t≈0, and the
   message arrives at t=5 s to a subscription that by then exists.

## What this PR changes

- **`AzureServiceBusMessageGatewayProvider`** provisions the topic and subscription before handing out
  a channel, guarded on `OnMissingChannel.Create`. `AdministrationClientWrapper.CreateSubscriptionAsync`
  creates the topic first if missing, so one call covers both entities. This copies the hand-written
  tests' proven approach and mirrors how AWS and GCP ensure infrastructure before returning a channel.
- **The 11 ASB ledger cells** flipped to `Pass` and the suite regenerated — removing the `Deferred`
  skips from 22 files. Regeneration touched exactly those 22 files plus the ledger row, and the
  generator's own suite still passes 216/0.
- **Merges #4311** (`fix/4310-asb-relative-cloudevents-source`) so the probe measures *this* defect.
  Without it the 26 tests would unmask straight onto #4310's `UriFormatException` and tell us nothing.

## How to read the result

Baseline is run [`34127954869`](https://github.com/BrighterCommand/Brighter/actions/runs/34127954869):
`Total 161 / Passed 131 / Failed 30`, split 26 `MT_NONE` + 2 `UriFormatException` + 2 `SubCode=40900`.

| outcome | reading |
| --- | --- |
| the 26 go **green** | **#4309 confirmed.** Move the ASB cells for real, and settle the design question below. |
| the 26 stay `MT_NONE` | **#4309 refuted.** The lazy-subscription theory is wrong — re-triage. |
| the 26 become `UriFormatException` | the #4311 merge did not take; fix the branch before concluding anything. |

⚠️ **The 2 `When_multiple_threads_try_to_post_a_message` failures are expected to survive.** That is
**#4312**, a concurrent topic-creation race in the producer, and it is not evidence either way here.

## The design question this probe does NOT settle

Whether lazy subscription creation is itself a **gateway** defect, rather than something the test
harness should paper over. A caller that builds a producer and a consumer and then sends currently
loses the first message silently, with no error anywhere — and `AzureServiceBusChannelFactory` receives
the subscription's `MakeChannels` setting and ignores it, while AWS and GCP both honour it eagerly.
That is a maintainer call, recorded in #4309, and it is deliberately out of scope for a probe.


iancooper:
## ✅ Result: hypothesis CONFIRMED

`azure-ci` job [`101853410391`](https://github.com/BrighterCommand/Brighter/actions/runs/34157403107/job/101853410391),
against baseline run [`34127954869`](https://github.com/BrighterCommand/Brighter/actions/runs/34127954869):

| | baseline | this probe | |
| --- | --- | --- | --- |
| `MT_NONE` wave (#4309) | **26** | **2** | ✅ 24 of 26 fixed |
| `UriFormatException` (#4310) | 2 | **0** | ✅ fixed |
| `SubCode=40900` (#4312) | 2 | **2** | expected to survive |
| **distinct failing tests** | **30** | **6** | |

**#4309 is real, and provisioning the subscription before the send is the right fix.** 24 of the 26
`MT_NONE` failures went green.

**#4311 is now proven against a live broker.** The delayed-send pair no longer throws
`UriFormatException` — it receives the message and fails on a *value* comparison instead, which is
strictly further through the code. Until now that fix had only local unit coverage.

**#4312 survived exactly as predicted**, both `MultipleThreads` variants. Not evidence against anything.

## Two new findings, neither previously visible

Both were masked by #4309. Neither is a reason to doubt the result above.

**(a) `When_rejecting_message_with_unacceptable_reason_should_send_to_invalid_channel` (both pumps)** —
still `MT_NONE`. The only canonical behaviour that did not go green, and it looks like it *should not*:
`AzureServiceBusMessageGatewayProvider`'s own doc comment says ASB "dead-letters natively: the DLQ is
always the built-in sub-queue... There is no separate invalid-message channel", and `CreateSubscription`
takes `invalidMessageRoutingKey` but records that it has "no structural effect on the ASB subscription
itself". `GetMessageFromInvalidChannelAsync` reads a `{topic}.Invalid` topic that nothing publishes to.

So this cell is probably a genuine **`Deferred`** — ASB has no invalid-message-channel concept — rather
than a failing `Pass`. That needs a maintainer ruling, not a code change.

**(b) `When_sending_a_delayed_message_should_deliver_after_delay` (both pumps)** —
`Assert.Equal() Failure: Values differ`. The message now arrives, but something in the round-trip does
not match what was sent. **New, undiagnosed.** One candidate is the escaping round-trip loss noted in
`bugfixes/0019-…/bugfix.md` (the SDK writes a `Uri` property as `AbsoluteUri` but reads it back via
`ToString()`), but that is a guess, not a diagnosis. Not yet triaged, no issue raised.

## Recommended next steps

1. **Settle the design question in #4309** — is lazy subscription creation a *gateway* defect, or is
   eager provisioning the harness's job? This probe fixed it harness-side; that choice is not ratified.
2. **Land the real fix on #4297**, since `AzureServiceBusMessageGatewayProvider` only exists there.
3. **Move 9 of the 11 ASB ledger cells to `Pass`**; leave the invalid-channel cell `Deferred` per (a),
   and decide the delayed-send cell once (b) is triaged.

Closing this unmerged now, as #4308 was — it has served its purpose.

iancooper:
Probe complete — hypothesis confirmed, result recorded above. Closing unmerged by design, as #4308 was. The branch `probe/4309-asb-eager-subscription` is left in place so the CI run and the diff stay inspectable.

PR #4323, opened 2026-09-08
Test Generator: audit the generated tree against what the configuration would produce

### LINKED ISSUE #4305: Test Generator: no check detects a generated file the current configuration would not produce
## Summary

`tools/Paramore.Brighter.Test.Generator` writes files. It never removes one it previously wrote. So the moment a capability flag flips from `true` to `false`, a template is renamed or deleted, or a generation code path is dropped, whatever was previously emitted stays on disk — it keeps compiling, and it keeps running. The tree drifts from what the configuration says it should contain, and **no test, audit or build step in this repo can detect it.**

`generate-test.sh` is idempotent with respect to *writes* and silent with respect to *deletes*.

## Evidence: this has happened three times

1. **#4238 / #4300.** `f16a0c896` dropped the `Sync` call from the singular `Outbox` branch of `OutboxGenerator`. **37 files** across `DynamoDB.Tests`, `DynamoDB.V4.Tests` and `MongoDb.Tests` were orphaned — still executing, no longer refreshed — and nothing reported it for the entire interval since. It stayed benign only by luck: `git log f16a0c896..HEAD -- tools/Paramore.Brighter.Test.Generator/Templates/Outbox/Sync` is empty, so the frozen copies still match. Any edit to those templates would have silently divorced the three projects from them.
2. **`f7fa9de3c` (during #4240).** Narrowing the Kafka/Consumer infrastructure gate meant two `assume_channel` files should no longer exist. The generator left them; they had to be `rm`'d **by hand**, and nothing would have caught it had we not been watching.
3. **The RocketMQ partition-key pair.** `tests/Paramore.Brighter.RocketMQ.Tests/MessagingGateway/Generated/{Reactor,Proactor}/When_posting_a_message_with_partition_key_via_the_messaging_gateway_should_be_received.cs` were hand-written into a `Generated/` directory under an `<auto-generated>` header at #3996. No `.liquid` of that name has ever existed — checked across all of history, not just deletions. They assert a partition-key round trip no current template covers, and they have survived unnoticed ever since.

## Measurement

Modelling the expected output set — mirroring `BaseGenerator`'s template enumeration, both `SkipTest` predicates, and the prefix logic — and diffing it against the tree:

| tree | expected | on disk | orphans (on disk, not expected) | missing (expected, not on disk) |
|---|---|---|---|---|
| `master` (`cd50a43a9`) | 741 | 780 | **39** | 0 |
| PR #4297 branch (`f80127f04`) | 1259 | 1298 | **39** | 0 |
| PR #4297 branch, with #4300 applied | 1296 | 1298 | **2** | 0 |

The totals differ between the first two rows because #4297 adds templates and wired configurations. **The orphan set is the same 39 paths in both trees** — this is a `master` problem, not something #4297 introduces.

**#4300 is expected to clean up 37 of the 39.** It fixes the cause of the largest group at source, so this issue should be scoped to what remains: the two RocketMQ files, and the absence of any guard.

Zero `missing` in every row means the model is exact — it reproduces every file the generator would write — so the counts above are trustworthy rather than approximate.

## Why nothing currently catches it

- **ADR 0066 already names the class of problem** — *"deleting a `.liquid` template does not delete what it previously produced"* — but only handled the instances it created. Phase 5 hand-swept 80 legacy orphans and 40 stale provider-interface copies. The general case was never addressed.
- **`Cleanup/When_legacy_templates_deleted_…`** guards a **closed list of four** legacy names. It is a restore-guard for those four, not an orphan detector.
- **`LedgerSkipCrossCheckAudit` and `GatewaySkipConventionAudit`** walk files that **exist** and check each against its ledger cell. They are structurally blind to a file that should not exist — an orphan has no cell to disagree with.
- **`When_generating_everywhere_should_emit_skipped_canonical_suite_in_all_wired_projects`** asserts presence. It never asserts the absence of extras.
- **A regenerate-and-fail-on-a-dirty-tree CI job** — suggested as follow-up 3 on #4300, and worth having on its own merits — would **not** have caught any of the three. It guards the *content* of files the generator writes. Regeneration leaves an orphan untouched and the tree stays clean.

## Proposal: an expected-set audit, asserting both directions

Compute the set of files the current configuration would produce, and compare it with what is on disk:

- **orphans** — on disk under a `Generated/` directory, not in the expected set
- **missing** — in the expected set, not on disk

Both must be zero. The two are different assertions and neither subsumes the other: `orphans` catches a flipped gate or a deleted template, `missing` catches a dropped code path. **The `missing` direction is what would have failed CI the day #3996 landed** — precisely what #4238 needed and did not have.

Implementation notes:

- The expected-set computation should be extracted from `BaseGenerator` and the two generators so a test can call it. `CanonicalBehaviours` (`08d6b4f3d`) is the precedent for lifting exactly this kind of shared definition out of generator-private state so an audit and the generator read one copy.
- It must be **proven non-vacuous** — a scan passes trivially on zero files. Plant a canary, as the Phase 6 audits do.
- The whole `Generated/` tree is generator-owned. The `IAmAMessageGateway*Provider.cs` and `IAmAnOutboxProviderSync.cs` files that live there **do** have templates, which is why they do not show as orphans; nothing hand-written legitimately lives under `Generated/` once the RocketMQ pair is moved out.

## Why an audit rather than teaching the generator to prune

Pruning — having the generator delete anything in a `Generated/` directory that it did not just write — is the stronger guarantee and makes regeneration truly idempotent. It is also the wrong first move here:

**At `f16a0c896`, a pruning generator would have silently deleted 37 real tests.**

An orphan means the configuration and the tree disagree. It does not say which side is wrong — and in the measurement above the *configuration* was the wrong side 37 times out of 39. An audit reports and stops; a human decides. Pruning turns a generator bug into data loss.

Once the audit is green in CI, a `--prune` convenience flag becomes a defensible follow-up. It should not lead.

## Preconditions

The audit should be introduced **green**, as a passing guard rather than a red one. Two things have to land first:

- [ ] **#4300** — reclaims the 37 `Outbox/Generated/Sync/` files (39 → 2)
- [ ] **#4306** — moves the RocketMQ partition-key pair out of `Generated/`, into `MessagingGateway/{Reactor,Proactor}/`, stripping the `<auto-generated>` header and leaving the assertions untouched (2 → 0). It is a hand-written test and always was; the move stops it claiming otherwise. The namespace is already `Paramore.Brighter.RocketMQ.Tests.MessagingGateway.{Reactor,Proactor}` in both locations, so nothing else changes.

Making it genuinely generated was considered and rejected as disproportionate: the ledger has no FR column for partition-key behaviour, and this repo expresses ordering coverage as a *dedicated wired configuration* (`Kafka / PartitionKey`, `GCP / StreamOrdering`, `AWS / SqsFifo`), not as an FR. A `RocketMQ / PartitionKey` configuration — new provider, new ledger row, `EXPECTED_WIRED_CONFIGURATION_COUNT` 24 → 25 — is a reasonable separate proposal, but it is not a precondition for this.

## Notes

Found while auditing PR #4297 (#4240). The two-orphan symptom was recorded during that work; the full 39 surfaced only when the check was done at file level rather than by basename — a basename comparison is blind to a file whose name also exists legitimately elsewhere, which is exactly the Kafka `assume_channel` case.

iancooper:
Closes #4305.

## The problem

`tools/Paramore.Brighter.Test.Generator` writes files and never removes one it previously wrote. The moment a capability flag flips from `true` to `false`, a template is renamed or deleted, or a generation code path is dropped, whatever was previously emitted stays on disk — it keeps compiling, and it keeps running. Nothing in the repository could detect that. #4305 records three occasions it has already happened.

Two directions have to be asserted, and neither subsumes the other:

- **orphans** — on disk under a `Generated/` directory, not in the expected set. A flipped gate or a deleted template.
- **missing** — in the expected set, not on disk. A dropped code path, or a project nobody regenerated.

## What this does

**1. `refactor: describe generator output as a plan the generator renders`**

The expected output set existed only as a side effect of rendering: each generator interleaved *which templates, into which folder, minus which skips* with the rendering itself, so nothing could ask what the generator owns without running it.

- `PlannedFile` — one file the generator will write.
- `GenerationSuite` — one template-folder-to-destination rendering.
- `BaseGenerator.Plan` computes a suite's planned files; `GenerateAsync` now renders exactly that list.
- `OutboxGenerator` and `MessagingGatewayGenerator` describe their work once, in `Suites`, which both `GenerateAsync` and a new public `Plan` walk. The files written and the files claimed are one description, not two that drift.

Structural only, and verified as such: running `./generate-test.sh` over the whole repository afterwards changed **no** generated file.

**2. `fix: generate the Sync DataSchema test that #4300 unblocked`**

The audit's first catch, found while building it. #4300 restored the `Sync` suite to the singular `Outbox` branch of the generator, but the three projects that configure an outbox that way — DynamoDB, DynamoDB.V4 and MongoDb — were never regenerated, so the Sync arm of the relative-DataSchema test has been absent from them since. `master` was **3 missing**, not 0.

**3. `test: audit the generated tree against what the configuration would produce`**

`GeneratedTreeAudit` asks the generators themselves, through the `Plan` methods, what they would write for each checked-in configuration, and diffs that against the `Generated/` tree in both directions. The expected set is not a description of the generator kept alongside it — a second description would be a second thing to keep in step, and would agree with the generator exactly until the day it mattered.

Proven non-vacuous twice over:

- **Canaries** — two tests generate a small tree with the real generators, confirm the audit agrees with it, then break it each way and assert that *exactly* the broken file is reported.
- **A real defect** — with the three files above removed, the audit fails and names all three.

Scope is the `Generated/` tree, which is generator-owned in full. Shared files a generator writes to a project root sit alongside hand-written code and are not audited.

Per #4305, this is an audit and not generator-side pruning: at `f16a0c896` a pruning generator would have silently deleted 37 real tests. An orphan says configuration and tree disagree; it does not say which is wrong, and the configuration was the wrong side 37 times out of 39.

**4. `ci: run the test generator's own tests`**

`Paramore.Brighter.Test.Generator.Tests` is in `Brighter.slnx`, so the `build` job compiles it — but nothing has ever run it. Every audit written against the generator, this one included, could not fail CI. It now runs alongside the other tests the `build` job owns.

## Notes for #4240 / #4297

That branch adds a **third** suite to `MessagingGatewayGenerator` (`MessagingGateway/Shared`) and a `prepareModel` argument for the ledger-driven canonical Skip, both in the method this PR restructures. The merge must fold both into `SuitesFor(...)` — the Shared suite takes no ignore and no `prepareModel`, the Reactor/Proactor suites take both — or the audit's expected set silently loses the Shared files.


iancooper:
Thank you — this was a good catch on item 1, and the rest was worth doing too. All of it is actioned, pushed as `1ba62334f` → `17a9ad98d`.

### 1. Order-dependent assertion — confirmed, and it really is a coin flip

Verified the mechanism exactly as described: of the 14 `Templates/Outbox/Sync` templates, only `IAmAnOutboxProviderSync.cs.liquid` never mentions `MessageBuilder`, and `SupportsTransactions` defaults to `true` so it is always written.

Rather than take it on reading, I demonstrated it as a genuine RED by sorting the enumeration:

```
Assert.Contains() Failure: Sub-string not found
String:    "#region Licence\n\n/* The MIT License (MIT)"···
Not found: "TestMessageBuilder"
```

`44b805066` names `When_Adding_A_Message_It_Should_Be_Stored_With_All_Properties.cs` in all three tests. You were right that the other two are safe by luck rather than construction, so they are pinned the same way.

### 2. Configuration loading — `e7c32e32b`

Confirmed the divergence is already there: `Program` honoured an explicit `DestinationFolder` and defaulted only an empty one, the audit overwrote it unconditionally. Latent, since no checked-in configuration names one — but exactly the drift the PR is about, in the half it had not covered.

`TestConfigurationLoader.Load(path, defaultDestinationFolder)` is now the single statement. `Program`, the audit **and the orphan canary** all read through it — the canary was a third copy, deserializing the JSON string separately from the file it had just planted. It also pins the `JsonSerializerOptions` and owns the file name all three had spelled out.

### 3. CI placement — `fb7863f35`, plus the csproj half in `9b358ce55`

Moved above `Core Tests`. Already visible on this push: the audit reported at 21:30:10, `Core Tests` finished at 21:32:14.

On the two things you asked to eyeball rather than assume:

- **Templates do flow transitively.** Verified on a clean rebuild — all 59 reach the test assembly's output directory through the `ProjectReference`.
- **`build` is green on the new head**, with `Test Generator Tests` reporting **28/28 on net9.0 and net10.0**.

Your csproj point is real and not hypothetical: `master`'s csproj has no `Templates\MessagingGateway\Shared\*`, and #4297 adds it by hand. `Templates\**\*.liquid` now covers everything — all 59 templates are `.liquid`, so nothing else was being copied by the old list.

### Smaller points — all actioned

`71e661f36` reworded the `Suites` logs to "Planning". `17a9ad98d` covers the rest, and two of them are worth reporting back because the measurement changed the picture:

- **`IsBuildOutput` never matched anything.** `find tests -type d -name Generated` returns 35 both with and without `bin`/`obj` pruned — generated `.cs` files are compiled, never copied to output. So removing it is behaviour-preserving; it only ever paid for the walk. The waste you identified was real, though: after a full `dotnet build -c Release`, `tests/` holds **4,317** directories and the pruned walk visits **456** — and it happened four times before. Descending past `bin`/`obj` rather than into them removed both the cost and the need for the length-slicing, so `Path.GetRelativePath` was not needed either.
- **`LocateTestsRoot` markers on `Brighter.slnx`, deliberately not `.git`.** In a git worktree `.git` is a *file*, not a directory, so a `Directory.Exists` probe would have failed in every worktree — which is where this work actually happens.

Also done: an `ICollectionFixture` shared by the two facts, and `GeneratedTreeAudit.Of` in place of the constructor that did the I/O.

### Not actioned

Two items from your first review, neither re-raised here: content checking as well as existence, and `Ordinal` comparison against a case-insensitive enumeration. Both are recorded as follow-ups. The pruning walk does now compare the `Generated` folder name `Ordinal`, so at least the two sides of the comparison agree with each other.

**28/28 on net9.0 and net10.0, and `./generate-test.sh` over the whole repo changes no generated file.**

iancooper:
Thanks — all of it read, and the last round in particular caught things worth catching.

**Everything marked "would fix before merge" across the review rounds is now actioned on this
branch.** What was left is deferred to **#4333**, which collects it in one place rather than
leaving it in a thread. Nothing deferred is known to be live: every item is latent, cosmetic, or a
hardening step.

### Actioned since the first review

| commit | what |
|---|---|
| `a685f215e` | `WithDefaultsFrom(TestConfiguration)` beside `WithPrefix`, applied in `SuitesFor`; **both `GenerateAsync` overrides deleted**, so `Plan` and `GenerateAsync` see one model. |
| `db28fbfb2` | `TestConfigurationLoader.Load` resolves a relative `DestinationFolder` instead of leaving each caller to resolve it. |
| `dca86e28d` | `Assert.NotEmpty` on `Expected`/`OnDisk`; the content-drift caveat written into `GeneratedTreeAudit`'s remarks; `-f net10.0`; `s_buildOutputFolderNames`; `{FileCount}` → `{DestinationFolder}`; `GC.SuppressFinalize` dropped; `MemberwiseClone` caveat recorded. |
| `ccd4c5863` | `ignore` matched on `file.Name`, not the absolute path; the walk collects **every** extension, not just `*.cs`. |
| `ceee163c4` | Resolution base is the **configuration file's own folder**, so the two readers agree by construction rather than by convention; the defaulted `DestinationFolder` made absolute; both parent-inheritance assertions tightened; `.agent_instructions/generated_tests.md` documents the audit. |
| `198fec438` | Shared defaults moved out of `SharedGenerator` into the loader, so **no generator mutates the root configuration** and none has to run before another; three canaries — flag off → orphans, flag on → missing, deleted configuration → whole tree orphaned. |

### Three things worth reporting back, because they change what the evidence is worth

- **The `Plan`/`GenerateAsync` defaulting refactor is structural, and that is checkable rather than
  asserted**: `./generate-test.sh` over the whole repository afterwards changes **zero** generated
  files, on every one of the commits above.
- **The first `DestinationFolder` test was vacuous and passed against either implementation.** It
  put the configuration and the default in two *sibling* temp folders, and `../Sibling.Tests`
  resolves identically from both. Replaced with one that nests them, and verified by reverting the
  base and watching only that test go red. Same trap the non-vacuity guards exist for — worth
  recording that it caught us too.
- **The first attempt at the flag-on canary used `HasSupportToDeadLetterQueue` and failed.** That
  template's name also carries `requeuing`, so `HasSupportToRequeue` kept it skipped. It uses
  `HasSupportToPublishConfirmation`, the only rule matching its substring. Had that template been
  gated by one rule rather than two, the test would have passed for the wrong reason.

**31 → 39 tests, green on net9.0 and net10.0.**

### Deferred to #4333

Grouped there by when it matters rather than by severity. The three that most deserve a look:

- **`IsUnderAGeneratedFolder` splits the absolute path while `Collect` starts below `testsRoot`.**
  Correctly identified as latent-until-a-third-suite — which is exactly what the #4240/#4297 merge
  adds, so it is filed under "do this as part of that merge" along with a guard against the fold
  silently *shrinking* the expected set.
- **`BaseGenerator.GenerateAsync` is still `protected virtual` with no override left.** Agreed that
  non-virtual turns "do not reintroduce this" into a compile error; held back only because it is
  the kind of change that reads better separated from the one that removed the overrides.
- **Config discovery one level deep vs a whole-tree walk.** The point that the failure mode here is
  *destructive advice* — a nested project reported wholesale as orphans, under a message that ends
  in "delete by hand" — is the sharpest thing raised in these rounds, and #4333 records it that way.

Also carried over: content-comparison auditing, the `Missing` direction for the shared root files,
`Ordinal` vs case-insensitive comparison, the standalone `generator-audit` job so an audit failure
stops suppressing `Core Tests` and package publication, and the naming and documentation nits —
**including two this branch introduced and one it made redundant**: a second sibling `<remarks>` on
`TestConfigurationLoader` (the same defect fixed on `WithPrefix` earlier in this PR, then
reintroduced), a test whose name now states the one thing its behaviour is not, and a `using System;`
left unused by the deleted override.

The Tidy First point about the `ignore` file-name change landing in a `refactor:` commit is taken
and recorded in #4333 — it is behavioural, and a zero-diff regeneration does not establish
otherwise, since the two behaviours diverge only for checkout paths nobody has.

PR #4331, opened 2026-09-10
samples: fix a fatal subscription type in the MsSql task queue, and show the Outbox and Inbox

iancooper:
## What this fixes

**Two receivers in `samples/TaskQueue/MsSqlMessagingGateway/` could never have started.** Both built their subscriptions as `Subscription<T>`, and `MessagingGateway.MsSql`'s `ChannelFactory` downcasts to `MsSqlSubscription` and throws in all three of its channel-creation methods (`ChannelFactory.cs:46`, `:66`, `:88`). Measured, with a control:

```text
Subscription<T>      : ConfigurationException: MS SQL ChannelFactory We expect an
                       MsSqlSubscription or MsSqlSubscription<T> as a parameter
MsSqlSubscription<T> : accepted
```

It compiles, which is why nothing caught it. Two further applications in the same folder were also unstartable: `CompetingSender` passed a bare `new Publication()` and threw `ConfigurationException: Topic is missing from the publication`, and `CompetingReceiverConsole` defaulted to a Proactor pump against a sync handler.

**`GreetingsSender` had no Outbox and no host.** At the base commit it built a bare `ServiceCollection`, took `.Create()` on a producer registry and called `Post` — no Outbox, no Box Provisioning, no `Host`. This PR adds all three, and the ordering matters: `UseBoxProvisioning` registers an `IHostedService`, and `BoxProvisioningHostedService` is the only caller of `ProvisionAsync` in `src/`, so the sender uses `StartAsync` to let provisioning complete before the send. (An earlier revision of this description said the call existed and nothing ran it; that described a snapshot from partway through the work, not the base.)

**The global `InboxConfiguration` was silently doing nothing.** It reaches the pipeline only through `AddEventBus`'s `ExternalBus` arms; with no producers the `NoExternalBus` branch is taken, and that overload accepts no inbox. Measured: **0 Inbox rows with no producer registered, 1 with one.** De-duplication now comes from `[UseInbox]` on the handler, which does not depend on a bus; `InboxConfiguration` remains as what supplies the store.

## What this adds

One database carrying the queue, the Outbox and the Inbox, and **the sample now runs from a clean machine**.

- **The queue table is provisioned by the sample**, in `SampleInfrastructure/QueueTableProvisioner.cs`, guarded and idempotent. Its catalog lookups take the names as parameters — the values are constants, so nothing is injectable, but a sample is a copy-paste source. The MSSQL gateway has no provisioning path — `OnMissingChannel.Create` is inert there — so `MsSqlQueueBuilder.GetDDL`/`GetIndexDDL` are public precisely for callers to run. This is now their only production call site in the repository.
- **`BrighterSqlQueue.sql` is 13 lines of portable UTF-8** creating only the database. It was a 2018 SQLEXPRESS dump: UTF-16, with data files pinned to `C:\Program Files\...\MSSQL12.SQLEXPRESS\`, so it could not run off Windows.
- **Each process provisions the box it owns**, so either can be started first — including concurrently, which the provisioner handles.
- **A `README.md`**, documenting the changed setup contract: script → database, applications → tables.
- **All four applications read `ConnectionStrings:Brighter`**, defaulting to local SQL Express with `TrustServerCertificate=True` — required, since `Microsoft.Data.SqlClient` (pinned 7.0.2) defaults `Encrypt` to `true`.
- **Both receivers call `.ValidatePipelines()`**, so a pump/handler mismatch is a named startup error rather than a per-message failure.
- **A `SampleInfrastructure` project** holding the provisioner and `SampleDatabase`, so `Events` carries contracts and no database adapter — it is the assembly every application hands to `AutoFromAssemblies`.
- **A test pinning the downcast**, in `tests/Paramore.Brighter.MSSQL.Tests/MessagingGateway/`, needing no database: three channel-creation methods throw `ConfigurationException` for a plain `Subscription<T>`, and a control gets a channel for `MsSqlSubscription<T>`. Red-proofed by neutering the three guards — 3 failed, the control passed.

## Testing

Proven by **running**, on a fresh SQL Server 2022 container with the `.sql` run once, no manual table creation, committed source and one environment variable:

```text
two receivers started CONCURRENTLY on a database with zero tables:
  instance A: 0 race errors, instance B: 0 race errors
  tables: __BrighterMigrationHistory, InboxMessages, QueueData
  IX_QueueData_Topic present exactly once

convergence: index dropped by hand, recreated on next start, 0 errors

round trip:
  sender: Provisioned Outbox 'Outbox' / Decoupled invocation
  delivery 1, message id X : handler ran 1 time, 0 validation errors
  delivery 2, message id X : handler ran 0 times
    warn: UseInboxHandler  Command 01a08bdb-... has already been seen
  InboxMessages rows: 1
```

## The Competing pair now round-trips, and #4338 was this sample's own defect

It sent inside a `TransactionScope` it deliberately never completed, on the premise that a message is queued whether the transaction commits or aborts. `Post` opens a `SqlConnection`, `Enlist` defaults to `true`, so the insert joined that transaction and rolled back with it. True of a broker outside your database; false when the broker **is** your database.

```text
default connection string : 3 Posts, no errors, QueueData 0 rows
...;Enlist=False          : 3 Posts, no errors, QueueData 3 rows
with scope.Complete()     : 6 sent, handled 2 and 4 by two receivers, queue drained
```

**#4338 closes with this PR.**

## Filed from this PR

**#4334** validation spec for subscription vs `ChannelFactoryType` · **#4335** `NoExternalBus` discarding `InboxConfiguration` · **#4336** `InboxScope` inert · **#4337** `CommandProcessor.Call`'s `"fpr"` typo naming the wrong type · **#4338** the Competing pair, **closed by this PR** · **#4339** a test pinning the `ChannelFactory` downcast, **now closed by the test in this PR** · **#4340** superseded by the README here.

## Why

This accompanies a Brighter documentation guide on composing MSSQL as transport, Outbox and Inbox, and the sample is what the guide points readers at. Raised separately because the subscription fix stands on its own.


iancooper:
Both points were right, and the blocking one was right for the reason you gave — I verified it before changing anything: `UseBoxProvisioning` registers an `IHostedService` (`BrighterBuilderBoxProvisioningExtensions.cs:92`), and `BoxProvisioningHostedService.cs:89` is the **only** caller of `ProvisionAsync` anywhere in `src/`. A bare `BuildServiceProvider()` never starts hosted services, so my comment was simply false.

Fixed in `807b7e7e4`.

### 1. `GreetingsSender` is a host now

Converted to `Host.CreateApplicationBuilder` + `await host.StartAsync()`, copying the PostgreSQL sibling you pointed at — including its reason for `StartAsync` over `RunAsync`.

### 2. Each process provisions the box it owns

Sender provisions the Outbox, receiver provisions the Inbox. The receiver can now be started on its own, which was your second point.

### 3. A third one, which only running it surfaced

**The global `InboxConfiguration` was doing nothing at all.** It reaches the pipeline only through the `ExternalBus` overloads (`ServiceCollectionExtensions.cs:657-676`); with no producers registered the `NoExternalBus` branch is taken, and **that overload accepts no inbox**. So a consumer-only process — the natural shape for a receiver — gets no Inbox and no warning.

Measured, with a control:

```text
InboxMessages rows, receiver with NO producer registered : 0   (table existed, was provisioned)
InboxMessages rows, receiver WITH a producer registered  : 1
```

`GreetingEventHandler` now carries `[UseInbox]`, which does not depend on there being a bus. The `InboxConfiguration` registration stays, because it is what supplies the store.

Related, and worth knowing: **`InboxScope` is referenced nowhere in `src/` outside its own declaration** — 0 hits, against 43 for `OnceOnlyAction` as a control. The `scope:` argument is inert. I have not used it here.

### Proven by running, not building

You were right that the description's "the rest is a build" caveat left this open, so I stopped leaving it open. SQL Server 2022 in Docker, the queue table created from `MsSqlQueueBuilder`'s own DDL:

```text
Provisioning Outbox 'Outbox'...          Provisioned Outbox 'Outbox' successfully
Provisioning Inbox 'InboxMessages'...    Provisioned Inbox 'InboxMessages' successfully

delivery 1, message id X : handler ran 1 time
delivery 2, message id X : handler ran 0 times
  warn: Paramore.Brighter.Inbox.Handlers.UseInboxHandler
        Command 01a08aef-c8a4-738e-ad13-317ce5686212 has already been seen
InboxMessages rows after both : 1
```

Connection strings in the diff are unchanged from the original SQLEXPRESS ones; the Docker run used a temporary edit that is not committed.

### On your two remaining notes

- **Duplicated configuration** — both files now carry a comment pointing at the other and saying the values must agree. The receiver no longer sets `outBoxTableName`, since it has no Outbox.
- **`InboxMessages` vs the `Inbox` default** — kept deliberately, because the Brighter documentation's MSSQL Inbox page uses that name and this sample is what the new guide points readers at. Happy to take the default instead if you would rather the sample be minimal.

### On the two things you suggested as separate work

Both sound right and neither is in this PR: the `ChannelFactory`-rejects-a-non-`MsSqlSubscription` test, and having the sender create the queue table via `MsSqlQueueBuilder`. **The `InboxConfiguration`-silently-ignored behaviour in (3) also looks like a product defect rather than a documentation one** — a `ConfigurationException`, or honouring the inbox on the `NoExternalBus` path, would both beat silence. Say the word and I will raise any of the three; I would rather not open issues on your repo uninvited.

iancooper:
No blocker this round, but two of these were defects I introduced and both are fixed in `411b97e11`.

### 4 — `[UseInbox]` in the shared `Events` project. You are right, and it was mine.

All four apps reference `Events` and all four call `AutoFromAssemblies()`; only the receiver registers an inbox. Three processes were registering a handler whose pipeline cannot be built. Latent for the reason you give, and real. `GreetingEventHandler` now lives in `GreetingsReceiverConsole`, which owns the policy — and the comment says why, so the next person does not move it back.

### 5 — the line-number citation had already rotted, in a comment about rotting citations

`ServiceCollectionExtensions.cs:657-666` against an actual `:655` / `:665` / `:676`, and ambiguous between two files of that name. It now names **`AddEventBus`**. Taking the general point: citing a method is stable, citing a line is a hostage.

### 2 — the committed form was not the form that ran

Fair, and the sharpest note here. Both apps now read `ConnectionStrings:Brighter` with the SQLEXPRESS string as the fallback, so Windows readers are unaffected and everyone else can point it at a container. The packages were already referenced and unused, as you said.

**Re-proven end to end on the committed source, with only an environment variable set:**

```text
Provisioned Outbox 'Outbox' successfully
Decoupled invocation of message: Topic:greeting.event Id:01a08b63-...
receiver: provisioned Inbox 'InboxMessages'
delivery 1, message id X : handler ran 1 time
delivery 2, message id X : handler ran 0 times
  warn: UseInboxHandler  Command 01a08b63-5db8-711a-905d-9582f4557f77 has already been seen
InboxMessages rows after both : 1
```

Also done: `GreetingsSender` now references `Microsoft.Extensions.Hosting` explicitly rather than leaning on `BoxProvisioning`'s transitive one, and the stale `using System.Threading.Tasks;` is gone. **The PR body is rewritten** — you were right that it still described the pre-`Host` shape.

### What I have not done, and why

- **The validation spec** comparing `(subscription.ChannelFactory ?? options.DefaultChannelFactory).GetType()` against `Subscription.ChannelFactoryType`, and **the test pinning the downcast.** Both are clearly the right fix — they would make the explanatory comments in these files unnecessary — and both are `src`/`tests` changes that belong on their own branch rather than bolted onto a samples PR.
- **Shared `Db` constants.** I take the point that the sample teaches a coupling and then models the fragile way of managing it. It is a bigger edit than the rest of this PR and touches `CompetingReceiverConsole`/`CompetingSender` too, so I would rather do it deliberately than late.
- **`BrighterSqlQueue.sql`** (UTF-16, hardcoded `FILENAME`) and **a `README.md`.** Both worth doing, and both now more visible because this PR leans on that script.
- **`InboxScope` being dead.** Agreed it deserves an issue rather than a commit-message footnote.

That is five follow-ups. **I would rather not open issues on your repo uninvited** — say which of them you want raised and I will, or tell me to raise the lot.

iancooper:
Point 1 was right and it was mine — and running it turned up a third one. Fixed in `1b32c94ee`.

### 1. `CompetingReceiverConsole` — Proactor pump, sync handler

Verified before touching it: `CompetingConsumerCommandHandler` is a sync `RequestHandler<T>`, the subscription omits `messagePumpType` so `MsSqlSubscription<T>` defaults to Proactor, and `ConsumerValidationRules.PumpHandlerMatch` classes that as an **Error** in Brighter's own words. Your framing is the right one: fixing the subscription type moved the failure one layer down. Now `Reactor`, with the reason in a comment.

### 2. …and `CompetingSender` has never started either

You said the Testing section only exercised the Greetings pair. It did, so I ran the other one:

```text
Unhandled exception. Paramore.Brighter.ConfigurationException:
  MS SQL Message Producer Factory: Topic is missing from the publication
   at MsSqlMessageProducerFactory.Create() … MsSqlMessageProducerFactory.cs:line 62
   at Program.<Main>$ … CompetingSender/Program.cs:line 33
```

It passes `[new Publication()]` — no `Topic`. It now publishes `Publication<CompetingConsumerCommand>` on the routing key the receiver subscribes to.

### 3, 4, 5, 6 — all taken

- `IsNullOrWhiteSpace` rather than `??`, for exactly the reason you give.
- The comment said "appsettings or the environment"; there is no `appsettings.json` and a console app would not copy one, so it now says `ConnectionStrings__Brighter`. I have **not** claimed the Competing pair is portable — it still hardcodes SQLEXPRESS, and I would rather leave that honest than half-fix it.
- `contextKey: typeof(GreetingEventHandler)`, matching the WebAPI samples.
- `#region Licence` restored on `GreetingsSender/Program.cs` — dropped when I rewrote the file, and `.agent_instructions/documentation.md` asks for it.
- The provisioning ordering is now stated as load-bearing, not merely explained.
- Your last point is fair: *"pipeline cannot be built"* was stronger than the mechanism warrants. The comment now says latent rather than certain. I have also folded in `HasExistingUseInboxAttributesInPipeline` — thank you, that is the property that makes keeping `InboxConfiguration` alongside the attribute safe rather than just redundant, and I did not know it.

### What I am not claiming

**The Competing pair builds and starts, and neither of its two proven defects remains — but it does not round-trip.** The sender logs three `Decoupled invocation` lines with no error and nothing lands in `QueueData`; the receiver then reports *"There were 0 commands handled by this consumer"*. That is pre-existing, I have not diagnosed it, and I do not think a samples PR should grow a third investigation. Flagging it rather than papering over it.

The Greetings pair is re-proven end to end on the committed source with only an environment variable: Outbox provisioned, Inbox provisioned, delivery 1 handled once, delivery 2 of the same id handled zero times with *"has already been seen"*, Inbox holding 1 row.

### On `.ValidatePipelines()` and the `ChannelFactoryType` spec

Both agreed, and your ranking is right — the spec turns this whole class from "compiles, dies at startup" into a named startup error, and `MsSqlSubscription.ChannelFactoryType` already carries the data. Still holding off on opening issues uninvited; that is now **seven** follow-ups I have offered on this PR and its sibling. Say the word and I will file them, or tell me to stop offering and just file.

iancooper:
Your first point was the important one and it was right. **That is the third time in this PR a comment of mine has promised more than the code delivered** — first the Docker proof that used an uncommitted edit, then "runs against a container without editing source", now "the receiver can be started on its own". Fixed in `1610d5862`.

### The sample now runs from a clean machine, and I proved it that way

I took your suggestion rather than shortening the claim: **the queue table is provisioned by the sample, guarded by `GetExistsQuery`**, in `Events/QueueTableProvisioner.cs`. It makes "either can be started alone" literally true, and it gives `MsSqlQueueBuilder.GetDDL` / `GetIndexDDL` / `GetExistsQuery` **their only production call site in the repository** — they are public precisely because callers must run them, and until now nobody did.

`BrighterSqlQueue.sql` is now **23 lines of portable UTF-8** creating only the database, because a database cannot create itself. The 2018 SQLEXPRESS dump with `FILENAME = N'C:\Program Files\...\MSSQL12.SQLEXPRESS\...'` is gone.

Fresh SQL Server 2022 container, the `.sql` run once, **no manual table creation**, committed source, one environment variable:

```text
receiver started FIRST, on a database with zero tables:
  Created queue table 'QueueData'
  Provisioned Inbox 'InboxMessages' successfully
sender:
  Provisioned Outbox 'Outbox' successfully / Decoupled invocation
delivery 1, message id X : handler ran 1 time
delivery 2, message id X : handler ran 0 times
  warn: UseInboxHandler  Command 01a08bdb-... has already been seen
InboxMessages rows: 1
```

Starting the receiver first is the case that could not work before.

### 2, 3, 4, 6 — all taken

- **`databaseName:` dropped.** Your telemetry argument settles it: pinning a value that only reaches `BoxSpanInfo` beside an overridable connection string silently mislabels `db.name` on every span. A value that is inert-but-wrong is worse than an absent one.
- **Shared constants.** You were right that this is the cheaper option, not the more expensive one — `Events/SampleDatabase.cs` is shorter than the comments it replaced, and a mismatch was silent rather than loud.
- **Comment density.** Cut from ~20 lines to 6. Your reason is the one that convinced me: the mechanism is accurate today and asserted nowhere, so it belongs in the guide, which can be revised when the mechanism changes.
- **The Competing pair is configurable too**, so the directory is coherent; licence year back to 2014; `try`/`finally` with `Log.CloseAndFlushAsync`; and `DepositPost` now takes the transaction provider, so the code demonstrates the sentence above it instead of asserting it.

### What I am still not claiming

**The Competing pair does not round-trip.** Both apps start, are configurable and provision what they need, but `CompetingSender` logs three `Decoupled invocation` lines with no error and nothing reaches `QueueData` — under any topic — while the Greetings pair round-trips against the same table in the same session. Filed as **#4338** with what is ruled out, rather than guessed at here.

### The five follow-ups are filed

The maintainer asked for them rather than more offers:

- **#4334** — validation spec comparing a subscription against its channel factory's `ChannelFactoryType`. Your ranking was right; this is the one that stops the next occurrence.
- **#4335** — `NoExternalBus()` discarding `InboxConfiguration`.
- **#4336** — `InboxScope` inert.
- **#4337** — `CommandProcessor.Call`'s `"fpr"` typo, which also prints `typeof(T)` where the match is on `TResponse`.
- **#4338** — the Competing pair above.
- **#4339** — a test pinning the `ChannelFactory` downcast.
- **#4340** — a README for this sample.

Thank you for pushing on the runnability point rather than accepting a narrowed comment. It turned out to be the difference between a sample that documents a two-step setup and one that just works.

iancooper:
Both headline findings were real. Fixed in `22394a6e3`.

### 1. The documented default could not connect

Right, and **the diagnosis of why I missed it is the sharper half**: I tested through `ConnectionStrings__Brighter` against a container, which is exactly the path that bypasses the constant. The fallback — what a reader hits first, with no environment variable — was the one path never exercised. And your point that it was *previously unreachable* because the receiver threw first is what makes this PR the first time it matters.

`TrustServerCertificate=True` added, with the reason in the XML doc: `Microsoft.Data.SqlClient` defaults `Encrypt` to `true` from 4.0, this repo pins **7.0.2**, and SQL Express presents a self-signed certificate.

### 2. The race, in exactly the scenario the provisioner was added for

Also right, and it undercut the property this PR is built on — "either can be started alone" invites starting *both*. Now one batch with `IF NOT EXISTS` guards plus a catch for **2714**/**1913**, so a genuine race resolves as success rather than as a flake that reads like a broken sample.

**Measured, two receivers started concurrently against a database with zero tables:**

```text
instance A: 0 race errors
instance B: 0 race errors
tables    : __BrighterMigrationHistory, InboxMessages, QueueData
IX_QueueData_Topic present exactly once
```

### 3. Convergence, and the schema disagreement

Taken in full, since I was rewriting the method anyway. The index is guarded independently, so a table that exists *without* its index gets one — including a table left by the old script, whose index had a different name. Both guards test `SCHEMA_NAME()` rather than a literal `dbo`, because `GetDDL` emits an unqualified `CREATE TABLE`: a login whose default schema is not `dbo` would never match its own table and would re-attempt the create on every start.

```text
index dropped by hand, then restart:
  before: 0    after: 1    errors: 0
```

### 4, 5, 6, 8 and the nits

- **The description is rewritten** — you were right that it was the record and that it still listed a file this diff rewrites.
- **The README is in this PR**, not deferred. It documents the changed setup contract in the place a reader will look: script → database, applications → tables.
- **`SampleDatabase.Name` now has a caller** — the default connection string is built from it, so the one value the class failed to centralise no longer is.
- **`Events` declares `Microsoft.Data.SqlClient`** rather than inheriting it.
- **Both receivers call `.ValidatePipelines()`.** This was the best catch of the three smaller ones: my comment cited `ConsumerValidationRules.PumpHandlerMatch`, and that rule only runs behind an opt-in call none of these apps made — **a comment describing a check that was not running**. Now it runs, and the round trip reports `0 validation errors`.
- Host disposed; unused `using` gone.

### 9 — deferred, by the maintainer's call

Adding tests to samples is a policy change wider than a PR whose job is to support a documentation page, so it stays filed as **#4339**. I take the `CLAUDE.md` point and I do not think you are wrong about the cost; it is just not this PR's decision to make.

### 7 — not fixed

`Events` now carries infrastructure as well as contracts. You marked it non-blocking and I have left it, because all four applications need the provisioner and splitting it means a fifth project for a sample. Recorded here rather than silently ignored.

### 10 — already filed

`NoExternalBus` dropping `InboxConfiguration` is **#4335**, alongside `InboxScope` as **#4336**. Seven issues in total came out of this PR and its sibling.

Thank you for splitting the review into parts — the two that mattered were unambiguous, and finding #1 by asking *which path did the testing not take* is the kind of thing I would not have got to on my own.

iancooper:
All of it taken bar one piece, in `6b449d1be`. The item I pushed back on deferring last round was the one you were right about, and for a reason worse than disagreement: **my refusal rested on a misreading of my own PR.**

### 1. The test — and the deferral was wrong on its facts

I wrote *"adding tests to samples is a policy change"*. **The test is not in samples.** You said where it goes — `tests/Paramore.Brighter.MSSQL.Tests/MessagingGateway/`, beside `When_mssql_channel_factory_forwards_scheduler_to_consumers`, which needs no database — and I answered as though it were going in the sample folder. There was no policy question to defer; there was a test to write.

`When_a_subscription_is_not_an_mssql_subscription.cs`, four facts, no container:

```text
Should_throw_when_creating_a_sync_channel                      Passed
Should_throw_when_creating_an_async_channel                    Passed
Should_throw_when_creating_an_async_channel_asynchronously     Passed
Should_create_a_channel_for_an_mssql_subscription              Passed   <- the control
```

The control is there because without it the three would also pass against a factory that threw for everything. And since a test written after the fix proves nothing until it has been seen to fail, I neutered all three `if (rmqSubscription == null)` guards in `ChannelFactory` and re-ran:

```text
Failed!  - Failed: 3, Passed: 1, Total: 4
```

Three red, control green, `src/` reverted. **#4339 is closed by this PR.** #4334 — the validation spec comparing `subscription.ChannelFactoryType` against the resolved factory — stays deferred, because that one really is a design change with its own review surface, and it is the fix that would retire this test for every transport at once.

**What I did not do: the `SampleDatabase.ConnectionString` two-case test.** You are right that the whitespace-not-null decision lives only in an XML doc, and it is a pure function begging to be pinned. But no test project in the repository references `samples/`, and making the framework suite the first one inverts the dependency to pin two lines. If you would rather have it than have that boundary, say so and it is five minutes.

### 2. Parameterized

Both catalog lookups now bind `@queueTable` and `@indexName`; only the DDL interpolates, because an object name cannot be a parameter and that DDL is Brighter's own. The comment says why a constant still gets parameters — the reader copying this file is the one at risk, not this file. `MsSqlBoxMigrationRunner:244` cited as the precedent.

### 3. The index name

Now a named `IndexName(queueTableName)` method whose remarks state the duplication outright: it repeats `QUEUE_TABLE_INDEX_DDL`'s `IX_{0}_Topic`, `MsSqlQueueBuilder` exposes no `GetIndexName` to ask instead, and if that convention moves the guard stops matching, 1913 is caught forever and **a permanently failing statement looks exactly like success**. One place to change, and the dependency named rather than buried in a string.

### 4. Layering — split, and you asked twice

`SampleInfrastructure` now holds `QueueTableProvisioner` and `SampleDatabase`. **`Events.csproj` is byte-identical to master again** — no `MessagingGateway.MsSql` reference, no `Microsoft.Data.SqlClient` — so the assembly every application hands to `AutoFromAssemblies` is back to commands, mappers and the handler. You raised this at review 5 §7 and again here; I had recorded a refusal rather than acting, and a fifth project turned out to be four `ProjectReference` lines and one entry in `Brighter.slnx`. **That solution-file edit is outside `samples/`**, and so is the test; flagging both rather than letting them pass as sample changes.

### 5, 6, 7

- **README**: a paragraph saying neither `Encrypt=false` nor `TrustServerCertificate=True` belongs in production, and what each one actually switches off — the first drops TLS, the second keeps it and accepts any certificate. Plus a line at the top of *The applications* sending readers to the greetings pair first, which is your point about someone running the competing pair and concluding the sample is broken.
- **Description**: re-checked as you asked. The *"behaviour-preserving where arguments were omitted"* phrasing and the stale *"Not in this PR"* entry both went in the 16:14 rewrite; what was left was the provisioner's old path, and the new project and the test are now in it.
- **Nits**: both unused `using System;` gone. `<Nullable>enable</Nullable>` on the new project, and your CS8632 diagnosis was exactly right — proven both ways, since a claim about a warning is worth as little as any other unrun claim:

```text
Events (no <Nullable>), string? probe : warning CS8632
SampleInfrastructure with <Nullable>  : 0 warnings
```

- **The repeated four-line prologue** I have left explicit, and that is a choice rather than an oversight: the two steps are the setup contract the README describes, and a reader who cannot see the provisioner run in `Program.cs` has to be told it happens. Happy to fold it into `SampleInfrastructure` if you would rather have it once.

### Re-run after the move

The move touched every application's composition root and the provisioner's SQL, so the measurements were taken again rather than carried over. Fresh SQL Server 2022 container, `BrighterSqlQueue.sql` once, committed source, one environment variable:

```text
two receivers started CONCURRENTLY on a database with zero tables:
  0 errors either side
  tables: __BrighterMigrationHistory, InboxMessages, QueueData
  IX_QueueData_Topic present exactly once
convergence: index dropped by hand, recreated on next start, 0 errors
round trip: same message id queued twice
  "Received Greeting" 1 time, "has already been seen" 1 time, queue drained
```

---

Six rounds, six real findings, and this one caught me refusing a test on grounds that did not survive reading my own diff. Thank you.

iancooper:
**Finding 1 is right, and #4338 is not a Brighter defect at all — it is this sample's own `TransactionScope`.** You predicted it from the diff; here it is measured, same binary, one connection-string change:

```text
default connection string (Enlist defaults true) : 3 Posts, no errors, QueueData 0 rows
...;Enlist=False                                 : 3 Posts, no errors, QueueData 3 rows
```

That is the whole of it. `Post` opens a `SqlConnection`, `Enlist` defaults to `true`, the insert into the queue table joins the ambient transaction, and the scope was deliberately abandoned — so every message rolled back, silently, exactly as reported. And it explains the asymmetry that made it look like a transport bug: the Greetings pair has no ambient transaction, which is why it round-tripped against the same table in the same session.

**The old comment's premise was the defect**: *"a message is always queued, whether the transaction commits or aborts"* is true of a broker outside your database and false when the broker **is** your database. `scope.Complete()` now, with the comment saying that, and pointing at `GreetingsSender` — `DepositPost` inside your transaction, `ClearOutbox` after it commits — because this is the single best argument for the Outbox that this sample can make.

**The competing pair round-trips for the first time. Measured, six messages, two receivers:**

```text
queued: 6      instance A handled: 2      instance B handled: 4      queue drained: 0 rows
```

So the README's *"does not round-trip"* row is gone and **#4338 can close** when this merges.

### 3. The provisioner — both taken, and the second deletes a problem instead of documenting it

**Two commands, not one batch.** You are right that a swallowed 2714 abandons the rest of its batch, and that "safe to run concurrently" was resting on the winner finishing. Table and index now execute separately, each with its own catch.

**And that let me take your better suggestion**: the index is attempted unconditionally with 1913 treated as success, so `IndexName` is **deleted**. There is no longer a duplicated convention to drift out of step with `MsSqlQueueBuilder` — which is a better outcome than the remark I wrote worrying about the drift.

**The identifier guard is in too**, and you caught me stopping half way: I reasoned about copy-paste injection, parameterized the lookups, and left `GetDDL` formatting the name into `CREATE TABLE [{0}]` where a `]` closes the bracket. Now rejected against `^[A-Za-z_][A-Za-z0-9_]*$` before anything runs.

### 4. Comment volume — the style guide is explicit and I was not following it

Re-read `.agent_instructions/code_style.md`. *"Inline comments are not notes to a reviewer"* and *"do not describe the change relative to how the code used to be"* condemn several of mine outright. Gone: the changelog paragraph in `BrighterSqlQueue.sql`, the twelve-line `IndexName` remark (with the code it guarded), the `TrustServerCertificate` essay, the four-line hosted-service ordering note. The README carries what was worth keeping, and the file:line references in comments are replaced by method names — you were right that `MsSqlBoxMigrationRunner:244` had already moved.

### 2. `GreetingEventHandler` — probed rather than argued, and it stays put

Your three costs are all real, so I tested the premise instead of defending it: moved the handler and its `[UseInbox]` back into `Events`, rebuilt, and started `CompetingReceiverConsole`.

```text
started clean, ran to my timeout, 0 exceptions, ValidatePipelines passed
```

**So the hazard review 2 named is latent, not fatal** — pipelines build lazily and nothing else sends a `GreetingEvent`. It comes down to a judgement between a latent hazard and a conventional layout, and I have kept the move for one reason: the attribute *is* a policy that only one process configures, and putting it next to the process that configures it is what makes the code true. What was actually wrong was the README claiming `Events` still held the handler — **that sentence is fixed, and the asymmetry is now explained where a reader meets it**, which is the part of your finding I think mattered most. `CompetingConsumerCommandHandler` stays in `Events` because it carries no such policy.

### 5 and 6

- **AAA structure** added to the new test.
- **Class naming** — I have left `When_…`, matching all twenty-four files in that folder. If `testing.md`'s `[Behavior]Tests` rule is live, the folder needs a sweep and this file should not be the only one obeying it; happy either way, but not one-off.
- **Three unused `using Events;`** gone — you were right, the only type in that namespace is `IAmACommandCounter`, and `SampleDatabase` moved to `SampleInfrastructure`.
- **Trailing `USE`/`GO`** gone. **PR body** corrected to `SampleInfrastructure/`.
- **The connection-string logging** now has a line in the README. Good catch that the README is what makes it matter — it is the thing telling readers to put an `sa` password in there.

### 5, the part I have not done: provisioner tests

You are right that `EnsureQueueTable` is the one piece of real logic here and that `Paramore.Brighter.MSSQL.Tests` has a live server. But the class lives in `samples/`, so pinning it means the framework test project taking a `ProjectReference` on a sample — no test project in this repository does, and it inverts the dependency. The alternatives are duplicating the SQL into the test (which then pins a copy, not the code) or promoting the provisioner into the gateway, **which is #4334's territory and a design change**. Tell me which you would prefer and I will do it; I did not want to pick that on my own.

### Re-run after all of the above

Every claim re-measured rather than carried over, since the provisioner's SQL and three composition roots changed:

```text
two receivers started CONCURRENTLY on a database with zero tables:
  0 errors either side, tables: __BrighterMigrationHistory, InboxMessages, QueueData
  IX_QueueData_Topic present exactly once
convergence: index dropped by hand, recreated on next start, 0 errors
round trip: same message id twice — handled 1, "has already been seen" 1, queue drained
competing:  6 sent — 2 and 4 handled, queue drained
```

---

Seven rounds, seven real findings. This one is the best of them: it took a "known limitation" I had filed as unexplained and explained it from the diff.

iancooper:
Finding 2 is the sharpest one anyone has made on this PR, and it is my defect. `e1d2e4de1`.

### 1. The stale text is the PR body, and it was already gone

You read a snapshot taken minutes before the rewrite: the *Known limitation* section no longer exists, because **#4338 turned out to be this sample's own defect**. The current body and the README agree, and the README figures are the measurement, not the other way round.

Your diagnostic pointed at exactly the right place, though, and independently of the fix: *"Decoupled invocation" is logged before the send, so no Error line alongside it means `Send` returned normally — the row was inserted and then rolled back.* That is precisely what happened. `Post` opens a `SqlConnection`, `Enlist` defaults to `true`, the insert joined the ambient `TransactionScope` that `CompetingSender` deliberately abandoned. One connection-string token separates the two outcomes:

```text
default (Enlist defaults true) : 3 Posts, no errors, QueueData 0 rows
...;Enlist=False               : 3 Posts, no errors, QueueData 3 rows
with scope.Complete()          : 6 sent, handled 2 and 4 by two receivers, queue drained
```

**#4338 closes with this PR**, and the sample now teaches what the old comment got backwards: `Post` is decoupled from your transaction only when the broker is not your database.

### 2. The comment promised a shared transaction the code never opened — fixed, and proven by breaking it

You are right, and this is the fourth time on this PR that I have written the sentence I wanted to be true next to code that did not make it true. `RelationDatabaseOutbox` attaches a transaction only when `transactionProvider is { HasOpenTransaction: true }`, and nothing opened one — so that `DepositPost` auto-committed on its own connection and shared exactly nothing.

`GreetingsSender` now has the shape you sketched: open, deposit, commit, clear after the commit, with `Rollback` and `Close`. And because a claim about transactional behaviour is worth nothing unread, I made it fail on purpose:

```text
transaction.Rollback()  : Outbox 0 rows
transaction.Commit()    : Outbox 1 row
```

The deposit is inside the transaction now. Before this commit both numbers would have been 1, which is the whole finding in two lines.

### 3. Controls for the async paths — and the non-generic subscription

Taken in full. All three creation methods have a control, plus the non-generic `MsSqlSubscription` that the error message advertises and nothing covered. **Red-proofed in both directions**, since a control that has never failed is just an assertion:

```text
guards neutered (never throw)      : 3 throw-tests FAIL, 4 controls pass
guards forced to throw always      : 4 controls FAIL, 3 throw-tests pass
```

### 4. Test naming

Done: `MsSqlChannelFactorySubscriptionTypeTests`, file name unchanged, which is what `testing.md` actually says. **You are the second round to raise it** — I deferred it last time on the grounds that all twenty-four neighbours use `When_` for the class, and "new file, follow the doc" is the better answer. The helpers no longer shadow the types they construct either.

### Smaller notes

- **`Execute` now takes `params SqlParameter[]`** instead of sniffing the SQL for `"@queueTable"`. You put your finger on why it mattered: this file is written to be copied, so the implicit coupling is the version a reader inherits.
- **The caught 1913 on every start after the first is now stated as deliberate** in the remarks, with "do not fix this by adding a guard" — which is exactly the trap you identified.
- **Ctrl-C** is in the README table row.
- **`OutboxProducerMediator.cs:1388`** — I confirmed it. `resiliencePipeline.Execute(_ => action, requestContext.ResilienceContext)` binds `Execute<TResult>` and returns the delegate rather than invoking it, so the send would silently not happen; the async twin at `:1417` correctly awaits `send(context.CancellationToken)`. Dead today because nothing assigns `RequestContext.ResilienceContext`, live the moment someone does. **Genuinely good catch, and it is a real defect in `src/`.**
- **The gateway should honour `OnMissingChannel.Create`** the way the PostgreSQL one does, so `QueueTableProvisioner` can shrink to nothing. Agreed, and it is not in the filed list.

**Neither of those last two is filed**, and deliberately: opening issues is not mine to decide, and the maintainer has not been asked. They are recorded here so the evidence is not lost — the mediator one especially, since it took tracing an unrelated swallow to find it.

### What I have not done

`EnsureQueueTable` still has no test, for the reason in my last reply: it lives in `samples/`, so pinning it means the framework test project taking a `ProjectReference` on a sample, which nothing in this repository does. Duplicating the SQL into a test pins a copy rather than the code, and promoting the provisioner into the gateway is #4334's territory. It needs a decision rather than a preference.

---

Eight rounds, eight real findings. This one found a false promise in a comment on the single most important line of the sample — the one the guide points readers at to learn the Outbox.

iancooper:
**Finding 1 is a defect I introduced in the commit answering your last round** — the fourth time on this PR that a fix has produced the next finding. `885949b3d`.

### 1. The rollback path could throw over the real exception

Both halves confirmed, and the first is measurable. `ClearOutbox` is the one call in that block doing I/O against the transport, so it is *the* likely failure — and on that path the catch called `Rollback()` on an already-committed `SqlTransaction`. With a simulated dispatch failure after the commit:

```text
raw transaction, ClearOutbox inside the try  : "This SqlTransaction has completed; it is no
                                                longer usable"  — the real exception LOST
provider methods, ClearOutbox outside        : "simulated dispatch failure" — it survives
```

The second half I verified against `src/` rather than by running: `RelationalDbTransactionProvider.Commit()` clears `Transaction`; committing the **raw** one leaves it non-null, so `HasOpenTransaction` stays `true`, `Close()` skips the `Dispose`, and `IsSharedConnection` defaults `true` so the connection is not closed either (`RelationalDbTransactionProvider.cs:20-30`). **`Close()` was doing nothing at all**, in a sample whose subject is that lifecycle.

`GetTransaction`/`Commit`/`Rollback`/`Close` now all go through the provider, and `ClearOutbox` sits after the `finally` where a failure cannot reach a rollback — your suggested shape, with the reason in a comment so it does not get "tidied" back inside the try.

### 2. `.ValidatePipelines()` on both senders

Taken, and the argument for it is better than symmetry: `ProducerValidationRules.PublicationRequestTypeSet` is exactly what catches the `new Publication()` with no `Topic` — **the bug this PR fixed by hand three rounds ago**. So the sample now demonstrates the guard that would have caught its own defect, on the producer side a reader composing an Outbox actually needs.

### 3. The smaller points

- **The empty catch says why it is empty**, at the call site a copy-paster lands on rather than only above the method.
- **`databaseName:`** — you are right that "harmless here" is not "harmless". The README now says it is absent because the connection string carries it and the MSSQL path never reads it, **and** that the MySQL migration runner and Mongo do read it and default it to `"Brighter"`, so it goes back if you adapt this.
- **The control tests dispose their channels** — `using var`, which is the ownership the rest of the sample is teaching.
- **Comment volume** — the `TransactionScope` block is three lines now instead of eight. You and review 7 have both made this point and I keep re-earning it: the README is where the argument belongs, and a sample read as a template should not carry the essay.
- **Test method naming** — left as `Should_…`, matching the folder. Noted as a nit on your side too; if `testing.md`'s `When_[condition]_should_[behavior]` is to be enforced, that folder wants one sweep rather than one conforming file.

### Re-run

```text
fresh container, zero tables: Outbox 1, Inbox 1, greeting handled once, queue drained, 0 errors
competing: 6 sent, handled 3 and 3, queue drained
```

---

Nine rounds, nine real findings, and the last three have each found the defect inside the previous round's fix. **The pattern is not that the fixes are careless — it is that each one moves the risk somewhere the previous round's evidence does not reach.** Round 8 proved the deposit joined the transaction; it did not ask what happened when the *dispatch* failed. That is the question this round asked.

iancooper:
Thank you for splitting it — and for finding that my own argument in the README was wrong about the handler it did not mention. `c59819d28`.

### 3. "Latent rather than fatal" was wrong, and `ValidateOnBuild` is why

You are right, and it is measurable rather than arguable. `CompetingConsumerCommandHandler` takes `IAmACommandCounter`, registered only in `CompetingReceiverConsole`, and `Host.CreateApplicationBuilder` turns on `ValidateOnBuild` under `Development`:

```text
before, GreetingsReceiverConsole with DOTNET_ENVIRONMENT=Development:
  Unable to resolve service for type 'Events.IAmACommandCounter'
after, all three applications under Development: start clean, 0 unresolved
```

So a reader running the greetings pair from an IDE profile got a failure naming a type from the competing sample. **The handler and the interface now live in `CompetingReceiverConsole`**, `Events` is contracts only, and the README states the rule both handlers follow — *handlers live with the process that owns their dependencies* — instead of defending one exception. The asymmetry review 7 objected to is gone, and it turns out the right resolution was to move the *other* one.

### 1. My `ValidatePipelines` comments were wrong

Added in the commit answering your last round. `ProducerValidationRules` is three specs — `PublicationRequestTypeSet`, `PublicationRequestTypeImplementsIRequest`, `WrapTransformResolvable` — and **none looks at `Topic`**; `PublicationRequestTypeSet` interpolates `p.Topic` into its own failure message, which is as clear a sign as you could want that a topicless publication reaches it and passes.

The failure this PR fixed comes from `MsSqlMessageProducerFactory.Create`, which runs at `CompetingSender/Program.cs:48` before `AddBrighter()` is called at all. The comments now describe what the rules actually check. `ValidatePipelines()` stays in both senders — it just is not the guard I claimed.

I wrote *"the sample now demonstrates the guard that would have caught its own defect"* one round ago. It would not have, and that sentence is exactly the kind of satisfying claim this PR keeps teaching me to distrust.

### 2. Nothing handed `Events` to `AutoFromAssemblies`

Also right, and worse than a README inaccuracy: all four called the **no-argument** overload, which scans `AppDomain.CurrentDomain.GetAssemblies()` — *loaded*, not referenced. It worked only because each `Program.cs` happens to construct a generic over an `Events` type before the call. **Reordering two lines would have silently unregistered every mapper**, with no error anywhere.

All four now pass their assemblies explicitly, with the reason in a comment, and the README sentence is true instead of accidentally true.

### 4 and 5

- **The logging claim** is now stated as what it is: true of the receivers by configuration, and true of `GreetingsSender` **by ordering** — that application does set `MinimumLevel.Debug()`, and the producer registry is simply built inside the `AddProducers` callback, before Brighter swaps the logger factory into `ApplicationLogging`. Move that construction after `Build()` and the `sa` password lands on stdout. Said in the README, in those terms.
- **The export is promoted into step 1**, framed as required unless you are on Windows with SQL Express, since the default is `Server=.\sqlexpress` and step 1 already points everyone at `localhost,1433`. That is the `Named Pipes Provider` error you predicted, and it was one section too far down the page.

### Nits

- `README.md` is in `Brighter.slnx` beside the `.sql`.
- **Three unused Serilog references dropped from each receiver** — neither uses Serilog any more.
- **`CompetingSender` stops itself.** `IHostApplicationLifetime.StopApplication()` after the scope completes: `CompetingSender 6` now exits in **1.3 seconds** instead of waiting for Ctrl-C, and the README no longer has to explain that.
- **Test methods renamed** to `When_[condition]_should_[behaviour]`, file named after one of them. **Three rounds have now asked** — reviews 7, 9 and yours. The folder is still mixed; this file is not.
- **The 1913 catch** says an index of that name with a *different definition* is accepted rather than corrected. You are right that this is not quite convergence, and a sample should not claim more than it does.

### Re-run

Both changes could have silently stopped handlers registering, so the numbers were taken again rather than carried:

```text
Outbox 1, Inbox 1, greeting handled once, queue drained, 0 errors
competing: 6 sent, handled 3 and 3, queue drained
CompetingSender 6: exits on its own in 1.3s
```

---

Ten rounds. Three of the last four found defects in the previous round's fix, and this one found the flaw in an *argument* rather than in code — which is the harder class, because nothing compiles a rationale.

iancooper:
Finding 1 is right and it is the worst kind of documentation error — a **wrong mental model** stated next to a real trap, where believing it leads you to the wrong conclusion about the trap. `8b96bf7b4`.

### 1. The named list is additive, and I taught it as an allow-list

`ServiceCollectionBrighterBuilder.AutoFromAssemblies` always enumerates `AppDomain.CurrentDomain.GetAssemblies()` and then `Concat`s `extraAssemblies` (`:118-124`). So naming `Events` buys a **guarantee that it is scanned whatever the load order**, and nothing else.

Your point about *where* it sits is the one I have taken hardest. The paragraph above explains the `IAmACommandCounter` / `ValidateOnBuild` trap — so a reader believing the list was an allow-list would conclude that naming assemblies was the fix, **when the only fix is the handler move, precisely because every loaded assembly is scanned anyway.** The comment and the README now say so in as many words, including that last clause.

`typeof(Program).Assembly` dropped from both receivers: the entry assembly is always in the `AppDomain` list and passes the filters. You are right that it is noise in code whose job is to be copied.

### 2. The index guard now tests what the index *is*

Taken, and you are right that the name is not the property I care about. The guard joins `sys.indexes` / `sys.index_columns` / `sys.columns` for any non-primary-key index whose **leading column is `Topic`** on that table. That duplicates no convention, cannot silently stop matching, and **removes the caught-1913-on-every-start steady state** rather than defending it. Measured both directions:

```text
index dropped by hand, restart        : recreated, 0 errors
index over Topic under a DIFFERENT name (IX_LegacyTopicName), restart
                                      : recognised, NOT duplicated, 0 errors
```

The second case is the one the name guard could not do, and it is also the "different definition is accepted" hole the old comment admitted to, closed from the other end.

**1205 is handled too** — a single retry. You are right that "safe to run concurrently" was luckier than the code guaranteed: concurrent DDL surfaces as a deadlock as readily as `2714`/`1913`, and my catch only knew about the latter.

### 3. The connection-string logging should be an issue, not a README paragraph

**Agreed, and this is the second of two library defects this PR has surfaced and documented rather than fixed** — the other being `OutboxProducerMediator.cs:1388` returning the delegate instead of invoking it. *"Documenting it in a sample README is how it stays unfixed"* is exactly right.

**I have not filed it.** Opening issues is outward-facing and the maintainer has not been asked — the seven that exist (#4334-#4340) were filed on instruction. Both are recorded here with their evidence so nothing is lost, and I will file them the moment that is wanted.

### 4, 6, 7, 8

- **`InboxTable`** now says why it departs from the framework's `"Inbox"` default, and **`Name`** says what it is: the database the `.sql` creates and the default connection string points at, **not** "the database we are connected to" — set `ConnectionStrings__Brighter` and the connection string wins while the constant stays put. Good catch; that is exactly how a constant becomes a lie.
- **The competing demo's shape** is in the README: one commit, then a burst, and `readpast` meaning receivers never block on the open transaction. Plus your MSDTC point, which is the better half — **adding your own `INSERT` on a second connection inside that scope promotes to MSDTC and off Windows that fails**, which is precisely what the Outbox route avoids. That belongs in the section most likely to be copied.
- **The description bullet was describing a snapshot from partway through the work.** You are right: the base `GreetingsSender` has no `UseBoxProvisioning` call and no such comment — it built a bare `ServiceCollection`, took `.Create()` and called `Post`. Rewritten to describe the base, with the correction stated rather than quietly swapped.
- **Nits**: the blank line the BOM removal left in `CommandCounter.cs`, and the double negative in `GreetingEventHandler`. `[GeneratedRegex]` I have left — it needs the containing type to be `partial`, which is a bigger change to a static sample class than the nit warrants.

### 5. The provisioner still has no test, and the identifier branch is the strongest case yet

You have moved me on this: the regex is a **security control**, and a security control with no test can be widened later by someone who does not know what it was for. The obstacle is unchanged — the class is in `samples/`, and no test project here references a sample — so this is a decision rather than a preference, and it is now on the list I have put to the maintainer.

### Re-run

```text
fresh container, zero tables: 3 tables, exactly one Topic index, 0 errors
convergence: index dropped by hand, recreated on next start
round trip: greeting handled once, Outbox 1, Inbox 1, queue drained
competing: 6 sent, handled 5 and 1, queue drained, sender exits on its own
```

---

Eleven rounds. This one found a wrong mental model rather than a wrong line — and a mental model is the thing a sample actually transmits.

iancooper:
**Finding 1 is right, and the code I shipped an hour ago was strictly worse than what it replaced.** `2552dccd2`.

### 1. The retry threw `ArgumentException` instead of retrying

`SqlCommand.Dispose()` does not detach parameters, so the retry handed the first command's `SqlParameter` instances to a second one. Proven with a control, and no deadlock needed to see it:

```text
same parameter instances on a second command : ArgumentException — "The SqlParameter is already
                                               contained by another SqlParameterCollection"
parameters built fresh for the attempt       : accepted
```

That exception escapes **both** filtered catches, so **the one path written to survive a race would have turned it into an unhandled startup exception.** I added a retry to make "safe to run concurrently" airtight and made it less true than before. The parameter is now built inside `ExecuteOnce`, so each attempt owns its own, with the reason in a comment — otherwise it reads as incidental and the next person inlines it back.

Both smaller points taken: the comment claimed a back-off that does not exist, and a second 1205 propagating is now stated rather than implied.

### 2. `GetTransaction()` outside the `try`

Taken. It opens the connection and begins the transaction, so a failure there skipped the `finally` and the `Close()` entirely — in the file that is explicitly a copy-paste template. One line inside the `try`.

### 3. The README promised something a reader cannot reproduce

**This is the finding I am most glad you made**, because I had been measuring idempotence all along with a step that was never written down. `GreetingEvent` takes a fresh `Id` per construction, so running the sender twice is two messages and both are handled — a reader following the README concludes the Inbox is broken, exactly as you say.

The README now says that plainly and gives the SQL I had actually been running:

```sql
INSERT INTO QueueData (Topic, MessageType, Payload)
SELECT Topic, MessageType, Payload FROM QueueData;
```

Followed literally from a clean database: **handled 1, "has already been seen" 1.** The applications table points at that section rather than promising the behaviour outright.

### 4 and 5

- **`CompetingSender`'s own unused Serilog references** — dropped. I removed exactly those three from both receivers last round and missed the sender, which is precisely the half-fixed-file pattern the sibling PR exists to complain about.
- **The Ctrl-C line** is gone for the sender, which now stops itself.
- **Comment tone** — you are right, and `.agent_instructions/documentation.md` is explicit. The *"it used to abandon it, on the stated grounds that…"* framing is the PR description leaking into the source; a reader in two years needs what the code does, not what it did. Rewritten in three places. Git carries the history.

### 6. The connection-string logging

Agreed, and it is now the **third** library defect this PR has surfaced. All three are recorded with their evidence and **none is filed**, because opening issues is the maintainer's call and #4334–#4340 were filed on instruction:

| | |
|---|---|
| connection string logged at `Debug` | `MsSqlMessageQueue.cs:34`, `:254` — redact, or log server and database via `SqlConnectionStringBuilder` |
| `Execute(_ => action, …)` returns the delegate | `OutboxProducerMediator.cs:1388` — the send silently never happens; the async twin is correct |
| gateway ignores `OnMissingChannel.Create` | the reason `QueueTableProvisioner` exists |

Your framing of the first is the one I would put in the issue: *"it doesn't leak today because of an ordering accident"* is not a property anyone should rely on.

### Test coverage

Your version of the ask is the strongest yet — **a concurrent `EnsureQueueTable` against the test container, asserting one table, one index, no exception** — because it covers the retry path, and the retry path is what just broke. It is still blocked on the same thing: the class lives in `samples/` and no test project here references a sample. That is a decision rather than a preference, and it is with the maintainer along with whether these three issues get filed.

### Re-run

```text
two receivers concurrently on zero tables : 3 tables, one Topic index, 0 errors
round trip                                : greeting handled once
the README's own idempotence recipe       : handled 1, "already been seen" 1
competing                                 : 6 sent, handled 1 and 5, drained
```

---

Twelve rounds. Five of them have now found the defect inside the previous round's fix, and this one found it in a *hardening* change — which is the sharpest version of the lesson: the retry was added to make a claim airtight and made it false instead.

iancooper:
All five taken, `23c431b9a` — and one of them caught something about my own method rather than the code.

### 1. `GetExistsQuery` ships and I reimplemented it silently

Fair, and the `<remarks>` naming `GetDDL` and `GetIndexDDL` while staying silent on the third made it worse: a copy-paster would not have known the built-in existed. It now says why this version differs — `GetExistsQuery` formats both names into the SQL and defaults the schema to a literal `dbo`, which a login whose default schema is not `dbo` would never match.

### 2. `EnsureQueueTable` failing before any logging exists — and the control caught me too

Taken, and it is the right thing to fix in *sample* code specifically: this runs before the host is built, so in three of the four applications there is nothing to log into, and the commonest first-run mistake was a bare `SqlException`. Measured against an unreachable server:

```text
Could not provision 'QueueData'. Check the server is reachable and that BrighterSqlQueue.sql
has been run; set ConnectionStrings__Brighter to point somewhere else. SQL Server said: …
```

**And the first run of that control did not print it** — because `dotnet run --no-build` had handed me a stale copy of an application I had not rebuilt after changing `SampleInfrastructure`. A green-looking probe over a binary that does not contain the change is the same failure as the harness two rounds ago on the sibling PR, in a different costume. Worth writing down.

### 3. The duplication SQL takes `TOP 1`

Right, and it is the kind of thing a README snippet gets used for in ways the surrounding sentence does not cover. Against the competing demo's backlog the statement I wrote doubles the queue. Fixed, with a line saying why `TOP 1` is there.

### 4. Resolving a transient from the root provider

Taken. `AddProducers` registers the transaction provider `Transient`, so the root container was tracking a disposable until shutdown. Harmless as written — one instance, `Close()` is called — but this file is teaching the Outbox pattern to someone whose provider will be scoped to a `DbContext`, and `using var scope = host.Services.CreateScope()` is what they should carry away. The comment says that rather than leaving the scope looking like ceremony.

### 5. Test nits

All three: the shared `Paramore.Brighter.Base.Test.Requests.MyEvent` rather than a private redefinition in a project that already references it; `await using` for the two `IAmAChannelAsync` controls, since `using var` picks the sync path on a type whose intent is async; and the file renamed to the happy-path method, which is what `testing.md` asks when a class holds several.

### 6. The provisioner belongs in the gateway

**Agreed, and you have put the argument better than the issue does**: if that behaviour is worth 150 lines it is worth being reachable, and in `Paramore.Brighter.MessagingGateway.MsSql` the MSSQL test project already has SQL Server in CI *and* `OnMissingChannel.Create` would stop being inert — one change closing both. That is the third library-side defect this PR has surfaced without fixing, alongside the connection-string logging and `OutboxProducerMediator.cs:1388`.

You are also right that widening here would be the wrong call. **All three are recorded with their evidence and none is filed**, because opening issues is the maintainer's decision.

### Re-run

```text
two receivers concurrently on zero tables : 3 tables, one Topic index, 0 errors
round trip                                : greeting handled once
competing                                 : 6 sent, handled 3 and 3, queue drained
unreachable server                        : the new message, not a stack trace
```

---

Thirteen rounds. This one is the first where nothing found was a defect I had introduced — which after five consecutive rounds of exactly that is the most encouraging thing in it.

iancooper:
`c2872cb7d`. The mapper is a real defect and I have answered the design question rather than deferring it.

### The design question: `Complete()` stays, the experiment becomes reproducible

You are right that `scope.Complete()` deletes a teaching point rather than correcting it, and right that `Enlist=False` would keep the original demonstration and make it true. I have kept `Complete()`, and the reason is about what a *skimmer* takes away: an application whose logs show messages surviving an aborted transaction reads as "Brighter loses atomicity", which is the opposite of what the sample is for. The competing pair's subject is competing consumption; the transaction interaction is the lesson, not the behaviour.

So the README now carries the whole experiment, each row measured rather than reasoned:

| | `Complete()` | rows in `QueueData` |
|---|---|---|
| default connection string | no | **0** |
| default connection string | yes | **3** |
| `…;Enlist=False` | no | **3** |

with the point your review made explicit: **that third row is the decoupling a broker outside your database gives for free, bought back by opting out of enlistment — and losing, in exchange, any guarantee that the message and your own write agree.** Which is exactly what the Outbox keeps, and why `GreetingsSender` is the other half of the sample.

If you would rather the code took the `Enlist=False` route and left the scope abandoned, say so and I will swap them — the table stays true either way.

### 1. The hardcoded topic

Real, and the worst place for it: `SampleDatabase`'s own doc comment warns that a mismatch here is silent, and this is the one file the centralisation never reached. `publication.Topic` now, matching `GreetingEventMessageMapper` next door. Verified rather than assumed — the queued rows carry `Topic = 'multipleconsumer.command'` and the pair still round-trips, 6 sent and handled 4 and 2.

### 2 and 3. Project files

- **`GreetingsReceiverConsole` names `Microsoft.Extensions.Hosting` explicitly.** Dropping `Serilog.Extensions.Hosting` last round left it resolving `Host.CreateApplicationBuilder` through a transitive — alone among the four. That is the half-fixed-file pattern again, and I created it one round ago.
- **Dead references trimmed**: `Configuration.Binder` from all four, and DI plus both Configuration packages from `Events`. Your framing is the right one — *"Events is contracts only"* should be visible in the project file, not only asserted in a README. `Events.csproj` is now five lines and one `ProjectReference`.

### The smaller notes

- **`ArgumentException` escaped the wrapper.** Caught now, alongside `SqlException`: a malformed connection string throws it from the `SqlConnection` constructor, and that is exactly as common a first-run mistake as an unreachable server — so it was hitting the bare stack trace the wrapper exists to prevent. Good catch; I fixed the failure I had reproduced and not the one I had not.
- **`builder.Logging.AddSerilog()`** rather than `AddSingleton<ILoggerFactory>`, which worked only because last registration wins.
- **The `export` block appeared twice** — one is now a cross-reference.
- **The split is arbitrary**, and the README says so rather than naming a pair: 3 and 3, 5 and 1, 2 and 4 across three runs. A reader getting a fourth split should not wonder whether something is wrong.

### The connection-string logging

**Third round to raise it, and I agree with all three of you.** *"One refactor away from an `sa` password on stdout"* is a bad property to document and leave, and the fix — `SqlConnectionStringBuilder` for `DataSource`/`InitialCatalog`, or redaction — is small.

It is **not filed**, along with `OutboxProducerMediator.cs:1388` and the gateway ignoring `OnMissingChannel.Create`. Opening issues is the maintainer's call and #4334–#4340 were filed on instruction; all three are recorded with their evidence and can be filed the moment that is wanted.

---

Fourteen rounds. The two things this one found in my own work were both created in the previous two commits — the transitive `Hosting` reference and the unwrapped `ArgumentException` — which is the pattern this PR has been teaching all along: **the fix is where the next defect lives.**

iancooper:
Both behavioural findings were real. `2ad94e131`.

### 2. A name collision was reported as success

You are right that the probe and the swallow disagree: `IF NOT EXISTS` looks only in `sys.tables`, while **2714 fires for any object-name collision**. Measured with a view named `QueueData` — and the before-state came free, because the first run used a stale binary I had not rebuilt:

```text
before : "Cannot create index on view 'QueueData' because the view is not schema bound"
after  : "'QueueData' was not created and does not exist as a table. Something else in this
          database owns that name."
```

The "before" is exactly the first-run confusion the wrapper exists to prevent, arriving one layer further on and naming something the reader never mentioned. `EnsureQueueTable` now re-probes after the create, so **"lost the race" and "someone owns that name" stop looking identical.**

### 3. The count printed before the pump drained

Correct, and it is the whole output of that demo. `RunStuff` was registered after the dispatcher, and hosted services stop in **reverse** registration order. Swapped, with the reason in a comment so it does not get "tidied" back.

**Verified after: 6 queued, 6 handler lines, the count reports 6, queue drained.** I could not produce a deterministic *before* — it needs a message in flight at shutdown — so this one rests on the ordering guarantee rather than on a failing run, and I would rather say so than imply a measurement I do not have.

### 4 and 6

- **`RunCommandProcessor` is a `BackgroundService`.** You are right about what it teaches: work belongs in `ExecuteAsync`, not in the call that starts the host. The vestigial `await Task.CompletedTask` went with it.
- **The README was handing the reader the weaker concession.** `Encrypt=false` drops TLS; `TrustServerCertificate=True` keeps it and skips only the certificate identity check — so the export now matches both the code default and the README's own explanation two sections down. **Verified verbatim against a container**, and used for every measurement in this round.
- **The Outbox conclusion, stated twice** in consecutive paragraphs — trimmed.

### 1, 5 and 7 — the follow-ups, and finding 1 is the sharpest

**`MsSqlQueueBuilder.GetExistsQuery` formats both names into `'{0}'`/`'{1}'` string literals**, so a name containing a single quote injects there — in framework code, not a sample. Your framing is the part I would put in the issue: *this PR makes itself the class's first production call site, and the guard should not have to be re-derived by every caller.* That also answers finding 5 the same way, as you say: in `MsSqlQueueBuilder` the guard would be testable for free, which is where the provisioner's logic wants to live anyway.

That now makes **four** library-side items this PR has surfaced and not fixed:

| | |
|---|---|
| `GetExistsQuery` interpolates both names into string literals | `MsSqlQueueBuilder.cs:84` |
| connection string logged at `Debug` | `MsSqlMessageQueue.cs:34`, `:254` — raised by three separate rounds |
| `Execute(_ => action, …)` returns the delegate rather than invoking it | `OutboxProducerMediator.cs:1388` |
| the gateway ignores `OnMissingChannel.Create` | which is why `QueueTableProvisioner` exists |

Plus **7**, the dead `new Connection<GreetingEvent>(...)` / `MsSqlInputChannelFactory` API in the gateway's own README — which, as you note, is the first doc a package-page reader hits and is the same defect this PR fixes in the sample.

**None is filed.** Opening issues is the maintainer's call and #4334–#4340 were filed on instruction; all five are recorded with evidence and can go up the moment that is wanted.

---

Fifteen rounds. The thing this one is worth remembering for: **the stale binary that wasted a run two rounds ago handed me a free control this time** — the "before" behaviour I could not otherwise have captured, because I had already fixed it.

iancooper:
Three findings, and the first is the defect class this PR keeps teaching me. `ccfeffd18`.

### 1. The comment described behaviour the method did not have

You are exactly right, and it is worse than a wrong comment: **a sample's comments are the payload**, and this one asserted the precise thing the code failed to do. No `await` in the body, so `BackgroundService.StartAsync` ran the scope, the loop and `StopApplication()` to completion inside the call that starts the host.

`await Task.Yield()` now makes the claim true, with the reason stated so it does not look like ceremony and get deleted. And `stoppingToken` is threaded into the loop — you are right that Ctrl-C during the shipped 250-message run did nothing at all.

### 2. The index path had none of the re-probe I had just given the table

This is the sharpest of the three, because I fixed exactly this hazard for the table **one commit earlier** and did not carry it across. The guard asks *is there a non-PK index leading on `Topic`?* while `GetIndexDDL` creates one by **name**, so the two disagree in exactly the case that matters. Measured by planting `IX_QueueData_Topic` over `MessageType`:

```text
before : provisioning "succeeds", no signal anywhere
after  : "No index leading on Topic exists for 'QueueData', and one could not be created.
          An index of that name over different columns already exists."
```

**Making that work required undoing one of your own nits**, which is worth flagging: widening the friendly-message wrapper to catch `InvalidOperationException` made it swallow the precise messages the re-probes exist to produce and replace them with the generic one. The wrapper now covers **only the connect** — the failure that genuinely has no logging to fall into — and everything after it reports itself. Both paths measured:

```text
wrong-column index : "No index leading on Topic exists..."
unreachable server : "Could not connect in order to provision..."
```

### 3. A credential property that depends on construction order

Agreed — *"worth removing rather than documenting"* is the right standard for a file whose purpose is to be copied. `MinimumLevel.Override` for `Paramore.Brighter.MessagingGateway.MsSql` at `Information`, so Brighter's own `Debug` output survives where it is harmless. Verified: **0 occurrences of the password** in the sender's output.

### 5. The second defect is now pinned too

You were right that only one of the three was. `MsSqlMessageProducerFactory.Create` throwing on a topicless `Publication` needs no database either, and **nothing in the repository touched `MsSqlMessageProducerFactory` at all** before this. Two cases — the throw and a control for the topic-present path — red-proofed by removing the guard in `src/`: 1 failed, the control passed, reverted.

### 4 and 6

- Stray indentation, `Convert.ToInt32` instead of unboxing an `object?`, and the `Serilog.Extensions.Logging` using that should have left with `SerilogLoggerFactory`. The `InvalidOperationException` nit is in too, at the connect where it belongs.
- **The description's "23 lines" was wrong** — the file is 13. Corrected; you are right that the description is doing explanatory work here and should not be sloppy about its own numbers.
- **Your `ValidatePipelines` observation for #4335 is a genuinely new one**: the validator builds its pipelines with the global `InboxConfiguration` while the runtime `NoExternalBus` path never receives one, so in `GreetingsReceiverConsole` the validator inspects a pipeline the runtime will not build. Harmless here because the explicit `[UseInbox]` makes them converge — but it is the same root cause, and it belongs on that issue. Recorded, not filed, with the other five.

### Re-run

```text
two receivers concurrently on zero tables : 3 tables, one Topic index, 0 errors
greeting handled once; sender output carries no password
competing 6 sent, count reports 6, queue drained
```

---

Sixteen rounds. This one found a comment that lied, an asymmetry between two halves of a fix I wrote an hour apart, and a nit of yours that conflicted with a finding of yours — all three of which are only visible to someone reading the code as a whole rather than as a diff.

iancooper:
*"I found no functional bugs"* is the first time that has been said on this PR, and the four things you did find are worth more than the phrase. `a60b0e9c0`.

### 5. Cancellation that committed anyway — mine, from last round

The loop honoured `stoppingToken` and then `scope.Complete()` ran unconditionally, so **a cancelled run committed whatever it had reached**. You put it exactly right: the two disagreed about what cancellation means. I added the token check one round ago in response to another finding and did not follow it through to the commit. `Complete()` is now conditional.

### 4. The 60-second default

Real, and the shipped profile is what makes it real: 250 posts, each a round trip, all inside one scope that **now must complete**. Against a remote container that is reachable, and it aborts with a `TransactionAbortedException` that explains nothing — in the one demo whose entire point is now that the scope commits. Five minutes, explicitly, with the reason.

### 1. `Close()` does not close the connection

You are right and the comment was mine. `IsSharedConnection => true`, so `Close()` skips the connection, and after `Commit()` the transaction is already null — it is a no-op end to end, and what actually releases the `SqlConnection` is the scope's disposal. In a teaching sample that is worse than saying nothing. The comment now says what releases the connection and where `Close()` does earn its place: the rollback path.

### 3. Four copies of two predicates

Taken, and your reason is the one that matters: **each pair had to stay identical or the re-probe would throw against an object it had just created successfully.** That is a live drift risk, not tidiness — and I created it two commits ago by adding the re-probes without factoring the predicates out. One `const` each now, composed into both the `IF NOT EXISTS` guard and the check. Verified on a second start, where both re-probes run against existing objects: 0 errors.

### 6, 7, 8

- **`TODO(#4335)`** beside the attribute, so the dependency is greppable from the code and not only from the README. You are right that nothing would have said so when the issue is fixed.
- **The password note** — an `export` reaches shell history and every child process, `sqlcmd -P` reaches the process table. `SQLCMDPASSWORD` and `dotnet user-secrets` named. Worth its space in a file that exists to be copied.
- **The meta-commentary is gone**: *"the race backstop it is described as"* (described where? not in that file), *"Measured, before and after the move"*, *"See the README, and GreetingsSender for the Outbox answer"*. Your diagnosis is exact — in committed source they read as an argument with a reviewer who is no longer present. The measurements belong in the PR description, and that is where they now only live. The API-constraint comments you singled out as the good ones all stay.

### 2 and 9 — the two that are not mine to close

**The provisioner test.** Five rounds have now asked, and your pure-unit version — `EnsureQueueTable(conn, "Queue];DROP")` throws `ArgumentException`, no database, no fixture — is the cheapest anyone has proposed. It is still blocked on the same thing: the class is in `samples/`, and no test project in this repository references a sample. **That is one decision, and it is the maintainer's**; review 15 pointed out that moving the guard into `MsSqlQueueBuilder` would answer it and fix a real injection hole in framework code at the same time, which is the version I would argue for.

**Scope, and the `Complete()` reversal.** You are right to name it as a maintainer decision rather than an objection, and I have said the same in the reply to review 14: the argument that the original premise is false for a database broker is evidenced, but it is still a change to what a sample teaches. **It stays flagged and unmerged until whoever owns these samples nods.** The `Enlist=False` alternative is one edit away and the README's table is true either way.

### Re-run

```text
fresh: 3 tables, one Topic index, 0 errors
second start (both re-probes against existing objects): 0 errors
greeting handled once; competing 6 sent, count reports 6, queue drained
```

---

Seventeen rounds, and this is where I stop pushing. Every remaining item is a decision — the two above, plus six library defects gathered and unfiled and two merges. **Nothing further should change here without the maintainer, because the next change would be answering a review with another review's worth of new surface.**

iancooper:
**Four of the five conventions are done in `2bb052a6c`** (rebased onto `master`, which had moved). (1) is yours to take and I have not taken it; (5) is recorded and not filed.

### 2. `TODO(#4335)` — the style guide settles it, and against me

I added that line last round answering review 17's finding 6, so this is two reviewers disagreeing and a document deciding. `code_style.md:40` is unambiguous and the inline reference is gone. The README keeps its markdown link, which is where the traceability belongs.

### 4. Licence headers

Both new test files now carry `Copyright © 2026 Ian Cooper <ian_hammond_cooper@yahoo.co.uk>`, matching the sibling in the same folder. `QueueTableProvisioner.cs` and `SampleDatabase.cs` had `2014` copied from the template; both now say 2026. For what it is worth, `Copyright © 2026` is already the plurality in `src` and `tests` at 672 files, so the template is the outlier rather than the convention.

### 6. `s_identifier`, and the stray `Events` file

`^[A-Za-z_][A-Za-z0-9_]{0,127}$`, so the guard's message is the one you get rather than SQL Server's. **Measured with a control on both sides**, because a bound that rejects everything would look identical to a bound that works:

```text
QueueTable (shipped)   len=  10  old=True  new=True
128 chars              len= 128  old=True  new=True
129 chars              len= 129  old=True  new=False
bad char ]             len=  11  old=False new=False
```

The message now reads *"is not a plain SQL identifier of 128 characters or fewer"*.

`Events/acceptance_test_message_body.txt` is gone. It was a dump of one `GreetingEvent` body with a `MessageType=2` header, referenced by nothing in the repository, and it survived a move it should have gone with.

### 3. The comments that pin another assembly's internals

All three trimmed, and you are right about which half was load-bearing:

- The two `ValidatePipelines` comments no longer enumerate what producer-side validation does and does not cover; they say why the call is there and, in `GreetingsSender`, why it is last.
- The `Close`/`IsSharedConnection` sentence is gone. What is left — *"Commit clears the provider's transaction, which is what keeps the Rollback below a no-op"* — is the part a reader of this sample needs, and it is about this file rather than about `RelationalDbTransactionProvider`.
- The `QueueTableProvisioner` remarks keep the conclusion and drop the comparison: the check asks `SCHEMA_NAME()` because the `CREATE TABLE` it guards is unqualified and so is everything the gateway then issues, so the guard has to agree. The `GetExistsQuery` alternatives-considered paragraph is gone.

I have not asked for a samples exemption from `code_style.md:43`. If one is ever wanted it should be written down rather than inferred from this PR.

### 6, first bullet — the repeated preamble, which I would keep

You framed this as conditional on my agreeing, and I do not. The four `Program.cs` files are read one at a time by someone who has opened exactly one of them, and the connection string → `EnsureQueueTable` → `RelationalDatabaseConfiguration` sequence is the thing they came to see. Folding it into `SampleInfrastructure` buys tidiness across files nobody reads together and costs visibility in the file they do. Happy to be overruled.

### 1. Where `QueueTableProvisioner` lives — yours, and three reviews now agree

I am not moving it on my own authority: it leaves `samples/`, it adds public surface to `Paramore.Brighter.BoxProvisioning.MsSql`, and `AddMsSqlQueue(configuration)` is an API name. But the argument is right and it is the third time it has been made, so recording where it stands: the code is needed by every MSSQL transport user rather than by this sample, it is the most intricate thing here and the only untested thing here, and moving it puts the concurrent-start behaviour I verified by hand into a CI suite that already runs SQL Server. Say the word and it is a small change.

### 5. Both worth issues, neither filed

`MsSqlMessageQueue.cs:34` logging the connection string at Debug, and `:169` interpolating the topic while `:215` and `:236` parameterize. Both recorded with evidence; filing is your call, as #4334–#4340 were.

---

### Verification

The two new test classes run in **51 ms with no database** — 9 facts, 0 failures, on both `net9.0` and `net10.0`:

```text
Passed!  - Failed: 0, Passed: 9, Skipped: 0, Total: 9, Duration: 51 ms
```

All six sample projects and `Paramore.Brighter.MSSQL.Tests` build with 0 errors after the rebase.

iancooper:
**Green at `2bb052a6c`: 24 pass, 0 failing.** Recording how `sqlserver-ci` was settled, because this one was *not* covered by the samples-only re-run rule and I nearly mis-read it twice.

`sqlserver-ci` and `rabbitmq-sync-ci` both went red on the first run. `sqlserver-ci` **is reachable from this diff** — the PR edits `tests/Paramore.Brighter.MSSQL.Tests/` — so it did not get waved through.

**What the log actually said**, 1 of 247:

```text
Paramore.Brighter.MSSQL.Tests.Outbox.Binary.Sync
  .WhenStoringAMessageWithARelativeDataSchemaItShouldBeReadBack
    Microsoft.Data.SqlClient.SqlException : A severe error occurred on the current command.
      at Paramore.Brighter.MSSQL.Tests.Configuration.EnsureDatabase (Configuration.cs:45)
```

It failed in **database setup at `00:00:01.51`**, not in an assertion, and the Azure SQL Edge container log shows the race that produced it — `[DBMgr::FindFreeDatabaseID] Next available DbId EX locked: 5` on spid51 and `: 6` on spid52 at the same instant, the two TFM test hosts creating a database concurrently against one container.

**The failing test is not one of mine**; it arrived from `master` in the rebase. The 247 is master's 238 plus the 9 added here.

**A re-run of the failed jobs on the same commit passed, 24 of 24.** By the rule this repository has been using — *two reds on one commit is a defect, a red that clears on a re-run of the same commit is a distribution* — that settles it as the latter, in the same test-isolation-racy family as the `postgres-ci` and `dynamo-ci` cases.

**One connection I am not waving away:** this PR adds 9 tests to that project, and adding tests changes the timing of a race. That is weaker than a code change and it is not nothing, so it is written down rather than argued away.

**And a correction to my own method, which is the part worth keeping.** My first reading of the log was *"no tests ran at all"* — I had grepped for the execution markers, found none, and was about to conclude the step died before running. Running the same grep against a **passing** `sqlserver-ci` job returned the same count, so the absence was an artefact of my filter and proved nothing. The two-way control is what caught it; a control that only confirms absence confirms nothing about the grep.

PR #4343, opened 2026-09-12
feat(mssql): the gateway provisions its queue table, as the PostgreSQL one does

iancooper:
The MSSQL gateway accepted `OnMissingChannel` and never read it. The PostgreSQL gateway creates its queue table from the same setting, at the same points in the lifecycle. This closes that gap.

## The asymmetry this fixes

```text
PostgresChannelFactory.cs:27,48,69       → EnsureQueueStoreExists(..., MakeChannels)
PostgresMessageProducerFactory.cs:32,65  → EnsureQueueStoreExists(..., publication.MakeChannels)
PostgresMessagingGateway.cs:49,112       → if (makeTable == OnMissingChannel.Create) CREATE TABLE IF NOT EXISTS

MessagingGateway.MsSql — OnMissingChannel appeared in exactly three places before this PR,
all of them defaults: MsSqlSubscription.cs:81,130 and MsSqlMessageProducer.cs:77.
No EnsureQueueStoreExists. No CREATE TABLE.
```

So every MSSQL transport user ran the DDL themselves, which is why `MsSqlQueueBuilder` is public.

## What is here

`MsSqlMessagingGateway` mirrors `PostgresMessagingGateway` — a base class both factories derive from, with `EnsureQueueStoreExists` / `EnsureQueueStoreExistsAsync` honouring all three cases:

| `MakeChannels` | Behaviour |
|---|---|
| `Create` (the default) | creates the queue table and its topic index if absent; idempotent and safe under concurrent start |
| `Validate` | throws `ConfigurationException` at startup naming the missing table |
| `Assume` | does nothing and **opens no connection** |

`ChannelFactory` calls it as it opens a channel; `MsSqlMessageProducerFactory` as it builds a producer, because a sender may be the first thing to run against a new database.

The provisioning logic comes from `samples/TaskQueue/MsSqlMessagingGateway`, where it had grown to handle real races: separate batches so a swallowed create does not abandon the index statement, `2714`/`1913` tolerated as a lost race, `1205` retried once, a post-create re-probe so a name collision stops looking like a race, and an index guard on *what the index is* rather than what it is called. **In `samples/` none of it could be tested — no test project references `samples/`.** That is the point of moving it.

### 1. This changes behaviour, and `Create` is the default

An MSSQL user who provisions the queue table by hand will now see an idempotent existence probe at startup where nothing ran before. `Validate` and `Assume` are the ways out and both are tested. This is the deliberate content of the change, not a side effect — but it is the thing most likely to surprise someone on upgrade, so it should be in the release notes.

The concrete shape of that surprise, measured: an application login with DML rights and no DDL rights connects successfully and then gets SQL Server error 262, `CREATE TABLE permission denied`. That is now a `ConfigurationException` naming the table and the `MakeChannels` setting that avoids it, rather than a raw `SqlException` out of channel open. A release-note draft is in the [reply to review 19](https://github.com/BrighterCommand/Brighter/pull/4343#issuecomment-5645663151); `release_notes.md` is untouched here.

### 2. The sample's identifier guard was wrong as framework code, and the existing suite caught it

Ported unchanged, the guard demanded a plain unquoted identifier: `^[A-Za-z_][A-Za-z0-9_]{0,127}$`. The DDL is `CREATE TABLE [{0}]`, and **a bracketed identifier legally holds hyphens** — which is exactly how this repository's own tests name queue tables:

```text
ConfigurationException : The queue store table 'queue_test_09e7d459-9ab4-4edb-b588-05e72fef2411'
                         is not a plain SQL identifier of 128 characters or fewer.
```

Shipped, it would have rejected names this gateway has always accepted — a breaking change wearing the costume of a security fix. The guard now refuses only `]`, which closes the bracket and turns everything after it into free SQL, and names over SQL Server's 128. There is a test for the hyphenated case so it cannot come back.

## `SCHEMA_NAME()` rather than `SchemaName`, deliberately

`MsSqlMessageQueue` emits an unqualified `[{QueueStoreTable}]` at `:169`, `:215`, `:236` and `:246`, so the runtime resolves through the caller's default schema. Provisioning has to agree with it. Honouring `RelationalDatabaseConfiguration.SchemaName` would create the table where the gateway never looks. `SchemaName` still applies to the Outbox and the Inbox, which do qualify. Documented in the gateway README and in the class remarks.

## Tests — twenty-nine, test-first

`tests/Paramore.Brighter.MSSQL.Tests/MessagingGateway/Provisioning/`

- **Assume opens no connection** — proved by pointing the configuration at a port nothing listens on. Paired in the same class with a `Create` fact on the *same* configuration that must throw. Alone the silence would prove nothing: **three of these assertions passed against the unprovisioned gateway before a line of implementation existed**, which is exactly why each has a control.
- **Create** — table and index exist; running twice does not throw; the producer side provisions too; a name containing `]` is refused before any connection; a 129-character name is refused; a hyphenated GUID name is provisioned.
- **Validate** — missing table throws and names itself; existing table is silent (the control).
- **Concurrency** — eight factories released together on a fresh table; every start succeeds and exactly one table exists. This is the case five review rounds on #4331 asked for. Dedicated threads and a barrier, after each has warmed the connection pool, so that "together" is a property of the test rather than a hope about the thread pool.
- **The async ladder** — its own class, driving Create, the producer's `CreateAsync`, Validate and both name guards, because duplicated ladders drift and every other fact here drives the sync one.
- **Permission** — a login with `db_datareader` + `db_datawriter` and no DDL rights meets `Create` and gets a `ConfigurationException` naming the table and `Validate`; the same login validating a table an administrator made is silent (the control).
- **Provisioned once** — twenty publications naming one queue table touch the database once, and a new factory provisions again (the control).

Each new guard was also run **with its own fix disabled**, to check the test fails without it. Three assertions in the first round of this PR passed against a gateway that had not been written yet, which is why that is now a step rather than a hope.

## Verification

Run against SQL Server 2022 (`mcr.microsoft.com/azure-sql-edge`, the CI image), both target frameworks:

```text
baseline, clean master, same container : 247 passed,  0 failed
this branch at 6d03217af               : 261 passed,  0 failed
this branch after review 19            : 277 passed,  0 failed   (net9.0 and net10.0)
```

The baseline was measured rather than assumed — without it I would have attributed pre-existing failures to this change, or missed that all of them were mine. `dotnet build Brighter.slnx` is 0 errors.

## Relationship to ADR 0066

This is the V10 shape that [#4259](https://github.com/BrighterCommand/Brighter/pull/4259) expects to pick up. The ADR keeps provisioning "at producer creation and at channel open" for V10 and maps the legacy `OnMissingChannel` onto a strategy via `CreateDefaultProvisioner`; MsSql is listed there as *"Both (shared) — one set running idempotent DDL"*. After this PR, MSSQL and PostgreSQL are in the same place and migrate identically, instead of one migrating from the gateway and the other from somewhere else.

## Not done here

- `MsSqlMessageQueue.cs:34` logs the connection string at Debug, and `:169` interpolates the topic into SQL while `:215`/`:236` parameterize. Both pre-existing; now filed as #4346 and #4347.
- `tests/…/Configuration.cs` sets `s_databaseCreated = true` inside a `finally`, so a failed `CREATE DATABASE` still marks the database as created, and the guard is a process-local static while CI runs two TFMs against one container. That is the `sqlserver-ci` flake seen on #4331; now filed as #4349.
- `MsSqlQueueBuilder.GetExistsQuery` formats both names into quoted literals — public API, so a guard cannot live in its callers. Filed as #4345.
- `Validate` checks the table and not the topic index, so a queue table missing its index passes. Postgres behaves the same way; changing it would fail startups that work today, and it should change for both gateways or neither.


iancooper:
@claude Let's review, ignore the max turns block

iancooper:
Thank you — this is the most useful round this PR has had, and three of the eight sections found real defects. `e6103e1b7` answers all eight. I checked each finding rather than taking it, because a reviewer's measurement can be wrong too; one of these was, and it is §5's first bullet.

Everything below that says "measured" was run against `mcr.microsoft.com/azure-sql-edge` — the CI image — not reasoned about.

## 1. `CreateAsync()` never reaches the async ladder — correct, fixed

`Task.FromResult(Create())` at `:87`, exactly as you say, and `EnsureQueueStoreExistsAsync` was dead code on the producer side. `CreateAsync` now has a real body that mirrors `Create()` and awaits, as `PostgresMessageProducerFactory.CreateAsync` does. The `Topic is null` check still runs first.

You were also right about which path this is: `MsSqlProducerRegistryFactory(...).CreateAsync()` is what `When_requeueing_a_message_aync.cs:44` and `When_a_message_is_sent_async.cs:36` take, so this was the route an application actually uses.

**One limit, stated rather than papered over.** The new tests prove the async producer path *provisions*; they do not prove it *awaits*. Every instrument I could think of for the latter — checking the returned task had not already completed, counting continuations posted to a synchronization context — is timing-dependent, and a flaky test asserting a true thing is worse than no test. That the body awaits is checked by reading it.

## 2. The 128-character bound is nine too generous — correct, and the arithmetic is exact

Measured, four table names, table create then index create, the gateway's own two DDL shapes:

```text
table name 119 chars -> index 128 : CREATE TABLE ok, CREATE INDEX ok
table name 120 chars -> index 129 : CREATE TABLE ok, CREATE INDEX SqlException 103
table name 125 chars -> index 134 : CREATE TABLE ok, CREATE INDEX SqlException 103
table name 128 chars -> index 137 : CREATE TABLE ok, CREATE INDEX SqlException 103
```

`128 - len("IX_") - len("_Topic")` = 119 is right to the character, and 103 is the number. Worse than the raw exception: the table **is** created before the index fails, so the failure leaves a queue table that can never gain its index and fails identically on every subsequent start.

**One thing I did differently from the fix you proposed.** The tighter bound is applied to `Create` only, not to `ValidatedQueueTableName` as a whole. `Validate` runs one parameterised probe and builds no identifier at all, so a 125-character table that already exists is a table this gateway can legitimately be pointed at — tightening the guard for both would have failed a working configuration to prevent a create that was never going to happen. There is a test for that asymmetry: `When_validating_a_queue_too_long_to_have_been_created_here_should_not_throw`.

The message names the derived index, its length, and the 119 ceiling, and says a longer existing table can still be used with `Validate`. Tests at 120 (throws) and 119 (creates, with its index) on both ladders. And because the `IX_..._Topic` convention lives in `MsSqlQueueBuilder` and is only exposed as finished DDL, there is now a test pinning it — if the convention moves, that fails rather than the bound silently ceasing to match.

## 3. `Create` wraps the connect but not the create — correct, and the failure mode is the one you predicted

Measured against a login in `db_datareader` + `db_datawriter` and nothing else:

```text
connect                     : ok      <- so the ConfigurationException wrapper never fires
CREATE TABLE                : SqlException 262, "CREATE TABLE permission denied in database 'BrighterTests'."
262 in {2714, 1913, 1205}?  : False   <- so it propagated raw
Validate probe (SELECT over sys.tables) with the same login : ok
```

That last line is why the message can name a way out rather than just apologise: the login that cannot create the table *can* check that it is there. An unexpected `SqlException` out of the DDL is now a `ConfigurationException` carrying the provider's own sentence, the table name, and `Validate`/`Assume`.

A second 1205 on the retry still propagates raw and deliberately: a deadlock is not a configuration problem, and wrapping it in `ConfigurationException` would mislabel it.

Two tests, with a control: the denied create names `Validate`, and the same login validating a table an administrator created is silent. The test creates and drops its own login, so it needs no fixture.

## 4. Provisioning runs once per publication — correct, fixed one level down

Confirmed your grep: the read of `Configuration.QueueStoreTable` in `MsSqlMessagingGateway` is the only source in the assembly, and MsSql has no per-publication or per-subscription override.

Rather than hoisting above the loop, the gateway instance now remembers which `OnMissingChannel` values it has already satisfied. That covers both call sites with one mechanism — your "less sharply" case for `ChannelFactory` too, so a dispatcher's N performers collapse to one probe — and it keeps per-publication `MakeChannels` honest, which hoisting would have flattened. Races only duplicate idempotent work, which the concurrency test already covers.

Two tests, and the second is the one that matters: twenty publications touch the database once, and a **new** factory provisions again. Without that control the first would also pass on a gateway that had simply stopped provisioning.

This also made `When_the_queue_is_created_twice_should_not_throw` vacuous — one factory called twice now returns early on the second call — so it uses two factories, which is what "safe to run on every start" meant anyway.

## 5. Smaller things — three taken, one declined on a measurement, one declined on a reason

**`await using` — declined, and this is the finding that was wrong.** This package targets `net462`:

```text
BrighterFrameworkAndCoreTargetFrameworks = net462;net8.0;net9.0;net10.0
```

and on `net462` `SqlConnection` does not implement `IAsyncDisposable`. Changing it is a compile error, measured:

```text
error CS8417: 'SqlConnection': type used in an asynchronous using statement must implement
'System.IAsyncDisposable' ... [TargetFramework=net462]
```

The parity argument does not carry either: `Paramore.Brighter.MessagingGateway.Postgres` is `BrighterCoreTargetFrameworks` — `net8.0;net9.0;net10.0`, no `net462`. There is now a comment at the `using` saying so, because this is exactly the kind of thing that gets "fixed" by the next reader.

**Null check — taken.** `Configuration` is `configuration ?? throw new ArgumentNullException(...)`.

**The dropped comments — taken.** Both explanatory comments are back on the async ladder. You are right that this is how duplicate ladders start to drift; it had drifted within one PR.

**`Validate` not checking the index — noted, declined.** Making `Validate` demand the index would fail startups that work today, on a table someone else provisioned, for a performance property rather than a correctness one. Postgres behaves the same way. If it should change it should change for both, in its own PR.

## 6. `MsSqlMessageConsumerFactory.Configuration` is new public API — correct, now `internal`

`internal` satisfies `ChannelFactory` in the same assembly and keeps a second public route to the connection string out of a shipped package. No test touched it. The XML doc now says *why* it is internal rather than merely stating the intra-assembly reason and then being public anyway.

The `ConfigurationOf` helper is in too — the parameter keeps the original name so the `ArgumentNullException` message is unchanged.

## 7. Test observations — all four taken

- **The two "before touching the database" tests now point at `UnreachableConnectionString`**, which is the instrument the Assume class already owned, as you say. It is `internal` now so both classes share it. The bracket test and the over-long test would previously have passed with the guard running *after* the connect; they no longer can.
- **The two Assume controls assert `Assert.IsType<ConfigurationException>`** and that the message names the table. You were right that `Assert.NotNull` passed on precisely the raw `SqlException` that `Connect` exists to prevent.
- **The concurrency test has a barrier.** Two changes, not one: dedicated threads rather than `Task.Run`, because the pool injects threads at roughly one a second and eight queued items need not be running at once; and each thread warms the connection pool with a throwaway connection *before* parking at the gate, so what follows the release is the DDL rather than eight staggered handshakes. Your reading was right — as written it could go green without ever reaching 2714 or 1205.
- **The async ladder has its own class**, eight facts: Create, the producer's `CreateAsync`, Validate-throws with its control, both name guards, and an Assume/Create pair. I did not go with `[Theory]` over a sync/async selector — the two ladders are separate methods with separate signatures, and a selector would have obscured which one a failure came from.
- **Channels are disposed** at the sites you listed and in the new tests.

## 8. Docs and samples

The gateway README now carries the two things a user meets in practice: the permission case with the sentence Brighter raises and the setting that avoids it, and the 119-character bound with the reason and the note that `Validate` keeps the full 128.

**The release note is the one thing I have not written**, because `release_notes.md` is outside what I have been asked to touch on this PR. Draft, for `## Master`, to add or amend as you see fit:

> **MSSQL transport provisions its queue table.** `OnMissingChannel` on an `MsSqlSubscription` or `Publication` is now honoured by the MSSQL messaging gateway, as it already was by PostgreSQL: `Create` — the default — creates the queue table and its topic index as a channel is opened or a producer is built. **This is a behaviour change on upgrade**: an application whose queue table is provisioned by a migration or a DBA will now run an idempotent existence probe at startup, and one whose login has no DDL rights will fail with a `ConfigurationException` naming the table. Set `MakeChannels` to `Validate` to check for the table instead of creating it, or to `Assume` to skip the check entirely.

## Verification

```text
baseline, this branch at 6d03217af, same container : 261 passed, 0 failed
after these changes                                : 277 passed, 0 failed  (net9.0 and net10.0)
dotnet build Brighter.slnx                         : 0 errors
```

**Each new guard was run with its own fix disabled**, so that none of the sixteen new facts is a test that passes either way — the lesson from this PR's first round, where three assertions passed against a gateway that had not been written yet:

```text
index-name bound disabled   -> When_the_queue_table_name_leaves_no_room_for_the_index_name : FAILS
DDL exception wrap disabled -> When_the_login_cannot_create_the_queue                       : FAILS
provisioning memory disabled-> When_one_gateway_provisions_repeatedly                       : FAILS
```

The 261 baseline was re-measured on this branch rather than inherited from the PR description, which said 260.

You noted you could not build or run here. Findings 1, 2 and 4 were static and held; 3 needed a database and held; 5's first bullet needed a `net462` compile, which is where it came apart. Worth knowing which of your findings that environment limit actually touched.

iancooper:
@claude Let's review again — second pass, on `e6103e1b7` rather than the whole PR.

Everything in this diff was written answering your last review and none of it has been read by anyone else, which is the case for looking again. CI is green at this head: 26 pass, 0 fail, including `sqlserver-ci`. Two jobs went red first — `memory` (`OutboxCompactionDispatchedOnlyTests`, a two-second poll on a background compaction) and `rabbitmq-sync-ci` — and both cleared on a re-run of the same commit; neither is reachable from a diff of `MessagingGateway.MsSql` plus its tests.

**Where I would look hardest, in order:**

1. **The per-instance provisioning memory** (`_ensured`, `MsSqlMessagingGateway`). This is the change nobody specified — your §4 asked for hoisting above the loop and I did something broader, keyed on `OnMissingChannel` so per-publication `MakeChannels` stays honest and `ChannelFactory` benefits too. Is keying on the enum right? A `Validate` that succeeds and a later `Create` that is therefore skipped would leave the index uncreated — I think that ordering cannot arise from either factory, but that is exactly the sort of thing I would rather have contradicted than believed.
2. **The `Create`-only length bound.** I did not apply your fix as proposed: `Validate` keeps the full 128 because it builds no identifier. That is a deliberate asymmetry inside one method and it is the kind that rots.
3. **The new `catch (SqlException)` arms in `Execute`/`ExecuteAsync`.** Four catch clauses now, order-dependent, with a `when (retry.Number != DeadlockVictim)` that exists so a second deadlock still propagates raw rather than being mislabelled as configuration. Please check the ordering does what the comments claim.
4. **`MsSqlMessageProducerFactory.CreateAsync`** is now a copy of `Create` with awaits. Two bodies that must not drift, and I have no test that would catch it if they did.
5. **The new tests themselves** — sixteen of them, and a test that passes either way is worse than no test. I ran each new guard with its fix disabled to check it fails, but that only covers the three guards, not the rest.

**One correction to your last round, and it is the only finding that did not hold.** `await using` on the async connection does not compile here: this package targets `net462`, where `SqlConnection` has no `IAsyncDisposable` — `CS8417`, measured. The Postgres parity does not carry either; that package is `net8.0` and up. There is a comment at the `using` now.

**And one claim from my reply, now measured rather than hoped.** I said the barrier makes the race test collide; instrumented locally, one run of the eight-way race hit `already-exists 2714` once — so one start genuinely lost the create and took the swallow path, and the other seven found the table through `IF NOT EXISTS`. A collision where there may have been none, not a guarantee of one per run.

Full reply to review 19: https://github.com/BrighterCommand/Brighter/pull/4343#issuecomment-5645663151

iancooper:
**Correction to my own trigger, before you finish checking it.** On §1 I wrote *"I think that ordering cannot arise from either factory"*. That is wrong, and falsifying it took two minutes — so please read §1 against `6ad162af6` rather than `e6103e1b7`.

**The ordering is reachable.** One `Dispatcher` shares one `ChannelFactory` across every subscription, and subscriptions may differ in `MakeChannels`. So a `Validate` subscription can open first, against a table provisioned elsewhere and missing its topic index, and a `Create` subscription can follow on the same gateway instance.

**Keying the memory on `OnMissingChannel` is exactly what saves it**, which I had not understood as load-bearing when I wrote it. A `Validate` that succeeds records only `Validate`; a later `Create` still finds no `Create` key and runs the full path, index included.

There is now a test for it, with the control that makes it mean something:

```text
keyed on OnMissingChannel (shipped) : Validate, then Create -> index created, PASSES
collapsed to a single flag          : Validate, then Create -> index missing,  FAILS
```

`When_one_gateway_validates_and_then_creates_should_still_create_the_index`. 278 pass / 0 fail on both target frameworks.

**So the question I asked you in §1 is answered, but the interesting part stands**: the memory is still the change nobody specified, and this is one ordering out of however many exist. If you can find a second one I have not thought of, that is worth more than the test I just wrote.

iancooper:
Both of your defects are real, both were introduced by the commit that answered your last round, and both are confirmed by exactly the experiment you described. `d85966604`.

## 5a and 5b — confirmed by deleting the guard, as you asked

```text
bracket guard deleted   -> both bracket tests            : PASSED   <- the defect
128 bound deleted       -> whole suite, 280 tests        : PASSED   <- the defect
```

**You identified the mechanism precisely**: I re-pointed those tests at `UnreachableConnectionString` to make the ordering claim in their names true, and traded away the only thing that distinguished a refused name from an unreachable server, because `Connect` wraps its failure in a `ConfigurationException` too. The 120-character pair survived for the reason you give — `Assert.Contains("119", …)` is an assertion on the sentence.

Fixed as you proposed. The bracket tests assert `"close the bracket"`; the 129-character test drives **`Validate`**, where the 119 bound cannot fire, and asserts the message names 128 and *not* 119. Both now fail with their guard deleted:

```text
bracket guard deleted   -> both bracket tests            : FAILS
128 bound deleted       -> whole suite                   : FAILS (1 of 280)
```

**The transferable part is about where I applied the discipline.** I ran the fix-disabled control on the three guards I added and on none of the guards I only changed the tests for. The defects are exactly in the gap. A changed test needs the control as much as a new one does.

## 5c, 5d — taken

The concurrency test's warm-up was outside `Record.Exception`, and you are right about both consequences: on a foreground thread an unhandled exception takes the host down rather than failing the test, and a thread dying before `SignalAndWait` leaves the rest parked with `Join()` never returning. Whole body guarded now, with a timeout on the gate and on the join, so a hang becomes a failure with a sentence.

The view test is in and it is as cheap as you said — `CREATE VIEW [name] AS SELECT 1 AS x`, then `Create`. It fails with the re-probe deleted, which makes it the test for the one place a swallowed 2714 could become a silent success.

## §3 — the transient point is the best thing in this review

You are right that `CouldNotProvision` was arguing from 262 to *every* unexpected `SqlException`. Being told to change `MakeChannels` because the database was throttled for four seconds sends someone to fix the wrong thing, and — your sharpest line — a cold Azure SQL instance at startup is likelier than a permission denial on a connection that has worked for months.

A transient set now propagates raw, for the same reason the second deadlock does: `-2`, `233`, `4060`, `4221`, `10928`, `10929`, `40197`, `40501`, `40613`, `49918`, `49919`, `49920`. That is the first of your two options; I preferred it to softening the sentence because a timeout should not be a `ConfigurationException` at all, whatever it says.

The inner filter now spells out `!= alreadyExists` rather than leaving clause order to carry it, and the polarity is `!= OnMissingChannel.Validate` so the bound and the branch ask the same question. Your point about a fourth member is not hypothetical for long — ADR 0066 is heading at strategies.

## §1, §4 and the smaller things

Your §1 analysis is better than the question I asked. I had already corrected my own trigger — the ordering *is* reachable, one `Dispatcher` shares one `ChannelFactory` — and you went further by naming the converse as harmless and the `QueueStoreTable` protected setter as the actual boundary. That clause is now in the comment, in the terms you put it: the memory assumes a fixed table name, and neither that nor the enum key survives being "simplified".

§4: `RequireTopic` / `AddProducer`, so the two bodies differ by the call that must differ. Async channels disposed, `ClearPool` rather than `ClearAllPools`, the login can no longer leak when the constructor throws, and the blank-name guard has its one-line test.

**One thing running it caught that reading would not have**: `DROP LOGIN IF EXISTS` is not valid syntax on this server, so my first cut of the login cleanup failed two tests. Guarded on `sys.server_principals` instead.

## Verification

```text
278 -> 280 pass, 0 fail, net9.0 and net10.0
dotnet build Brighter.slnx : 0 errors

fix-disabled controls, all five:
  index-name bound      -> FAILS      DDL exception wrap  -> FAILS
  provisioning memory   -> FAILS      bracket guard       -> FAILS
  128 bound             -> FAILS      post-create re-probe-> FAILS
```

iancooper:
@claude Final pass, please — on `4729204e6`, and on the whole PR this time rather than one commit.

Two rounds have each found real defects, and **review 20's two were both introduced by the commit answering review 19**, so the thing I most want from a last look is whether *this* fix round did the same again. `d85966604` changed six things in response to you and added four tests; `4729204e6` adds the release note.

**Where I would look, in order:**

1. **The transient-error set** (`IsTransient`, `MsSqlMessagingGateway`). This is your §3 finding taken as its first option: `-2`, `233`, `4060`, `4221`, `10928`, `10929`, `40197`, `40501`, `40613`, `49918`, `49919`, `49920` now propagate raw rather than becoming a `ConfigurationException`. Two questions I cannot answer by reading: is anything in that list *not* transient — `4060` in particular, which can mean a genuinely wrong database name as well as a failover — and is anything obviously missing? A wrong `4060` would now propagate as a bare `SqlException` where the connect wrapper used to catch its sibling.
2. **The two tests I repaired for 5a and 5b.** They assert on message substrings now — `"close the bracket"`, `"128"`, `"119"` — which pins behaviour to wording. I think that is the right trade here because the type genuinely cannot discriminate, but it is a new coupling and you should say if it is the wrong one.
3. **`When_the_queue_table_name_is_longer_than_sql_servers_limit_should_throw` now drives `Validate`.** That is your fix and it works, but it means no test drives a 129-character name through `Create` any more. The 120-character test covers that bound; I want to know if you think the >128 *and* `Create` combination needs its own.
4. **The new view-collision test** (`When_something_else_owns_the_table_name_…`). It creates a view, expects 2714 swallowed and the re-probe to throw. It passes and fails correctly with the re-probe deleted. Is the `CREATE VIEW`/`DROP VIEW` cleanup safe if the assertion throws first?
5. **The release note** (`release_notes.md`, `## Master`). New in `4729204e6` and unreviewed. It claims a behaviour change, names `Validate`/`Assume`, and states the 119 bound. Please check it against the code rather than against my description of the code.

**Still open and deliberately not done:** `Validate` does not check the topic index, so a queue table missing its index passes validation. Postgres behaves the same way and I did not change either — say if you think that should block.

Verification at this head: **280 pass / 0 fail** on net9.0 and net10.0, `dotnet build Brighter.slnx` 0 errors, CI 26 pass / 0 fail. Four CI reds across this session — `memory`, `rabbitmq-sync-ci`, `postgres-ci`, `kafka-ci` — all cleared on a re-run of their own commit, and none is reachable from a diff of `MessagingGateway.MsSql` plus its tests.

Reply to your last round: https://github.com/BrighterCommand/Brighter/pull/4343#issuecomment-5645871492

iancooper:
All of it taken, in `fc7dd160a`, and your §1a was right in the way that matters most — it found a defect in the guard I added *in the same commit as the discipline that would have caught it*.

## §1a — confirmed, exactly as you described

```text
IsTransient body replaced with `=> false` : 280 pass, 0 fail   <- the defect
after the new test                        : 282 pass, 1 FAIL
```

You were also right about why it escaped: my control list covered the guards I added and this was a guard I added, in the same commit, and I did not think of it as one. That is the third round running in which the defect has been *at the edge of my own fix* rather than in the mechanism, and the pattern is now specific enough to be useful: **the thing I just changed is the thing I stop testing.**

## §1b — measured, and you are right that the set was in the wrong place

```text
wrong database name -> Open() : SqlException 4060  "Cannot open database ... requested by the login"
                                so unreachable from ExecuteNonQuery: True
```

So 4060 was inert where I put it, exactly as you said, and a wrong database name behaves as it did before `d85966604`. `Connect`/`ConnectAsync` now consult `IsTransient` too.

**One thing I did differently: 4060 is out of the set entirely.** Moving the rule into `Connect` makes it live, and there it would turn my original worry real — a misspelled database name propagating raw instead of getting the message that names the table. `40615` (firewall) and `11001` (host not resolved) are absent for the same reason. And a genuinely unreachable server turns out to arrive as **number 0**, so it stays wrapped and the tests that use `UnreachableConnectionString` keep their meaning — that was the thing I checked before touching `Connect`.

## §1c — taken, and the transport family was the real omission

`64`, `10053`, `10054`, `10060` in, plus `1222`, `1204`, `40143`, `49977`. You are right that it is the execute-time equivalent of everything I had included, and that a hand-rolled list drifts; the comment now says which codes are deliberately *absent* and why, since that is the part that cannot be recovered from the list itself.

`1222` is measured too — `SET LOCK_TIMEOUT 1000` against the same held lock gives it.

## §3 — and your correction to yourself was right

I checked the ordering claim independently: the absolute 128 guard does run first, for every mode. The new test pins it, and your reasoning for why it needs pinning is the good part — with the guards swapped, a 129-character name would be told it "can still be used with `MakeChannels = Validate`", which is advice no one can take. It fails when the guards are swapped.

## §2, §4, §5, §6, §7 — taken

The `DoesNotContain("provider said")` refinement is in on both bracket facts. `§6` was one statement off exactly as you said — the `try` now opens around the `Execute` that creates the login. `§7`'s join results are collected before anything is asserted.

The release note's transient sentence is now true rather than trimmed, because `Connect` changed: it says "on both the connect and the DDL", and adds that a wrong server, a malformed string and a misspelled database name are deliberately *not* in the set.

## The one thing three review rounds read and running caught

`DropLogin` collected session ids and then `EXEC`'d a `KILL` batch. A pooled session closing between the two makes `KILL 67` fail with *"Process ID 67 is not an active process ID"*, which throws out of `Dispose` **after the assertions have passed**.

```text
before: 3 of 4 runs red, both TFMs against one container  <- which is what sqlserver-ci does
after : 6 of 6 clean
```

It is in the file you reviewed three times, and none of us saw it, because it is not visible in the text — it is visible in the gap between two statements under load. Worth stating plainly since your last three rounds have each said "not verified locally": this is the class of defect that reading cannot reach, and the one where your `--allowedTools` limit actually costs something.

## Verification

```text
280 -> 283 pass, 0 fail, net9.0 and net10.0
dotnet build Brighter.slnx : 0 errors

fix-disabled controls: index bound, DDL wrap, provisioning memory, bracket guard,
128 bound, post-create re-probe, IsTransient, guard ordering -- all FAIL when removed
```

Agreed on `Validate` not checking the index, and for your reason rather than mine: ADR 0066's strategies are where a thorough-validate mode can be opted into rather than imposed. Not blocking, and the note's "the table is checked" stays as written.

PR #4373, opened 2026-09-17
fix: report response type for missing reply subscriptions

### LINKED ISSUE #4337: CommandProcessor.Call: "No Subscription registered fpr replies of type" — typo, and it names the wrong type
`src/Paramore.Brighter/CommandProcessor.cs:1429`:

```csharp
var subscription = _replySubscriptions?.FirstOrDefault(s => s.RequestType == typeof(TResponse));

if (subscription is null)
    throw new InvalidOperationException($"No Subscription registered fpr replies of type {typeof(T)}");
```

Two things:

1. **`fpr`** should be `for`.
2. **It prints `typeof(T)`, but the match is against `typeof(TResponse)`.** So the message names the request type while the lookup that failed used the response type — a reader is told to register a subscription for the wrong thing.

This is the first guard on the `Call` path, six lines before the `_responseChannelFactory` check, so it is the message anyone with no RPC wiring actually sees.

Suggested:

```csharp
throw new InvalidOperationException($"No Subscription registered for replies of type {typeof(TResponse)}");
```

Found while documenting `Call`'s failure modes for the Brighter guide — the documentation originally named the *other* exception, because the second guard is the one that mentions what you skipped. The maintainer asked for this to be filed.

gabisonia:
## Description

When `CommandProcessor.Call<T, TResponse>` cannot find a reply subscription, its exception names `T` even though the lookup uses `TResponse`. Report the response type and correct "fpr" to "for" so callers are told which subscription to register.

Add a regression test that checks the exception type and complete diagnostic when no reply subscription is configured. Update the core guide's quoted error and explanation.

## Related Issues

Fixes #4337.

## Type of Change

- [x] Documentation update

## Checklist

- [x] I have read the Contributing Guide
- [x] I have checked the documentation for relevant guidance (core guide's request-reply section)
- [x] I have added/updated XML documentation for any public API changes (no API signatures changed)

## Additional Notes

The new regression test failed against the original diagnostic before the fix.

Validation: `dotnet test tests/Paramore.Brighter.Core.Tests/Paramore.Brighter.Core.Tests.csproj --framework net10.0 --verbosity quiet` — 979 passed, 7 skipped, 0 failed. .NET 9 was not run locally because that runtime is not installed. `git diff --check` passed.


iancooper (approved):

iancooper:
Thanks, @gabisonia. Will merge. Brighter is due a release very soon

PR #4374, opened 2026-09-17
fix: include message ID when rejecting exhausted requeues

### LINKED ISSUE #4363: Message pump dead-letters with "Handle Count Exceeded for message " (empty id) on transports that never set x-original-message-id
### Describe the bug
When the requeue budget is exhausted, `Proactor.RequeueMessage` (and the identical block in `Reactor`) builds the rejection description from the `x-original-message-id` bag entry:

```csharp
var originalMessageId = message.Header.Bag.TryGetValue(Message.OriginalMessageIdHeaderName, out object? value) ? value.ToString() : null;
...
return RejectMessage(message, new MessageRejectionReason(
    RejectionReason.DeliveryError,
    $"Handle Count Exceeded for message {originalMessageId}"));
```

The log line next to it guards against a missing value (`string.IsNullOrEmpty(originalMessageId) ? string.Empty : $" (original message id {originalMessageId})"`), the description does not. Only three gateways ever write that bag entry: RMQ (`RmqMessagePublisher`), MsSql and Postgres (their `Requeue`). Azure Service Bus (`AzureServiceBusConsumer.RequeueAsync` resends the message as is), SQS/SNS, Kafka and the rest do not, so on those transports every "handle count exceeded" dead-letter carries the description `Handle Count Exceeded for message ` with nothing after it. The reason column is fine (`DeliveryError`, since #4196).

Suggested fix, transport-agnostic, in the pump:

```csharp
var originalMessageId = message.Header.Bag.TryGetValue(Message.OriginalMessageIdHeaderName, out object? value)
    ? value?.ToString()
    : null;
originalMessageId = string.IsNullOrEmpty(originalMessageId) ? message.Id.Value : originalMessageId;
```

Optionally, `RequeueMessage` could also stamp `x-original-message-id` itself before `Channel.RequeueAsync` when it is absent (what MsSql and Postgres do locally today), so the log line and any gateway that round-trips the bag get the original id too.

### To Reproduce
No easy code available. I tried local Azure Service Bus emulator, queue subscription with `requeueCount: 3`, `requeueDelay: 5 s`, handler throwing `DeferMessageAction` on every attempt. After the third cycle the message is in the dead-letter sub-queue with:

```
DeadLetterReason 'DeliveryError', description 'Handle Count Exceeded for message '
HandledCount = 2   (application property, correct)
```

Pump log for the same message: `MessagePump: Have tried 3 times to handle this message <id> from ... , dropping message.`  (no "original message id" suffix, as in `'this message {Id}{OriginalMessageId} from'` log template).

### Exceptions (if any)
None.

### Further technical details
- Brighter version: 10.7.0
- `dotnet --info`: SDK 10.0.400, host 10.0.11, RID win-x64
- OS: Windows 11 (10.0.26200)
- Related: #4196 (reason/description now reach the native fields, which is what makes this visible)

---
Analysis was prepared with help of an AI assistant.

gabisonia:
## Description

When a message exhausts its requeue limit without an `x-original-message-id` header, the async pump produces `Handle Count Exceeded for message ` with no ID. A null header value also throws in both pumps before rejection completes.

Use the original ID when it is nonempty, otherwise fall back to `message.Id.Value`. Apply the same diagnostic to the sync pump, which previously used a generic rejection description without an ID. Preserve the existing delivery-error reason and optional original-ID log suffix.

Add four regression cases per pump: absent, null, empty, and populated original-ID headers. Tests exercise the real command processor with an in-memory dead-letter queue and require no sleeps or external services.

## Related Issues

Fixes #4363.

## Type of Change

- [ ] Documentation update

## Checklist

- [x] I have read the Contributing Guide
- [x] I have checked the documentation for relevant guidance
- [x] I have added/updated XML documentation for any public API changes (no API signatures changed)

## Additional Notes

Before the fix, seven regression cases failed for the missing-ID diagnostic or null dereference; the async case preserving a populated original ID already passed.

Validation: `dotnet test tests/Paramore.Brighter.Core.Tests/Paramore.Brighter.Core.Tests.csproj --framework net10.0 --verbosity quiet` — 986 passed, 7 skipped, 0 failed. `git diff --check` passed. .NET 9 was not run locally because that runtime is not installed.


iancooper (approved):

iancooper:
Thanks @gabisonia

PR #4375, opened 2026-09-18
fix: exclude credentials from MSSQL queue logs

### LINKED ISSUE #4346: MsSqlMessageQueue logs the full connection string, including the password, at Debug
`src/Paramore.Brighter.MessagingGateway.MsSql/SqlQueues/MsSqlMessageQueue.cs:34`:

```csharp
Log.MsSqlMessageQueueCtor(s_logger, _configuration.ConnectionString, _configuration.QueueStoreTable);
```

and the message template at `:254`:

```csharp
[LoggerMessage(LogLevel.Debug, "MsSqlMessageQueue({ConnectionString}, {QueueStoreTable})")]
```

A SQL Server connection string routinely carries `Password=...`. At `Debug` this reaches every configured sink — console, file, and whatever aggregator the application ships to.

### Why it matters more than it used to

The MSSQL messaging gateway's own documentation now tells a reader to put a password in that string, and the sample under `samples/TaskQueue/MsSqlMessagingGateway` demonstrates it. That sample keeps the credential off stdout **only** because `GreetingsSender` happens to set `MinimumLevel.Override` for this category — a mitigation that works for one sample and for nobody else. Anyone running Brighter at `Debug` gets the password in their logs.

This was raised independently by three separate review rounds on #4331, which is usually a sign the fix belongs in the library rather than in another README paragraph.

### Suggested fix

Log the server and database without the credential — `SqlConnectionStringBuilder` gives you `DataSource` and `InitialCatalog` directly:

```csharp
var builder = new SqlConnectionStringBuilder(_configuration.ConnectionString);
Log.MsSqlMessageQueueCtor(s_logger, builder.DataSource, builder.InitialCatalog, _configuration.QueueStoreTable);
```

That keeps everything the log line is actually useful for — which server, which database, which table — and drops the part nobody wanted in a log. The sample's `MinimumLevel.Override` and the README's warning about it could then both be deleted rather than maintained.

gabisonia:
## Description

When Debug logging is enabled, `MsSqlMessageQueue` currently sends the complete connection string, including credentials, to every configured logging sink. Parse it with `SqlConnectionStringBuilder` and log only the server, database, and queue table. Parsing is skipped when Debug logging is disabled.

Remove the sample's logging override and obsolete README warning. Add regression coverage for standard credentials, connection-string aliases, passwords containing semicolons, and integrated authentication. Assertions check both rendered messages and structured log values.

## Related Issues

Fixes #4346

## Type of Change

- [x] Documentation update

## Checklist

- [x] I have checked the documentation for relevant guidance (MSSQL sample README)
- [x] I have added/updated XML documentation for any public API changes (no public API changes)

## Additional Notes

Validation with .NET SDK 10.0.400:

- Whole solution: `dotnet build Brighter.slnx --no-restore` passed with zero errors.
- Full MSSQL test project on `net10.0`: 286 passed, zero failed or skipped, using the repository's SQL Edge Docker image.
- Confirmed the regression test fails against the original code because the password appears in the rendered log, then passes with the fix.


iancooper:
Thanks, @gabisonia. You are really helping with all these bugs. I appreciate that.

iancooper (approved):
Great contribution

PR #4382, opened 2026-09-20
fix: escape MSSQL queue builder names

### LINKED ISSUE #4345: MsSqlQueueBuilder.GetExistsQuery formats both names into SQL string literals
`src/Paramore.Brighter.MessagingGateway.MsSql/MsSqlQueueBuilder.cs:84`:

```csharp
public static string GetExistsQuery(string queueTableName, string schemaName = "dbo") =>
    string.Format(QUEUE_EXISTS_SQL, queueTableName, schemaName);
```

and the template it formats into:

```sql
WHERE t.name = '{0}' AND s.name = '{1}'
```

Both names land inside SQL string literals with no escaping, so a name containing a single quote changes the statement. `GetDDL` and `GetIndexDDL` have the same shape into `CREATE TABLE [{0}]`, where the character that matters is `]` rather than `'`.

### Why this is worth fixing rather than guarding at each call site

`MsSqlQueueBuilder` is **public API** — deliberately so, because the MSSQL gateway had no provisioning path and callers were expected to run the DDL themselves. So the guard cannot live in a caller without every caller re-deriving it, and each one getting it slightly differently.

A worked example of getting it wrong: while moving the sample's provisioner into the gateway (#4343) I ported a guard of `^[A-Za-z_][A-Za-z0-9_]{0,127}$`, which looks careful and is wrong — a **bracketed** identifier legally holds hyphens, and this repository's own test suite names queue tables `queue_test_<guid>`. The existing suite caught it. That is the argument for one guard, in the builder, next to the `string.Format`.

### Suggested fix

- `GetExistsQuery`: bind the two names as parameters rather than formatting them, or at minimum escape `'` as `''`.
- `GetDDL` / `GetIndexDDL`: escape `]` as `]]`, which is SQL Server's own convention for a bracketed identifier, and reject a name over 128 characters.

A parameterised existence check also fixes a second, quieter problem: `GetExistsQuery` defaults `schemaName` to a literal `dbo`, while every statement the gateway issues against the queue is **unqualified** and so resolves through the caller's default schema. A login whose default schema is not `dbo` gets an existence check that disagrees with the runtime. #4343 sidesteps this by asking `SCHEMA_NAME()`, but `GetExistsQuery` is public and still carries the `dbo` default.

Raised in review of #4331; #4343 is the first production caller of this family of methods.

gabisonia:
## Description

`MsSqlQueueBuilder` inserts table and schema names directly into SQL, so apostrophes break existence checks and closing brackets break table/index DDL. Escape SQL literals and bracketed identifiers, preserve Unicode names, and reject blank or oversized identifiers before returning SQL. Index validation includes the `IX_` prefix and `_Topic` suffix.

Keep the existing string-returning API and `dbo` default. Callers can explicitly pass `schemaName: null` to use `SCHEMA_NAME()`, matching the default schema used by the unqualified DDL. The gateway's existing rejection of closing brackets remains because its runtime queue statements still interpolate names without escaping them.

Add 22 regression cases covering literal and SQL-like names, Unicode, identifier boundaries, invalid arguments, and DDL/existence checks under a user with a non-`dbo` default schema. Update XML documentation and the gateway README.

## Related Issues

Fixes #4345

## Type of Change

- [x] Documentation update

## Checklist

- [x] I have checked the [MSSQL gateway documentation](../blob/master/src/Paramore.Brighter.MessagingGateway.MsSql/README.md) for relevant guidance

## Validation

- Confirmed the initial regression tests failed against the original implementation.
- Full `Paramore.Brighter.MSSQL.Tests` suite on .NET 10 against the repository's Azure SQL Edge container image: **331 passed, 4 skipped, 0 failed**.
- MSSQL gateway build for `net462`, `net8.0`, `net9.0`, and `net10.0`: **0 warnings, 0 errors**. Runtime tests were run on .NET 10 only.
- `git diff --check` passed.


iancooper (approved):

iancooper:
Thanks @gabisonia

PR #4383, opened 2026-09-20
fix: coordinate MSSQL test database setup across processes

### LINKED ISSUE #4349: MSSQL tests: EnsureDatabaseExists marks the database created even when CREATE DATABASE threw, and its guard cannot span the two TFM processes
`tests/Paramore.Brighter.MSSQL.Tests/Configuration.cs`, `EnsureDatabaseExists`:

```csharp
    using var command = connection.CreateCommand();
    command.CommandText = $"IF DB_ID('{databaseName}') IS NULL BEGIN CREATE DATABASE {databaseName}; END;";
    command.ExecuteNonQuery();            // :45
}
finally
{
    s_databaseCreated = true;             // set even when ExecuteNonQuery threw
    s_semaphoreSlim.Release();
}
```

Two problems, and together they produce a flake that has already cost real time.

**1. The flag is set in `finally`.** If `ExecuteNonQuery` throws, `s_databaseCreated` is still set to `true`, so every subsequent caller skips creation and proceeds against a database that may not exist. The first test to fail reports a connection or missing-object error; the ones after it report something else entirely.

**2. The guard is a process-local `static`, and CI runs two processes.** `sqlserver-ci` runs the suite for `net9.0` and `net10.0` against **one** SQL Server container. A `SemaphoreSlim` and a `static bool` coordinate threads inside one process; they cannot coordinate two. Both processes reach `CREATE DATABASE` for the same database at the same time.

### Observed

On #4331, run [34611891711](https://github.com/BrighterCommand/Brighter/actions/runs/34611891711), 1 of 247:

```text
Paramore.Brighter.MSSQL.Tests.Outbox.Binary.Sync
  .WhenStoringAMessageWithARelativeDataSchemaItShouldBeReadBack
    Microsoft.Data.SqlClient.SqlException : A severe error occurred on the current command.
      at Paramore.Brighter.MSSQL.Tests.Configuration.EnsureDatabase (Configuration.cs:45)
```

It failed in **database setup at `00:00:01.51`**, not in an assertion. The container log for the same job shows the race directly — two sessions allocating a database id at the same instant:

```text
2026-09-11 14:53:55.97 spid51  [DBMgr::FindFreeDatabaseID] Next available DbId EX locked: 5
2026-09-11 14:53:55.97 spid52  [DBMgr::FindFreeDatabaseID] Next available DbId EX locked: 6
```

A re-run of the same commit passed 24 of 24, which is what makes it a flake rather than a defect in the code under test — but the flake is in this helper and will recur.

### Suggested fix

- Move `s_databaseCreated = true` out of `finally` and into the `try`, after `ExecuteNonQuery` returns, so a failure is retried rather than remembered as success.
- Make the creation tolerant of the cross-process race rather than trying to prevent it: `CREATE DATABASE` racing itself surfaces as a specific error, and catching "already exists" and re-probing is the same shape the box provisioners already use. An application-level lock (`sp_getapplock` on `master`) would also work and is what `MsSqlAdvisoryLock` already does elsewhere in this repository.

Diagnosed while working on #4343, which adds tests that call this same path.

gabisonia:
## Description

MSSQL test setup records database creation as successful even when it throws, causing later callers to skip setup. Its process-local guard also cannot coordinate the separate .NET 9 and .NET 10 test processes sharing the CI database server.

Remove the cached flag and coordinate creation through a session-owned SQL Server application lock in `master`. Both sync and async paths check the database again after acquiring the lock and propagate failures so subsequent calls can retry. Dedicated, unpooled connections release the lock on disposal, including error paths, and avoid enlisting database creation in ambient transactions. Bind the database-name lookup and quote the DDL identifier.

Add 12 regression cases covering repeated setup, failure retries across both APIs, multiple databases, concurrent callers, quoted names, and ambient transactions.

## Related Issues

Fixes #4349

## Type of Change

- [ ] Documentation update

## Checklist

- [x] I have checked the [repository testing guidance](../blob/master/.agent_instructions/testing.md) for relevant guidance
- [x] I have added/updated XML documentation for changed public helpers

## Validation

- Reproduced the failed-creation retry bug before applying the fix.
- Full MSSQL suite on .NET 10 with Azure SQL Edge: **343 passed, 4 skipped, 0 failed**.
- Test project builds for .NET 9 and .NET 10: **0 errors**; existing repository warnings remain, with none in the changed files. Runtime tests were run on .NET 10 only.
- A separate local harness compiling the same `Configuration.cs` ran three rounds of six independent processes, mixing sync and async setup against one fresh database per round: all succeeded. The harness is not included in this PR.
- Verified temporary databases were removed and `git diff --check` passed.

## Additional Notes

Setup now performs a database existence check on each call instead of relying on a process-wide cached flag. The changes are confined to MSSQL test infrastructure.


iancooper:
Thanks. One request: can you tag the issue if you take one that's assigned to someone else? In this case, it's fine, since I checked the work that triggered these. So I was going to figure out which ones to untag so everyone can pick up fixes.

But without tagging the issue, there is a risk that folks pick up the same bugs.

I am grateful for the help, and no harm, no foul. I'll try to free up the ones tagged to me later tonight if needed. Then it should be clear.

iancooper (approved):

gabisonia:
@iancooper Understood, and sorry for missing that. Thanks for clarifying!

iancooper:
> @iancooper Understood, and sorry for missing that. Thanks for clarifying!

No problem

iancooper:
PS @gabisonia I have been travelling, but will review all your issues today or tomorrow

gabisonia:
No worries, looking forward to your feedback.

@iancooper  also, when you get a chance, could you check your LinkedIn or X DMs? I sent you a quick message.

PR #4385, opened 2026-09-20
fix: poll for PostgreSQL requeue test redelivery

### LINKED ISSUE #4376: postgres-ci: hand-written gateway tests use a fixed sleep plus one read, so they tip over under added load
`postgres-ci` fails intermittently on branches that add a large generated Postgres test tree, while being solidly green on `master`. Two different tests have failed, with two different mechanisms, so this is best read as **contention under added load** rather than one broken test — but one of the two has a specific, fixable shape that is worth closing regardless.

## The baseline

| where | result |
|---|---|
| `master`, last 10 runs | **10 green / 0 fail** |
| `feature/4240-universal-transport-conformance-tests`, last 10 runs | **2 failures** |

That branch (#4297) regenerates the Postgres conformance tree, roughly doubling what the job runs. Neither failing test is touched by it — `git log master..HEAD -- <file>` is empty for both.

## Sighting 1 — the requeue test, and this one has a clear mechanism

`Paramore.Brighter.PostgresSQL.Tests.MessagingGateway.PostgreSqlMessageConsumerRequeueTestsAsync.When_requeueing_a_message_async`

```
Assert.Equal() Failure: Strings differ
↓ (pos 0)
Expected: "{"value":"Test","wasCancelled":false,"tas"···
Actual:   ""
  at .../tests/Paramore.Brighter.PostgresSQL.Tests/MessagingGateway/When_requeueing_a_message_aync.cs:line 66
```

`Actual: ""` is the noop message — the requeued message did not arrive inside the bounded receive.

The test gives the broker **a single unretried read, ~1 s wide**:

```csharp
Assert.True(await channel.RequeueAsync(message, TimeSpan.FromMilliseconds(100)));

await Task.Delay(TimeSpan.FromMilliseconds(100));      // fixed wait, not a poll

var requeuedMessage = await channel.ReceiveAsync(TimeSpan.FromMilliseconds(1000));
```

There is no poll loop: one `Task.Delay` sized exactly to the requeue delay, then one `ReceiveAsync`. If the requeue lands a little late under load, the test fails rather than waiting.

This is the same pattern that was removed from the *generated* conformance helpers, where every rejection-destination read became a single bounded poll loop instead of a fixed sleep plus one read. The hand-written tests still use the old shape.

**Suggested fix:** replace the `Task.Delay` + single `ReceiveAsync` with a bounded poll loop — poll every ~200 ms up to a few seconds, break on the first non-`MT_NONE` — so the assertion measures *whether* the message is redelivered rather than whether it beat a 1 s stopwatch.

## Sighting 2 — a different test, and probably a different cause

`Paramore.Brighter.PostgresSQL.Tests.BoxProvisioning.PostgreSqlOutboxProvisionerSchemaTests.When_postgres_per_schema_scope_is_selected_it_should_create…`

```
Npgsql.PostgresException : 42P01: relation "public.__BrighterMigrationHistory" does not exist
POSITION: 200
```

Recording it here for completeness because it lands in the same job under the same conditions, but it is **not** a timing bound — it looks like schema/migration-history isolation between concurrently running provisioning tests. If it recurs it probably deserves splitting out rather than being fixed alongside sighting 1.

## Why this is worth an issue rather than a shrug

A single 1-of-258 failure that moves between tests is easy to wave off as noise, and it has been so far. But the job is 10/10 on `master` and red twice in ten on a branch that only adds load, which is the signature of tests that were always marginal and are now being tipped over. The requeue test's ~1 s budget is the concrete instance of that, and it is cheap to fix.

gabisonia:
## Description

The handwritten PostgreSQL requeue tests sleep for 100 ms and then attempt a single receive. PostgreSQL returns immediately when no message is visible, so these tests can fail under load with an empty message.

Poll after empty reads with a five-second deadline and 200 ms pauses in both sync and async tests. Keep a one-second SQL command timeout, verify message type, identity and body, and dispose channels and producer registries. Cover both 100 ms and 1,500 ms requeue delays, and align the test names and filenames with the repository conventions.

## Related Issues

Fixes #4376.

This addresses the requeue timing failure. The separate migration-history isolation failure mentioned in the issue is outside this change.

## Type of Change

- [ ] Documentation update

## Checklist

- [x] I have checked relevant repository guidance in `.agent_instructions/testing.md` and `.agent_instructions/code_style.md`
- [x] I have added/updated XML documentation for any public API changes (not applicable: test-only changes)

## Additional Notes

Validated against PostgreSQL 17 on .NET 10:
- Before the polling fix, both 1,500 ms cases failed with the reported empty-message assertion; both 100 ms cases passed.
- Full PostgreSQL suite: 261 passed.
- Three concurrent workers, each repeating the four cases three times: 36 passed.
- After the naming cleanup: all four cases passed again.
- No compiler or analyzer warnings in the modified tests; existing project warnings remain.


iancooper (approved):
Thanks @gabisonia

PR #4394, opened 2026-09-21
fix: respect global inbox scope

### LINKED ISSUE #4336: InboxScope is referenced nowhere outside its own declaration — the scope: argument does nothing
## What happens

`InboxConfiguration` accepts a `scope:`:

```csharp
public InboxConfiguration(
    IAmAnInbox? inbox = null,
    InboxScope scope = InboxScope.All,      // <-- this
    bool onceOnly = true,
    OnceOnlyAction actionOnExists = OnceOnlyAction.Throw,
    Func<Type, string>? context = null)
```

`InboxScope` is a documented `[Flags]` enum with `Commands = 1`, `Events = 2`, `All = 3`, and `InboxConfiguration.Scope` is a public property. **Nothing reads it.**

```text
git grep -w InboxScope <ref> -- src   excluding InboxConfiguration.cs : 0
git grep -w OnceOnlyAction <ref> -- src excluding its own file (control) : 43
```

Zero at `10.7.0` and zero at `master`. `PipelineBuilder.AddGlobalInboxAttributes` — the only consumer of `InboxConfiguration` on the pipeline path — reads `Context`, `OnceOnly` and `ActionOnExists`, and never `Scope`.

## Why it matters

It reads as a real knob. Setting `scope: InboxScope.Commands` to de-duplicate commands but not events is an obvious thing to reach for, it compiles, and it changes nothing — you get `All` behaviour whatever you pass. Our own documentation carried a table explaining when to choose `Commands` over `Events`; that text was wrong and has been removed.

## Suggested

Either implement it in `AddGlobalInboxAttributes` (filter by whether the request is an `ICommand` or an `IEvent` before adding the attribute), or mark it `[Obsolete]` and drop it from the constructor. **An inert argument with a type, a default and an XML comment is indistinguishable from a working one** — which is exactly why it went undetected.

Raised from documentation work; the maintainer asked for it to be filed.

gabisonia:
## Description

`InboxConfiguration.Scope` was ignored when adding the global Inbox to handler pipelines. With `InboxScope.Commands`, publishing the same event twice could therefore throw `OnceOnlyException` even though events should be excluded.

Apply the configured scope to synchronous and asynchronous pipelines and their descriptions. Command and event requests are filtered by their marker interfaces; plain `IRequest` types retain their existing behavior. Explicit Inbox attributes and `[NoGlobalInbox]` retain their existing behavior. Update the `Scope` XML documentation to explain these rules.

## Related Issues

Fixes #4336

## Type of Change

- [ ] Documentation update

## Checklist

- [x] I have checked the [documentation](https://brightercommand.gitbook.io/paramore-brighter-documentation/brighter-configuration/brighterbasicconfiguration) for relevant guidance

## Additional Notes

Added 52 regression and compatibility cases covering Commands, Events, All, no Inbox configuration, sync/async dispatch, pipeline descriptions, explicit attributes, opt-outs, and plain `IRequest` types. Duplicate-dispatch tests use separate request instances with the same ID and the default throw behavior to reproduce the reported failure.

Validation:

- On unchanged `master` at `5bd90f208`, 12 of the new cases fail, including duplicate event publication with `InboxScope.Commands`. All 52 pass with this fix.
- Core tests on .NET 10: 1,032 passed, 7 skipped.
- Extensions tests on .NET 10: 139 passed.
- Release build of `Paramore.Brighter` for .NET Standard 2.0 and .NET 8/9/10: zero warnings or errors.


iancooper (approved):
Thanks @gabisonia

PR #4396, opened 2026-09-21
fix: preserve Inbox configuration for consumer-only applications

### LINKED ISSUE #4335: NoExternalBus() silently discards InboxConfiguration, so a consumer-only process gets no Inbox
## What happens

`AddConsumers(options => options.InboxConfiguration = new InboxConfiguration(new MsSqlInbox(config)))` does **nothing at all** in a process that registers no producers.

`AddEventBus` passes the inbox only on its two `ExternalBus` arms:

```csharp
// src/Paramore.Brighter.Extensions.DependencyInjection/ServiceCollectionExtensions.cs
:655  if (!hasEventBus) instrumentationBuilder = messagingBuilder.NoExternalBus();
:665                        serviceActivatorOptions?.InboxConfiguration);   // ExternalBus, FireAndForget
:676                        serviceActivatorOptions?.InboxConfiguration);   // ExternalBus, RPC
```

`INeedInstrumentation NoExternalBus()` (`CommandProcessorBuilder.cs:226`) **takes no inbox parameter**, so the configuration is dropped. There is no warning.

## Why it matters

**A consumer-only process is the normal shape for a receiver** — it consumes from a broker and registers no producers of its own. That is precisely the case where you want an Inbox, and precisely the case where you silently do not get one.

Measured on a receiver with the Inbox table present and provisioned, with a control:

```text
InboxMessages rows, NO producer registered   : 0
InboxMessages rows, WITH a producer registered : 1
```

The table is created, the configuration is accepted, and nothing is ever written. The symptom is "every message is handled twice" with an empty Inbox and no diagnostic.

## Suggested

Either honour the inbox on the `NoExternalBus` path, or emit a warning when `InboxConfiguration.Inbox` is a non-default instance and that path is taken. Even the warning would have saved the investigation this came out of.

## Related

`InboxScope` is inert — see the companion issue. Between them, two of the four arguments to `InboxConfiguration` do nothing in the common case.

Raised from documentation work; a published guide asserted this behaviour and had to be corrected within hours of shipping. The maintainer asked for it to be filed.

gabisonia:
## Description

Consumer-only applications configured with `AddConsumers` currently discard their global Inbox configuration. Sending the same command twice executes the handler twice and leaves the Inbox empty, even when `ValidatePipelines()` succeeds.

Pass the consumer Inbox configuration into `CommandProcessorBuilder.StartNew` and preserve it when building without an external bus. Configured receivers now write to their Inbox and apply duplicate handling consistently with pipeline validation. Explicit `[UseInbox]` and `[UseInboxAsync]` attributes retain precedence.

The new builder overload is additive; existing public interfaces and method signatures remain unchanged. Its XML documentation explains configuration precedence when using `ExternalBus`.

## Related Issues

Fixes #4335

## Type of Change

- [ ] Documentation update

## Checklist

## Additional Notes

Added 24 test cases covering synchronous/asynchronous dispatch, both `AddConsumers` configuration overloads, validation/runtime consistency, explicit attribute precedence, Inbox scopes, throw/warn duplicate handling, record-only mode, and no-Inbox compatibility. The existing JSON configuration test now isolates and restores its shared serializer options because Inbox runtime tests also use serialization.

Validation:

- Confirmed the four global-Inbox regression cases fail with the old behavior: validation succeeds, but the handler runs twice instead of once.
- Extensions tests: 163 passed.
- Core tests: 1,032 passed, 7 skipped.
- Release builds of Core and DependencyInjection passed for netstandard2.0, net8.0, net9.0, and net10.0 with zero warnings or errors.
- `git diff --check` passed.

Runtime tests used .NET 10 and InMemoryInbox. SQL Server integration was not run.


iancooper (approved):
Thanks @gabisonia

PR #4400, opened 2026-09-22
feat(aws): configurable MaximumMessageSize for SNS topics and SQS queues

slang25:
## Description

SNS now takes payloads up to 1 MiB if you raise the topic's `MaximumMessageSize` attribute, and SQS has the same ceiling, but we had no way to set either when Brighter provisions with `OnMissingChannel.Create` — so topics and queues we created were stuck at the 256 KiB default.

- `SnsAttributes` and `SqsAttributes` (V3 and V4) take an optional `maximumMessageSize`, in bytes, defaulting to `null` so nothing changes unless you ask for it
- For SNS it's applied with a `SetTopicAttributes` call after `CreateTopic`, because SNS rejects `CreateTopic` when a supplied attribute differs from the existing topic's. Nice side effect: it also raises the limit of a topic we created earlier at the default size
- For SQS it just joins the `CreateQueue` attributes, which already get re-applied to an existing queue, and the DLQ gets it too so a redriven message still fits
- Integration tests for topic creation, raising an existing topic, and queue creation, in both packages
- The AWS `Compress` and `ClaimCheck` samples had comments describing 256 KiB as a hard limit, so I've updated those

## Related Issues

_No response_

## Type of Change

- [ ] Documentation update

## Checklist

## Additional Notes

The AWS constraints here are a bit sharp and I'd welcome a second opinion on how loudly we document them: a topic above 256 KiB only supports SQS, Lambda and Data Firehose subscriptions, and caps out at 100 subscriptions. Also worth remembering that the `[Compress]` and `[ClaimCheck]` thresholds compare the uncompressed body only — compressed bodies go out base64-encoded and message attributes count against the same limit, so you want headroom.

The new tests carry the `AWS` trait, so they need the usual mock (or real AWS) to run. Branch is merged up with `master` as of `38882915d` 🙂


iancooper:
> I'd like a second opinion on how loudly we document them: a topic above 256 KiB only supports SQS, Lambda and Data Firehose subscriptions, and caps out at 100 subscriptions

We can add to release notes, and the update the information in our Docs repo, related to how you configure SQS/SNS

iancooper (approved):
@slang25 Thanks for this, looks good to me.

iancooper:
@slang25 Guessing this is done?

slang25:
@iancooper yeah this is good to go 🙂

PR #4401, opened 2026-09-22
fix: propagate commit failures from relational transaction providers' CommitAsync

### LINKED ISSUE #4399: Fire-and-forget CommitAsync in EF Core transaction providers swallows commit failures
## Summary

`CommitAsync` in the relational EF Core transaction providers discards the underlying `DbTransaction.CommitAsync` task instead of awaiting/returning it, and unconditionally returns `Task.CompletedTask`. If the real commit fails, the exception is never observed by the caller — it becomes an unobserved task exception instead of propagating to the `await provider.CommitAsync(...)` call site.

## Affected code

`src/Paramore.Brighter.MsSql.EntityFrameworkCore/MsSqlEntityFrameworkCoreTransactionProvider.cs`:

```csharp
public override Task CommitAsync(CancellationToken cancellationToken)
{
    if (HasOpenTransaction)
    {
        _context.Database.CurrentTransaction?.CommitAsync(cancellationToken);
    }

    return Task.CompletedTask;
}
```

The `CommitAsync(cancellationToken)` call on `CurrentTransaction` returns a `Task` that is never awaited, stored, or returned. The method then always returns `Task.CompletedTask`, so the caller's `await` always completes successfully regardless of whether the actual database commit succeeded, failed, or even finished.

The identical pattern exists in all four relational EF Core providers:

- `src/Paramore.Brighter.MsSql.EntityFrameworkCore/MsSqlEntityFrameworkCoreTransactionProvider.cs`
- `src/Paramore.Brighter.PostgreSql.EntityFrameworkCore/PostgreSqlEntityFrameworkTransactionProvider.cs`
- `src/Paramore.Brighter.MySql.EntityFrameworkCore/MySqlEntityFrameworkTransactionProvider.cs`
- `src/Paramore.Brighter.Sqlite.EntityFrameworkCore/SqliteEntityFrameworkTransactionProvider.cs`

For comparison, `src/Paramore.Brighter.MongoDb.EntityFramework/MongoDbEntityFrameworkTransactionProvider.cs` implements this correctly:

```csharp
public async Task CommitAsync(CancellationToken cancellationToken)
{
    var currentTransaction = _context.Database.CurrentTransaction;
    if (currentTransaction is not null)
    {
        await currentTransaction.CommitAsync(cancellationToken);
    }
}
```

The synchronous `Commit()` methods on all providers are unaffected, since `.Commit()` is called directly and inline, so exceptions propagate normally.

## Impact

The standard outbox pattern used throughout the samples (e.g. `samples/WebAPI/WebAPI_EFCore/GreetingsApp/Handlers/AddGreetingHandlerAsync.cs`) relies on `CommitAsync` throwing on failure to trigger rollback and abort the flow:

```csharp
await uow.SaveChangesAsync(cancellationToken);
await provider.CommitAsync(cancellationToken);   // expected to throw on failure
...
catch (Exception)
{
    await provider.RollbackAsync(cancellationToken);
    return await base.HandleAsync(addGreeting, cancellationToken);
}
```

Because `CommitAsync` never throws even when the real commit fails, the `catch`/rollback branch never runs, and execution falls through to `postBox.ClearOutboxAsync(...)`. This can attempt to publish outbox messages (e.g. via Azure Service Bus) for a database write that was never actually committed — a silent commit failure that is never surfaced to the caller.

## Suggested fix

Await the underlying `CommitAsync` task and let its result/exception propagate, e.g.:

```csharp
public override async Task CommitAsync(CancellationToken cancellationToken)
{
    if (HasOpenTransaction)
    {
        await _context.Database.CurrentTransaction?.CommitAsync(cancellationToken);
    }
}
```

This should be applied consistently to all four affected relational EF Core transaction providers.

MikeFreeman-Flagstone:
## Summary

`CommitAsync` in the MsSql, PostgreSql, MySql, and Sqlite EF Core transaction providers discarded the `Task` returned by the underlying `IDbContextTransaction.CommitAsync` and unconditionally returned `Task.CompletedTask`. A failed database commit therefore never propagated to the caller — `await provider.CommitAsync(...)` always completed successfully, so the caller's `catch`/rollback branch (the standard outbox pattern used throughout this repo's samples) never ran, and outbox messages could be published for a write that was never actually committed.

The non-EF `MySqlTransactionProvider` had the identical defect, and additionally nulled its `Transaction` field *while the commit was still in flight*.

Closes #4399.

## Root cause & evidence

- All four `CommitAsync` overrides were non-`async` methods invoking the real commit as a discarded expression statement, then returning an already-completed task on every path — including the faulted-commit path.
- `CS4014` (the "this call is not awaited" warning) only fires inside `async` methods, so `TreatWarningsAsErrors=true` never caught it here.
- The already-correct `MongoDbEntityFrameworkTransactionProvider.CommitAsync` (reads the transaction into a local, `is not null` guard, `await`s the real commit) was the reference for the fix shape.
- The issue's originally suggested one-line fix (`await x?.CommitAsync(...)`) does not compile cleanly under this repo's `Nullable enable` + `TreatWarningsAsErrors=true` (CS8602 dereference-of-possibly-null), and independently introduces a new NRE via a double-read of `CurrentTransaction`. The applied fix avoids both.

## Scope

Confirmed via code-trace (see `bugfixes/0023-efcore-commitasync-fire-and-forget/bugfix.md` for the full triage → confirm → test → fix → verify record):

- `src/Paramore.Brighter.MsSql.EntityFrameworkCore/MsSqlEntityFrameworkCoreTransactionProvider.cs`
- `src/Paramore.Brighter.PostgreSql.EntityFrameworkCore/PostgreSqlEntityFrameworkTransactionProvider.cs`
- `src/Paramore.Brighter.MySql.EntityFrameworkCore/MySqlEntityFrameworkTransactionProvider.cs`
- `src/Paramore.Brighter.Sqlite.EntityFrameworkCore/SqliteEntityFrameworkTransactionProvider.cs`
- `src/Paramore.Brighter.MySql/MySqlTransactionProvider.cs` (non-EF; identical defect, not originally named in #4399, folded in for cross-backend parity)

**Not fixed here (flagged as a follow-up)**: fixing commit-propagation makes the rollback path reachable for the first time on `PostgreSqlEntityFrameworkTransactionProvider` and `SqliteEntityFrameworkTransactionProvider` (which override neither `Rollback()` nor `RollbackAsync()`) and on synchronous `Rollback()` for the other two EF providers (which override only `RollbackAsync`). Today this is masked because commit never threw; once it does, the inherited rollback dereferences a null base `Transaction` field, throwing an `NullReferenceException` that's silently swallowed by a `catch(Exception) { /*ignore*/ }`. This is a distinct defect class (rollback/state ownership across the EF provider family) and needs its own fix and tests — happy to file as a separate issue.

## Testing

- Added FakeItEasy-based regression tests for all four EF Core providers in `Paramore.Brighter.Extensions.Tests` (no live database required), mirroring the existing `MongoDbEntityFrameworkTransactionProviderAsyncTest` pattern. The MySql EF Core case is guarded with `#if NET9_0` since that provider doesn't target net10.0.
- `MySqlTransactionProvider` (non-EF) has no automated regression test: it hard-casts to the sealed `MySqlConnector.MySqlTransaction`, which has no public/fakeable constructor, so it can't be tested without live MySQL infrastructure. Fixed via code-trace verification only (`test after`, per this repo's testing guidelines for I/O implementations where test-first is impractical).
- All 4 new tests pass on net9.0 (4/4) and net10.0 (3/4, correctly excluding the net9.0-only MySql EF Core case).
- Full `Paramore.Brighter.Extensions.Tests` suite: 174/174 passing on net9.0, 173/173 on net10.0 — no regressions.
- `Paramore.Brighter.MySql` builds clean on net8.0/net9.0/net10.0.

## Test plan
- [x] New regression tests fail before the fix (RED) and pass after (GREEN)
- [x] Full `Paramore.Brighter.Extensions.Tests` suite passes on net9.0 and net10.0
- [x] `Paramore.Brighter.MySql` (non-EF) builds clean on all target frameworks


iancooper:
Thanks @MikeFreeman-Flagstone 

Any chance we could add this? "his is a distinct defect class (rollback/state ownership across the EF provider family) and needs its own fix and tests — happy to file as a separate issue."

src/Paramore.Brighter.MySql.EntityFrameworkCore/MySqlEntityFrameworkTransactionProvider.cs
```diff
@@ -40,14 +40,13 @@ public override void Commit()
         /// Commit the transaction
         /// </summary>
         /// <returns>An awaitable Task</returns>
-        public override Task CommitAsync(CancellationToken cancellationToken)
+        public override async Task CommitAsync(CancellationToken cancellationToken)
         {
-            if (HasOpenTransaction)
+            var currentTransaction = _context.Database.CurrentTransaction;
```
iancooper, line 45: In some cases, we are now using HasOpenTransaction; in some, we are not. We ought to reconcile these and just use one approach

iancooper (commented):
@MikeFreeman-Flagstone Minor thing to pick up here, otherwise this is good

src/Paramore.Brighter.MySql.EntityFrameworkCore/MySqlEntityFrameworkTransactionProvider.cs
MikeFreeman-Flagstone: I'm dropping the HasOpenTransaction.  We found if (HasOpenTransaction) reads _context.Database.CurrentTransaction once, then the body reads it again — a TOCTOU double-read against a live, externally-mutable property. That's not just style; it's a real correctness reason not to revert those four.

Therefore bringing MySqlTransactionProvider in line with the rest.

MikeFreeman-Flagstone: Good catch — reconciled in 2660630. The 4 EF Core providers deliberately read `CurrentTransaction` into a local once (to avoid the TOCTOU double-read the Confirm phase flagged for this bug), so I made `MySqlTransactionProvider` match the same idiom instead of reverting the others:

```csharp
public override async Task CommitAsync(CancellationToken cancellationToken)
{
    if (Transaction is MySqlTransaction transaction)
    {
        await transaction.CommitAsync(cancellationToken);
        Transaction = null;
    }
}
```

Also drops the hard-cast + null-forgiving `(MySqlTransaction)Transaction!`. No behaviour change for real usage — rebuilt clean on net8/9/10 and re-ran the full Extensions.Tests suite (174/174, no regressions).

iancooper (approved):

iancooper:
@MikeFreeman-Flagstone All good, merging for you

MikeFreeman-Flagstone:
Yes — done in 70514df, folded into this PR rather than a separate issue.

Root cause: none of the four EF Core providers (MsSql, PostgreSql, MySql, Sqlite) populate the base `RelationalDbTransactionProvider.Transaction` field, while they override `HasOpenTransaction` to report EF's own `CurrentTransaction` state. That decouples the guard from the data — the inherited `Rollback()`/`RollbackAsync()` guard passes, then dereferences the permanently-null `Transaction` field, throwing an NRE that the base's `catch(Exception) { /*ignore*/ }` silently swallows. Sync `Rollback()` was a complete no-op on all four; async `RollbackAsync()` was the same no-op on PostgreSql/Sqlite, while MsSql/MySql did roll back the raw ADO.NET transaction on live infra but bypassed EF's own bookkeeping and swallowed genuine failures too.

Worth calling out since it's a real design decision: the fix **propagates** rollback failures rather than swallowing them (matching `MongoDbEntityFrameworkTransactionProvider` and `SpannerUnitOfWork`, the two providers on this interface that already do it correctly). I checked — nothing in Brighter's own core ever calls `Rollback`/`RollbackAsync` on `IAmABoxTransactionProvider`, so the only affected call sites are the two EF Core sample handlers, and a silently-failed rollback is strictly worse there than a surfaced one. Happy to reconsider if you'd rather it stayed swallow-and-ignore.

Added 8 regression tests (4 providers × sync/async), no live database required, same FakeItEasy technique as the CommitAsync tests. Full suite green, no regressions (182/182 net9.0, 179/179 net10.0).

iancooper:
@MikeFreeman-Flagstone You will need to re-open this PR for those, as I can't

MikeFreeman-Flagstone:
Ah, right — a merged PR can't actually be reopened on GitHub (I tried: "can't be reopened because it was already merged"), that's a hard platform rule rather than a permissions thing. Opened a fresh PR for the rollback fix instead: #4405

PR #4402, opened 2026-09-22
docs: add ADR for outbox circuit breaker trip logging

MikeFreeman-Flagstone:
## Summary
- Adds ADR 0058, proposing that `InMemoryOutboxCircuitBreaker` log its own state transitions (trip / re-trip / reset) directly, since today nothing logs when `TripTopic()` or `CoolDown()` change a topic's circuit-breaker state.
- Recommends: Warning on a fresh trip, Debug on a re-trip (to avoid flooding Warning-level sinks during a sustained outage where every failed publish re-trips the topic), Information on a cooldown reset — following the codebase's existing static-logger + source-generated `LoggerMessage` convention (`ApplicationLogging.CreateLogger<T>()`), with no changes to `IAmAnOutboxCircuitBreaker` or `OutboxCircuitBreakerOptions`.
- Explicitly scopes out fixing the sync `Dispatch()` path (which never calls `TripTopic` today) — tracked as a separate issue.
- Status: Proposed — this PR is for design review before any implementation/task breakdown.

## Test plan
- [ ] N/A — documentation-only change (ADR). No code, no tests in this PR.
- [ ] Design review / discussion on the ADR itself.


iancooper (approved):
@MikeFreeman-Flagstone This looks good. I wonder if we want to add OTel for this too? Up to you if you want to extend the scope.

iancooper (approved):

